PR #170, opened 2014-12-23
MAINT: lmfit.printfuncs.fit_report, using alphanumeric sort

andyfaff:
lmfit.printfuncs.fit_report is a nice way of print the results of a fit.  However, it sorts the dictionary before printing (line83):

parnames = sorted(params)

At the moment I am using parameter names ['p0', 'p1', ...., 'p10', 'p11', ...]
Consequently fit_report prints them in the order

p0
p1
p10
p11
p2

I would much rather we have an alphanumeric sort, or for the dictionary to remain ordered in the way they first got inserted in the dictionary

p0
p1
p2
p3
...
p10
p11

Is there a reason for this sorting?  Shouldn't we keep the natural ordering of the lmfit.parameters.Parameters object, or at least use alphanumeric ordering?

This PR uses alphanumeric sorting.

### DIFF
--- a/lmfit/printfuncs.py
+++ b/lmfit/printfuncs.py
@@ -66,7 +66,8 @@ def gformat(val, length=11):
 
 CORREL_HEAD = '[[Correlations]] (unreported correlations are < % .3f)'
 
-def fit_report(inpars, modelpars=None, show_correl=True, min_correl=0.1):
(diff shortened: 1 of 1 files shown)


newville:
Thanks, and sorry I forgot to merge this earlier!

Issue #174, opened 2014-12-26
Misleading documentation for Minimize.minimize - Minimize.params is not the same Parameters instance used to initialise the Minimizer.


andyfaff:
Lets consider the following code:

    import numpy as np
    from lmfit import minimize, Parameters, Parameter

    x = np.linspace(0, 15, 301)
    data = (5. * np.sin(2 * x - 0.1) * np.exp(-x*x*0.025) +
            np.random.normal(size=len(x), scale=0.2) )

    # define objective function: returns the array to be minimized
    def fcn2min(params, x, data):
        """ model decaying sine wave, subtract data"""
        amp = params['amp'].value
        shift = params['shift'].value
        omega = params['omega'].value
        decay = params['decay'].value

        model = amp * np.sin(x * omega + shift) * np.exp(-x*x*decay)
        return model - data

    # create a set of Parameters
    params = Parameters()
    params.add('amp',   value= 10,  min=0)
    params.add('decay', value= 0.1)
    params.add('shift', value= 0.0, min=-np.pi/2., max=np.pi/2)
    params.add('omega', value= 3.0)

    # do fit, here with leastsq model
    result = minimize(fcn2min, params, args=(x, data))

    print result.params is params

The last line prints `False`.  i.e. `result.params` is not the same instance as the `params` that was used to construct the `Minimizer` object.  This is because in `Minimizer.unprepare_fit` there is a deepcopy:

    self.params = deepcopy(self.params)

I would naively expect
    
    result.params is params

to be `True`.  Why is a copy done in `unprepare_fit`? Is it done to keep the initial `Parameters` object unchanged?  If so, then it would be better to do the deepcopy before `Minimizer.minimize` is called.
For example you could have the confusing behaviour:
     
    obj = Minimizer(fcn2min, params, args=(x, data))
    #prints True
    obj.params is params
    obj.minimize()
    #prints False
    obj.params is params

danielballan:
@andyfaff Above, Matt alluded to an earlier discussion on this copy-on-input issue. It was in #56. His key points are in [this comment](https://github.com/lmfit/lmfit-py/issues/56#issuecomment-30375780). I agree with some of those points, in particular that changing to copy-on-input would break a lot of old user code. The original poster of that issue suggested working around the problem by implementing a new class, `Model`. Which, of course, we did.

[...]

PR #175, opened 2014-12-26
DOC : add warning to minimize

tacaswell:
Make it very hard to miss that `minimize` mutates it's itput
values.

### DIFF
--- a/doc/fitting.rst
+++ b/doc/fitting.rst
@@ -20,6 +20,9 @@ details on writing the function to minimize.
 
 .. function:: minimize(function, params[, args=None[, kws=None[, method='leastsq'[, scale_covar=True[, iter_cb=None[, **leastsq_kws]]]]]])
 
+   .. warning ::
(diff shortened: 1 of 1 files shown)


newville:
I think this old-ish PR *is* superseded by #174,  and addressed with #230.   Closing, but could be persuaded to re-open if there's still a need for this.

tacaswell:
@newville Can you point me to where this is changed in the code?

newville:
@tacaswell Well, #230 makes many changes to the `Minimizer` class, such that a `MinimizerResult` is created at the start of each fit (in `Minimizer.prepare_fit()`, https://github.com/lmfit/lmfit-py/blob/master/lmfit/minimizer.py#L305), and `Minimizer.params` are (deep)copied to `result.params`, which are actually used in the fit.   This means that the `params` passed into `Minimizer()` are unchanged, and that each fit gets a fresh copy of these input parameters.   

So, I think this issue is now addressed.

PR #176, opened 2014-12-26
Add CompositeModel hints test

tritemio:
* Add tests for hints propagation from base models to composite models
* Improve error messages for name collisions and non initialized parameters

### DIFF
--- a/tests/test_model.py
+++ b/tests/test_model.py
@@ -344,6 +344,23 @@ class TestUserDefiniedModel(CommonTests, unittest.TestCase):
         for mod in [model1, model2, model3, model4]:
             self.assertTrue(mod in model_total3.components)
 
+    def test_hints_in_composite_models(self):
+        # test propagation of hints from base models to composite model
+        def func(x, amplitude):
+            pass
+
+        m1 = Model(func, prefix='p1_')
+        m2 = Model(func, prefix='p2_')
+
+        m1.set_param_hint('amplitude', value=1)
+        m2.set_param_hint('amplitude', value=2)
+
+        mx = (m1 + m2)
+        params = mx.make_params()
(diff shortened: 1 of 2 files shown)


tests/test_model.py
```diff
@@ -344,6 +344,23 @@ def test_sum_composite_models(self):
...
+        m2 = Model(func, prefix='p2_')
+
+        m1.set_param_hint('amplitude', value=1)
+        m2.set_param_hint('amplitude', value=2)
+
+        mx = (m1 + m2)
+        params = mx.make_params()
+        param_values = {name: p.value for name, p in params.items()}
+        self.assertTrue(param_values['p1_amplitude'] == 1)
+        self.assertTrue(param_values['p2_amplitude'] == 2)
```
danielballan: Just FYI, `self.assertEqual` would be better here. It gives a more useful error message when violated, printing the actual and expected values.

danielballan:
I really like the improved error messages. This has been on my wishlist for awhile.

tests/test_model.py
tritemio: Good point, I will change it.

tritemio:
Changed assertTrue with assertEqual.

In the meanwhile I noticed that [this other test](https://github.com/tritemio/lmfit-py/blob/eae879458a82802c168259d78a8e1f7fad8c9d2e/tests/test_model.py#L197) may be missing a final assertion?

newville:
@trimetio @danielballan : This looks good.  I'd be willing to merge, but also willing to wait for that **other test** to be clarified.  Suggestions?

tritemio:
The "other test" is an unrelated problem I spotted in the same file while I was adding this test. I opened a new issue #184 so it will not be forgotten.

newville:
Great, thanks!

PR #178, opened 2014-12-27
ENH: Minimizer, adding mcmc sampler.

andyfaff:
My attempt at adding an MCMC sampler to lmfit.Minimizer.  MCMC is a way of sampling the posterior distribution for a parameter set.    Once the sampler finishes running the Minimizer.params attribute is updated using the statistics of the MCMC samples.  The mean of a parameter trace gives the 'fitted value' and the SD gives the parameter uncertainty.  You can work out the correlations as well.  The mcmc method returns the sampler instance, from which you can get a lot of extra information.  It's worth reading up from: http://pymc-devs.github.io/pymc/ to learn what kind of information you can extract from the sampler.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,7 @@ before_install:
     - conda update --yes conda
 
 install:
-    - conda install --yes pip python=$TRAVIS_PYTHON_VERSION numpy scipy pandas matplotlib dateutil nose
+    - conda install --yes pip python=$TRAVIS_PYTHON_VERSION numpy scipy pandas matplotlib dateutil nose pymc
     - python setup.py install
 
 script:
--- a/doc/fitting.rst
+++ b/doc/fitting.rst
@@ -339,6 +339,7 @@ The Minimizer object has a few public methods:
 
 .. method:: lbfgsb(**kws)
 
+
    perform fit with L-BFGS-B algorithm.  Keywords will be passed directly to
    :func:`scipy.optimize.fmin_l_bfgs_b`.
 
@@ -353,6 +354,10 @@ The Minimizer object has a few public methods:
     +------------------+----------------+------------------------------------------------------------+
     |   maxfun         | 2000*(nvar+1)  | maximum number of function calls (nvar= # of variables)    |
     +------------------+----------------+------------------------------------------------------------+
+
+.. warning::
+
+  :meth:`lbfgsb` is deprecated.  Use :meth:`minimize` with ``method='lbfgsb'``.
 
 .. method:: fmin(**kws)
 
@@ -369,6 +374,10 @@ The Minimizer object has a few public methods:
     +------------------+----------------+------------------------------------------------------------+
     |   maxfun         | 5000*(nvar+1)  | maximum number of function calls (nvar= # of variables)    |
     +------------------+----------------+------------------------------------------------------------+
+
+.. warning::
+
+  :meth:`fmin` is deprecated.  Use :meth:`minimize` with ``method='nelder'``.
 
 
 .. method:: scalar_minimize(method='Nelder-Mead', hess=None, tol=None, **kws)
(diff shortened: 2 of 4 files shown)


Tillsten:
@andyfaff Sorry for the slow response, i am on vacations :). First I'll address the second point: Yeah,
it is exactly what i meant, the sampling has a loot of hooks and nooks to play around with, getting all that in lmfit is quite a lot of work. I think my favorite solution would be an exporter.py module which would allow to "from lmfit.exporter import to_pymc" , which would just return a pymc model.

About the standard deviation: if the the std. value is in the ballpark of the real std, the sampling still works 
because it will be correct by adjusting the normalization (the denominator in bayes law, which is not calculated in mc). I think the right thing here is to add an stochastic for the deviation if no weight are given.  Irc the correct way is use an Gamma-distribution for the the precision (which is 1/std**2) with an gamma distribution, but i am murky on the details. Of course this is only necessary if the weights are
not given. One can get easily an estimate it from the fitted model.

andyfaff:
@Tillsten w.r.t the standard deviations, I'm pretty sure that the std deviations will be taken into account in the sampling procedure.  However this is IFF the user objective function returns the residuals after dividing by the std. i.e. the objective function needs to return

    (y - model) / std

I'll check with the pymc crew on stackoverflow to make sure.

andyfaff:
>it is exactly what i meant, the sampling has a loot of hooks and nooks to play around with, getting all that in lmfit is quite a lot of work. I think my favorite solution would be an exporter.py module which would allow to "from lmfit.exporter import to_pymc" , which would just return a pymc model.

The pymc.mcmc instance, `MDL`, could even be made in `Minimizer.__init__`.  It doesn't actually take any significant amount of calculation to do that, just to do all the sampling.  
However, there are also options required to construct the mcmc object.  For example, one has to select whether to use the `pymc.AdaptiveMetropolis` or the default `pymc.Metropolis` sampler.  The first can use a lot less function evaluations, but the second is more robust.

As I saw it one of the main issues with *not* doing the sampling and *not* calculating the statistics (means, etc) in lmfit is that it may be beyond the ability of a user to calculate all that.  
How about this approach:

1. Construct the mcmc object in `Minimizer.__init__`.
2. Have a separate `mcmc_sample` method to do the sampling and calculate statistics.

The only real thing that the user has to modify is the `sample`, `burn` and `thin` parameters.  I think they could be understood relatively easily with good documentation.  If the user wants to do anything more complicated they can just access the `Minimizer.MDL` attribute directly.

andyfaff:
@Tillsten I think it'll be quite hard just to return/export the MDL sampling object and leave it up to the user to do all the sampling, etc.

The reason is that constructing the MDL sampler relies inherently on the `Minimizer` object.  For example, the `Minimizer` needs to be left in a *prepared* state for the sampler to work as the *prepared* state holds all the constraints, dependencies, var_map, etc. Secondly, the sampler is setup with a specific set of parameters.
If the `mcmc` method is solely used to return a sampler object, then the user can't do anything to affect the state of the Minimizer, e.g. by performing a `leastsq` fit, or making any changes to the `Minimizer.params` (such as changing which parameters vary).  If they do make changes then the MDL object won't work because the `Minimizer` won't be in a prepared state, or there will be problems with the `Minimizer.params`, etc.

Another alternative is to have the following signature for `mcmc`:

    def mcmc(self, samples, sampler_kwds=None, mcmc_kwds=None):

The `sampler_kwds` are used for the sampling:

    MDL.sample(samples, **sampler_kwds)

The `mcmc_kwds` are used to construct the sampler:

    MDL = pymc.MCMC([likelihood, p], **mcmc_kwds)

Tillsten:
@andyfaff sorry for the delay, will have a look at it tomorrow.

andyfaff:
Closing because #249 does pretty much the same thing.

PR #181, opened 2015-01-02
BUG: printfuncs was not printing large negative correlations

andyfaff:
If two parameters had a negative correlation that was large then this is not displayed by `lmfit.printfuncs.fit_report`.
I.e. if `abs(correlation) > 0.1`, then the correlation should be printed.  This PR should fix the behaviour.


Tillsten:
lgtm!

newville:
Yep, good catch.  Thanks!

andyfaff:
Is there any testing for the printing functionality?

Issue #56, opened 2013-12-11
issues with minimize() modifying Parameters object


vnmanoharan:
I'm running into some "gotchas" when working with lmfit that seem to be related to how it modifies the Parameters object passed to minimize().  Consider the following code that fits a model to two different datasets sequentially:

``` Python
import numpy as np
from lmfit import Parameters, minimize, report_fit

x = np.arange(0, 1, 0.01)
y1 = 1.0*np.exp(1.0*x)
y2 = 1.0 + x + 1/2.*x**2 +1/3.*x**3

def residual(params, x, data):
    a = params['a'].value
    b = params['b'].value

    model = a*np.exp(b*x)
    return (data-model)

params = Parameters()
params.add('a', value = 2.0)
params.add('b', value = 2.0)

# fit to first data set
out1 = minimize(residual, params, args=(x, y1))
print "\n out1.params from fit to y1"
report_fit(out1.params)

# fit to second data set
out2 = minimize(residual, params, args=(x, y2))
print "\n out2.params from fit to y2"
report_fit(out2.params)

# now look at first fit results again
print "\n out1.params again"
report_fit(out1.params)
```

which outputs:

```
 out1.params from fit to y1
[[Variables]]
     a:     1 +/- 0 (0.00%) initial =  2
     b:     1 +/- 0 (0.00%) initial =  2
[[Correlations]] (unreported correlations are <  0.100)
    C(a, b)                      =  nan 

 out2.params from fit to y2
[[Variables]]
     a:     0.9892979 +/- 0.0007045643 (0.07%) initial =  1
     b:     1.049585 +/- 0.001006103 (0.10%) initial =  1
[[Correlations]] (unreported correlations are <  0.100)
    C(a, b)                      = -0.930 

 out1.params again
[[Variables]]
     a:     0.9892979 +/- 0.0007045643 (0.07%) initial =  1
     b:     1.049585 +/- 0.001006103 (0.10%) initial =  1
[[Correlations]] (unreported correlations are <  0.100)
    C(a, b)                      = -0.930 
```

Two things happen that are unexpected, at least to me: 1) the params object passed to the second minimize() call has been modified by the first minimize() call.  The user might expect that it should remain as specified.  2) the output of the first fit is modified by the second fit, which seems to be because the Minimizer() object includes a reference to the params object rather than a copy.

[...]

PR #183, opened 2015-01-04
Minimizer.params is always the same instance as the params used to construct the Minimizer, closes #174.

### LINKED ISSUE #174: Misleading documentation for Minimize.minimize - Minimize.params is not the same Parameters instance used to initialise the Minimizer.

andyfaff:
@danielballan to kick this PR off, I've modified the test_copy_params.py file, which was added to test #56.
I've added an extra test `test_copy_params_Minimizer`, that checks if the `Minimizer.params` attribute is the same instance as the `params` object used to construct the `Minimizer` in the first place.

Without the "fix" I get the following error:

======================================================================
FAIL: test_copy_params_Minimizer (test_copy_params.test_copy_params)
----------------------------------------------------------------------
  File "/Users/anz/Documents/Andy/programming/lmfit-py/tests/test_copy_params.py", line 80, in test_copy_params_Minimizer
    assert_(params is fitter.params)
  File "/Users/anz/Documents/Andy/programming/dev2/lib/python2.7/site-packages/numpy/testing/utils.py", line 53, in assert_
    raise AssertionError(smsg)
AssertionError

----------------------------------------------------------------------

### DIFF
--- a/lmfit/minimizer.py
+++ b/lmfit/minimizer.py
@@ -392,7 +392,6 @@ class Minimizer(object):
         removes ast compilations of constraint expressions
         """
         self.__prepared = False
-        self.params = deepcopy(self.params)
         for par in self.params.values():
             if hasattr(par, 'ast'):
                 delattr(par, 'ast')
@@ -814,4 +813,8 @@ def minimize(fcn, params, method='leastsq', args=None, kws=None,
     fitter = Minimizer(fcn, params, fcn_args=args, fcn_kws=kws,
                        iter_cb=iter_cb, scale_covar=scale_covar, **fit_kws)
     fitter.minimize(method=method)
+
+    params_copy = deepcopy(fitter.params)
+    fitter.params = params_copy
     return fitter
+
--- a/tests/test_copy_params.py
+++ b/tests/test_copy_params.py
@@ -22,7 +22,7 @@ class test_copy_params(unittest.TestCase):
         return (data - model)
 
     def params_values(self, params):
-        return params.valuesdict().values()
+        return np.asfarray(params.valuesdict().values())
 
     def test_copy_params(self):
         # checking output for gh-#56
--- a/tests/test_copy_params.py
+++ b/tests/test_copy_params.py
@@ -22,7 +22,7 @@ class test_copy_params(unittest.TestCase):
         return (data - model)
 
     def params_values(self, params):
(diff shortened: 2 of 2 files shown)


andyfaff:
The first commit obviously fails the tests I've just added.  But the second should pass, as I've made the amendments to minimizer.py

newville:
I believe this issue is addressed as of the merge of #230.  Is it OK to close this?

newville:
I'm pretty sure that #230 makes this PR impossible to merge.    `minimize.py` has changed too much, and the test of equality of `params` and `out1.params` in in the `test_copy_params.py` in this PR will certainly fail, by design.

If there are issues regarding values or identity of input and output `params` post-#230, please re-submit new relative to the current master.

PR #192, opened 2015-01-15
Guard against missing docstrings under python -OO (Fixes #191)

### LINKED ISSUE #191: Check for None in __doc__ before changing it.
Python code byte-compiled with -OO has doc-strings stripped out. 

This creates problems when compiling different packages which changes the doc-strings by doing something like this:
     __doc__ += "additional text"
(when the docstring is 'None', this will fail).

lmfit/models.py does this and thus creates problems when trying to bytecompile it with the -OO option using  cx_freeze/py2exe.

In order to fix this, I would propose to change all the lines in lmfit/models.py which say:
     __doc__ = XXXX.__doc__ + COMMON_DOC
with
    __doc__ = XXXX.__doc__ + COMMON_DOC if __doc__ else COMMON_DOC

See also:
http://bugs.python.org/issue23189#
http://stackoverflow.com/questions/22299532/unsupported-operand-types-for-nonetype-and-str-winappdbg-error-after-c

pbrod:
Fixes #191

### DIFF
--- a/lmfit/models.py
+++ b/lmfit/models.py
@@ -97,7 +97,7 @@ class ConstantModel(Model):
 
 
 class LinearModel(Model):
-    __doc__ = linear.__doc__ + COMMON_DOC
+    __doc__ = linear.__doc__ + COMMON_DOC if linear.__doc__ else ""
     def __init__(self, *args, **kwargs):
         super(LinearModel, self).__init__(linear, *args, **kwargs)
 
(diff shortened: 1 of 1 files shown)


newville:
Thanks!! LGTM.  Any objections in merging this for 0.8.3?

danielballan:
Looks to me also.

tritemio:
+1

PR #193, opened 2015-01-15
fixes MPLFitter NameError when ipython is installed but matplotlib isn't

anthonyalmarza:
Got the following error when trying to use lmfit in an ipython session.
```
/.../venv/lib/python2.7/site-packages/lmfit/__init__.py in <module>()
     47 from .uncertainties import ufloat, correlated_values
     48
---> 49 from .ui import Fitter
     50
     51 ## versioneer code

/.../venv/lib/python2.7/site-packages/lmfit/ui/__init__.py in <module>()
     38     from .ipy_fitter import NotebookFitter
     39     BaseFitter = BaseFitter
---> 40     MPLFitter = MPLFitter
     41     Fitter = NotebookFitter
     42

NameError: name 'MPLFitter' is not defined
```

### DIFF
--- a/lmfit/ui/__init__.py
+++ b/lmfit/ui/__init__.py
@@ -36,7 +36,7 @@ if has_matplotlib:
 
 if has_ipython:
     from .ipy_fitter import NotebookFitter
+    from .basefitter import MPLFitter
     BaseFitter = BaseFitter
     MPLFitter = MPLFitter
     Fitter = NotebookFitter
-


newville:
Thanks!    I think this can't possibly cause problems, so will merge.

danielballan:
Actually I would have removed the line

    MPLFitter = MPLFitter

rather than adding the import. If mpl is not installed, `MPLFitter` should not be imported.

newville:
@danielballan  that would be easy enough to change.  Should we just make that change?

danielballan:
Yeah, we can leave that false start in and just merge a fix. Do you mind writing it?

newville:
@danielballan  yes, I can do this.  In fact, would it be OK to have this in `ui\__init__.py`:

```python

from .basefitter import BaseFitter
Fitter = BaseFitter
if has_matplotlib:
    from .basefitter import MPLFitter
    Fitter = MPLFitter

if has_ipython:
    from .ipy_fitter import NotebookFitter
    Fitter = NotebookFitter
```

that is, also getting rid of some `BaseFitter = BaseFitter`?

newville:
I think this is now fixed.

PR #195, opened 2015-01-17
add dump()/load() to Parameters class

newville:
This PR gives the Parameters class methods dump/dumps/load/loads to use a JSON representation of the Parameters.  

A few other minor fixes to Parameter class are included, for example that having 'min==max' now raises a ValueError.

### DIFF
--- a/lmfit/parameter.py
+++ b/lmfit/parameter.py
@@ -1,6 +1,7 @@
 """
 Parameter class
 """
+from __future__ import division
 from numpy import arcsin, cos, sin, sqrt, inf, nan
 import json
 import sys
--- a/lmfit/parameter.py
+++ b/lmfit/parameter.py
@@ -3,7 +3,7 @@ Parameter class
 """
 from numpy import arcsin, cos, sin, sqrt, inf, nan
 import json
-
+import sys
 try:
     from collections import OrderedDict
 except ImportError:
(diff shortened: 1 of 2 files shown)


lmfit/parameter.py
```diff
@@ -238,19 +318,19 @@ def setup_bounds(self):
...
+            self.from_internal = lambda val: self.min - 1.0 + sqrt(val*val + 1)
+            _val  = sqrt((self._val - self.min + 1.0)**2 - 1)
         elif self.min in (None, -inf):
             self.from_internal = lambda val: self.max + 1 - sqrt(val*val + 1)
-            _val  = sqrt((self.max - self._val + 1)**2 - 1)
+            _val  = sqrt((self.max - self._val + 1.0)**2 - 1)
         else:
             self.from_internal = lambda val: self.min + (sin(val) + 1) * \
-                                 (self.max - self.min) / 2
+                                 (self.max - self.min) / 2.0
```
danielballan: We should have `from __future__ import division` to handle this kind of thing, part of the standard header for packages supporting 2 and 3.

danielballan:
This is great.

tritemio:
This is very useful! Thanks.

The only remark I have is that I would rename `dumps` and `loads` to `dumpstr`  `loadstr` (or something similar) to be more explicit.

newville:
@danielballan I added `from __future__ import division` to parameters.py

@tritemio  `dumps()` is what the Pickle and json modules use for "dump to string".  I agree that `dumpstr` is a better name, but am willing to follow the precedence of the stdlib.

Issue #198, opened 2015-01-23
adding 2 Parameters alters one of them


newville:
From a private conversation with @Luindil  : 

Adding 2 parameters:

```python 
par_sum = params1 + params2
```

alters `params1` to be the same as `par_sum`.   I can't imagine why that would be needed.

newville:
fixed with 17a14c0ec4fde377113252a2317333159d33bd27

PR #199, opened 2015-01-23
Parameters.__add__() makes a fresh copy for return

newville:
This addresses issue #198 

I cannot think of any reason where the current behavior is desirable.


tacaswell:
It might be worth keeping around the current version renamed as `__iadd__` so that `a += b` works as expected.

newville:
@tacaswell  Thanks! that's a good point.  Will add `__iadd__()`

PR #204, opened 2015-02-09
Model selection methods for ModelFit

yoavram:
I added a method called `aic` to `ModelFit`.
The method calculates and returns the AIC of a model fit based on the formula:
```
n log(rss/n) + 2k
```
where n is the number of data points, rss is the sum squares of the residuals and k is the number of varying parameters.
I included a simple test (`test_aic` in `test_model.py`). 
I ran the tests with `nosetests` and got `OK (SKIP=2)`.

### DIFF
--- a/lmfit/model.py
+++ b/lmfit/model.py
@@ -995,10 +995,18 @@ class ModelFit(Minimizer):
         return fig
 
     def aic(self):
-        """calculate and return the AIC for a ModelFit.
+        """Calculate and return the AIC for a ModelFit.
         AIC is Akaike information criterion, a measure of the relative quality of a statistical model for a given set of data.
         It is computed as `n * log(rss/n) + 2k`, where `n` is the number of data points, `rss` is the sum of residuals squares, 
         and k in the number of varying parameters.
+        
+        Returns
+        -------
+        aic : float
+
+        See also
+        --------
+        http://en.wikipedia.org/wiki/Akaike_information_criterion
         """
         rss = self.chisqr
         n = self.ndata
@@ -1006,11 +1014,19 @@ class ModelFit(Minimizer):
         return n * np.log(rss / n) + 2 * k
 
     def bic(self):
-        """calculate and return the BIC for a ModelFit.
+        """Calculate and return the BIC for a ModelFit.
         BIC is Bayesian information criterion, a measure of the relative quality of a statistical model for a given set of data.
         It is computed as `n * log(rss/n) + log(n) k`, where `n` is the number of data points, `rss` is the sum of residuals squares, 
         and k in the number of varying parameters. 
         It is closely related to AIC but gives a larger penalty to the number of paramters in the model.
+
+        Returns
+        -------
+        bic : float
+
+        See also
+        --------
+        http://en.wikipedia.org/wiki/Bayesian_information_criterion
         """
         rss = self.chisqr
         n = self.ndata
@@ -1018,8 +1034,8 @@ class ModelFit(Minimizer):
         return n * np.log(rss / n) + np.log(n) * k
 
 
-    def likelihood_ratio_test(self, other):
-        """calculate a likelihood ratio test for two model fits (self and other).
+    def lik_ratio(self, other):
+        """Calculate a likelihood ratio test for two model fits (self and other).
 
         The likelihood ratio test is performed on two nested models 
         (self is nested in other; that is, some of the parameters of other are fixed in self).
--- a/lmfit/model.py
+++ b/lmfit/model.py
@@ -1004,3 +1004,15 @@ class ModelFit(Minimizer):
         n = self.ndata
         k = self.nvarys
         return n * np.log(rss / n) + 2 * k
+
+    def bic(self):
+        """calculate and return the BIC for a ModelFit.
+        BIC is Bayesian information criterion, a measure of the relative quality of a statistical model for a given set of data.
+        It is computed as `n * log(rss/n) + log(n) k`, where `n` is the number of data points, `rss` is the sum of residuals squares, 
+        and k in the number of varying parameters. 
+        It is closely related to AIC but gives a larger penalty to the number of paramters in the model.
+        """
+        rss = np.float((self.residual**2).sum())
+        n = self.ndata
+        k = self.nvarys
+        return n * np.log(rss / n) + np.log(n) * k
--- a/lmfit/model.py
+++ b/lmfit/model.py
@@ -993,3 +993,14 @@ class ModelFit(Minimizer):
                             fit_kws=fit_kws, ax_kws=ax_res_kws)
 
         return fig
+
+    def aic(self):
+        """calculate and return the AIC for a ModelFit.
+        AIC is Akaike information criterion, a measure of the relative quality of a statistical model for a given set of data.
+        It is computed as `n * log(rss/n) + 2k`, where `n` is the number of data points, `rss` is the sum of residuals squares, 
+        and k in the number of varying parameters.
+        """
+        rss = np.float((self.residual**2).sum())
+        n = self.ndata
+        k = self.nvarys
+        return n * np.log(rss / n) + 2 * k
--- a/lmfit/model.py
+++ b/lmfit/model.py
@@ -1000,7 +1000,7 @@ class ModelFit(Minimizer):
         It is computed as `n * log(rss/n) + 2k`, where `n` is the number of data points, `rss` is the sum of residuals squares, 
         and k in the number of varying parameters.
         """
-        rss = np.float((self.residual**2).sum())
+        rss = self.chisqr
         n = self.ndata
         k = self.nvarys
         return n * np.log(rss / n) + 2 * k
@@ -1012,7 +1012,63 @@ class ModelFit(Minimizer):
         and k in the number of varying parameters. 
         It is closely related to AIC but gives a larger penalty to the number of paramters in the model.
         """
-        rss = np.float((self.residual**2).sum())
+        rss = self.chisqr
         n = self.ndata
         k = self.nvarys
         return n * np.log(rss / n) + np.log(n) * k
+
+
+    def likelihood_ratio_test(self, other):
+        """calculate a likelihood ratio test for two model fits (self and other).
+
+        The likelihood ratio test is performed on two nested models 
+        (self is nested in other; that is, some of the parameters of other are fixed in self).
+        The test is used to calculate the probability that, given the null hypothesis that the nested model is true,
+        fitting both models to the same data produces the model fits given in the input to this function.
+        
+        The test uses the formula:
+        `D = -2 log(lik0/lik1)` 
+        where `D` is approximately chisquare distributed with `df` degrees of freedom, where `df` is the difference 
+        in number of parameters between the models. This approximation is due to Wilks's theorem and depends on the 
+        number of data points.
+
+
+        Parameters
+        ----------
+        self, other : lmfit.ModelFit
+            ModelFit instances. self is the nested /restricted / null model.
+
+        Returns
+        -------
+        pvalue : float
+            The pvalue of the test, specifying the probability of getting the specified input given the data.
+        D: float
+            The test statistic, chisquare distributed with df degrees of freedom when the number of data points is large.
+        df: int
+            The respective degrees of freedom, calculated as the difference in number of paramaters of the input models.
+        
+        Notes
+        -----
+        Model likelihood is calculated as the reciprocal of the residual sum of squares of the model.
+        Therefore the formula used is
+        `D = -2 log(rss1/rss0)`
+
+        See Also
+        --------
+        http://en.wikipedia.org/wiki/Likelihood-ratio_test
+
+        """
+        n0 = self.ndata
+        k0 = self.nvarys
+        chisqr0 = self.chisqr # residual sum square - see minimizer.py
+        assert chisqr0 > 0
+        n1 = other.ndata 
+        k1 = other.nvarys
+        chisqr1 = other.chisqr
+        assert chisqr1 > 0
+
+        lam = chisqr1 / chisqr0 # chisqr ~ 1/lik
+        assert 0 <= lam <= 1
+        D = -2 * np.log(lam)
+        df = k1 - k0
+        return stats.chi2.sf(D, df), D, df
--- a/tests/test_model.py
+++ b/tests/test_model.py
@@ -160,6 +160,22 @@ class CommonTests(object):
         # to be overridden for models that do not accept indep vars
         pass
 
+    def test_aic(self):
+        model = self.model
+
+        # Pass Parameters object.
+        params = model.make_params(**self.guess())
+        result = model.fit(self.data, params, x=self.x)
+        aic    = result.aic()
+        self.assertTrue(aic < 0) # aic must be negative
+
+        # Pass extra unused Parameter.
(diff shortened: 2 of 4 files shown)


lmfit/model.py
```diff
@@ -993,3 +993,14 @@ def plot(self, datafmt='o', fitfmt='-', initfmt='--', numpoints=None,
                             fit_kws=fit_kws, ax_kws=ax_res_kws)
 
         return fig
+
+    def aic(self):
+        """calculate and return the AIC for a ModelFit.
+        AIC is Akaike information criterion, a measure of the relative quality of a statistical model for a given set of data.
+        It is computed as `n * log(rss/n) + 2k`, where `n` is the number of data points, `rss` is the sum of residuals squares, 
+        and k in the number of varying parameters.
```
tacaswell: docstrings should use numpydoc format and should be wrapped to < 80 columns.

tests/test_model.py
```diff
@@ -160,6 +160,22 @@ def check_skip_independent_vars(self):
...
+
+        # Pass Parameters object.
+        params = model.make_params(**self.guess())
+        result = model.fit(self.data, params, x=self.x)
+        aic    = result.aic()
+        self.assertTrue(aic < 0) # aic must be negative
+
+        # Pass extra unused Parameter.
+        params.add("unused_param", value=1.0, vary=True)
+        result    = model.fit(self.data, params, x=self.x)
```
tacaswell: nit: too much white space before '='

lmfit/model.py
tacaswell: This should probably also include a citation.

danielballan: For your reference, @yoavram, the [numpy doc standard](https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt). I don't think the entire lmfit codebase necessarily adheres, but the newer parts (model.py, fitter.py) do, and it's work in progress.

newville:
@yoavram Thanks!  My first thought was "what the heck is AIC?", which not only demonstrates my own ignorance, but also makes me suspect that many lmfit users would also be unfamiliar with this.  (us poor physicists, trained to arrogantly hold the world can be modeled mathematically and that statistics are there because the measurements are imperfect, not our understanding of nature).   That's OK -- raising the general level of analysis is great!  (and, again, Thanks!!)  -- but suggests that an explanation in the doc (as well as the docstring) is probably necessary.      

I'd also ask what other statistical tests might be included?  I think it's perfectly reasonable to say "well, this is that someone wanted badly enough to provide the code to do it",  but are there any obvious other tests that should be included?

I agree with @tacaswell and @danielballan on cleaning up the doc strings and code.

Anyway, Thanks!

yoavram:
- More tests of the `aic` method: a more sophisticed test can fit polynomials with different degrees to data and use the AIC to select the degree has the best balance between likelihood (n terms of residuals) and complexity (in terms of number of degrees).
- Cleaning up the doc strings - no problem.
- Cleaning up the code - just removing the spaces, you mean? I like it when consecutive equal signs are above each other but I have no problem to change it.
- Where should I add an explanation in the code? Is the docstring enough as it is? Is it OK if I refer to [R's AIC docs](https://stat.ethz.ch/R-manual/R-patched/library/stats/html/extractAIC.html) for more information?
- Other statistical tests - the Generalized Likelihood Ratio tests nested models (good for when you use the same model but fix or bound more and more parameters) and there are other ways to select and validate models - see this [great tutorial](http://nbviewer.ipython.org/github/fonnesbeck/Bios366/blob/master/notebooks/Section6_3-Model-Selection-and-Validation.ipynb) by @fonnesbeck which also includes a section on AIC. 

I'm not an expert on this subject but I'm doing a project in which I use lmfit to fit data to a model and I want to select between different "versions" of this model, each having a different number of parameters (6,4,2,1). Using AIC is a quick and simple way to do it and, from what I understand, althoguh there could be some statistical gotchas and "hidden" assumptions.

danielballan:
Like @newville, my physics background did not include AIC. Is it under the purview of the statsmodels package? Is [this page of the statsmodels docs](http://statsmodels.sourceforge.net/0.6.0/generated/statsmodels.tools.eval_measures.aic.html) addressing the same feature as your PR?

It was your mention of R that brought this to my mind. Statsmodels and [pandas](http://pandas.pydata.org/) together aim to offer the features of R in Python.

tritemio:
Another related related criterion is [BIC](http://en.wikipedia.org/wiki/Bayesian_information_criterion). AIC and BIC are commonly used to rank models with increasing number of parameters. Since with more parameters the model always fit better the data (overfitting) these criteria penalize models with more parameters to find a trade off, and thus an "optimal" model. However the "optimal" model is not guaranteed to be a "good" model, yet alone the "true" model.

I think AIC and BIC are commonly used because are very simple to implement, but their validity for model selection must be considered with a grain of salt.

Said that, would be great (and quite natural) to have these statistics in lmfit.

yoavram:
@danielballan the statsmodels ref is correct, however I wouldn't go as far as saying that pandas and statsmodels aim to offer the features of R in python - R is such a rich language and it would take more work to replicate it's capabilities.
I don't think AIC should be regarded as out of scope of lmfit or to require people to use the statsmodels package because AIC is so easy to calculate and is a standard way to select between models in many areas (see for example [ecology and evolution](http://www.sciencedirect.com/science/article/pii/S0169534703003458). The lmfit code already produces all the data, my method simply transforms the residual array to a single number, weighted by the number of observations and parameters.

danielballan:
Great. Then I'm on board. I just wanted to include statsmodels in the discussion. Agreed that R is more mature and feature-complete.

newville:
@yoavram -  Thanks again.  I am definitely +1 on having this in lmfit.  

I would prefer to see something added to the (rst) docs about this, even if mostly taken from the doc-strings (even with autodoc) and adding the sort of explanation and links  you've given in this discussion.   That is, make it so the new user knows how to use these in both "how do I run this function" and "how do I interpret and use the results" senses.

Also: Is it worth adding a BIC method as well?

Are you up for these?

yoavram:
I was just having a discussion with @uriobolski about using BIC and how it it different from AIC, so I plan to use it anyway - in that case I will add it as well.
So I'm up to this, I'll probably drag it for a little while (multitasking) but I'll do it.

andyfaff:
Perhaps I think the methods should be properties instead, there are no arguments. The values are actually properties of the system. 

    @property
    def AIC(self):
        # return AIC value.

lmfit/model.py
```diff
@@ -993,3 +993,26 @@ def plot(self, datafmt='o', fitfmt='-', initfmt='--', numpoints=None,
                             fit_kws=fit_kws, ax_kws=ax_res_kws)
 
         return fig
+
+    def aic(self):
```
andyfaff: This would be better as a property:

    @property
    def aic(self):
        # return AIC

In that way you get the AIC value as `model.aic` rather than `model.aic()`

lmfit/model.py
```diff
@@ -993,3 +993,26 @@ def plot(self, datafmt='o', fitfmt='-', initfmt='--', numpoints=None,
...
+        AIC is Akaike information criterion, a measure of the relative quality of a statistical model for a given set of data.
+        It is computed as `n * log(rss/n) + 2k`, where `n` is the number of data points, `rss` is the sum of residuals squares, 
+        and k in the number of varying parameters.
+        """
+        rss = np.float((self.residual**2).sum())
+        n = self.ndata
+        k = self.nvarys
+        return n * np.log(rss / n) + 2 * k
+
+    def bic(self):
```
andyfaff: This would be better as a property:

    @property
    def bic(self):
        # return bIC

In that way you get the BIC value as `model.bic` rather than `model.bic()`

andyfaff:
If these methods/properties are added to `Model` perhaps they should also be added to `Minimizer`?

newville:
@andyfaff  -- I agree we would like these for `Minimizer`, perhaps include aic and bic in the `fit_report()` would be enough?

I'd also very much like to see documentation for these methods.

newville:
@yoavram Any progress on this effort to include BIC?  It seems mostly complete, but stalled.   Is there something that can be done to encourage this?

yoavram:
I just need to get around to writing the documentation you've requested. 
Also, I'm adding to my own project a likelihood ratio test between nested models so I'm waiting to see if that is something that can also be incorporated into lmfit.
Sorry for the delay.

newville:
@yoavram OK, thanks!  No rush.

yoavram:
I added a likelihood ratio test, still need to add tests for that, maybe also add an [F-test](http://en.wikipedia.org/wiki/Lack-of-fit_sum_of_squares) and add docs.

newville:
@yoavram  Thanks, I do appreciate the effort and would like to merge these changes.  But I'm somewhat confused by the recently addition of more (likelihood_ratio_test) when there is still not documentation for the previous parts of this PR.  Now you have even more docs to write!

Adding more code would be OK, but my preference is to have PRs focused on a single topic.  Please add docs for existing code before adding any more new code.

yoavram:
I completely understand your concern.
These are basically three versions of the same idea - model selection. I'm changing the PR name to refect this. The F-test is similar to the likelyhood ratio test, I will only add it if I see that for my usecase it has an advantage (in order to keep the code additions down to a minimum).
I figured I would do the entire docs in one go after I know exactly what's in the PR.

Is it OK if I leave this methods as they are or do you want them in Minimizer and as properties? I think that this can be done in a second PR if needed, but if you think it is required for this PR than I will make the effort.

newville:
@yoavram   Well, two points: 1) the aic and bic code is small enough we can move it around after the PR is acceptable (hint: ADD DOCUMENTATION), and 2) I think the aic and bic statistics probably belong in the Minimizer class -- these statistics are essentially properties of the fit results.  They are in the same category as chi-square, both in terms of usage as a statistic and from a "what is needed to calculate this statistic" view.   

With these in Minimizer, we can then be used in the fit reporting functions, and in Model and ModelFit.    It's reasonable to think about making a better abstraction of "calculate fit statistics".   In fact, re-organizing Minimizer() to return a "MinimizerResult" object (similar to ModelFit) is on the ToDo list anyway, and eventually have these statistics live there probably makes the most sense.  But for now, leave the code where it is.  After this PR, we can make them properties of Minimizer, and we can worry about reorganizing Minimizer() after that.

I know we all have the tendency to write code first and documentation only when forced to do so -- I admit to doing this myself.  But, I have asked for documentation from the beginning, and several times during the process.   Honestly, the AIC calculation is a couple of lines of code (as is the BIC).   I think the request for documentation is completely reasonable and the fact that you are adding code, thinking about adding more code, reorganizing code, renaming the PR, and so on instead of writing the documentation is not a good sign at all. 

This has dragged on for a while.   I want to give you credit for adding this -- I certainly never would have added these statistics myself.   So, let me be clear.  Absolutely, without question, the next thing you do is ADD DOCUMENTATION.  Do not add code, do not reorganize code.  Write the documentation.

yoavram:
Got it. I appreciate the long reply - the documentation will be the next thing. I'll just say that I've added the code because that was my internal process working on my own project; that being said I will add the docs before anything else. Thanks .

tacaswell:
:+1: to documentation before more code.

yoavram:
I've written the docs (see [here](https://yoavram.github.io/lmfit-py/model.html#the-modelfit-class)), comments or corrections are very welcome.

newville:
@yoavram Thank you.  My inclination is to merge this, but then make a few modifications.   That is,  I would 

1. move `bic` and `aic` to be calculated in the Minimizer class.
2. modify fit_report to report these values
3. replace "residuals sum of squares" with "sum of squares of the residual array" in the docs (which I think would address @tritemio comment).
4. Rename ModelFit.lik_ratio to ModelFit.likelihood_ratio.

Any objections to these tweaks?

yoavram:
@newville None at all. Would you want me to add a test of `lik_ratio`?
@tritemio the likelihoods equal the reciprocal (1/x) of the sum of the squares of the residuals up to a contstant, but because we are using the likelihood ratio we don't care about that constant. I would be happy to stand corrected.

newville:
Thanks!  Merged (by hand to resolve conflicts in doc/model.rst) to master branch.

tacaswell:
@newville One thing I have found helpful on MPL is pasting the sha1 of the hand-merge into this thread.  It really helps with after-the-fact forensics.

newville:
@tacaswell excellent suggestion.  This was 4083d57f

yoavram:
Thanks everyone, happy to be involved to learn about the process.

tritemio:
@newville @yoavram
There seem to be something wrong here.

In the specific case of curve fitting, under the assumption that the residuals are Gaussian distributed (i.i.d. random variables) the **log-likelihood** (not the likelihood) function simplifies to the sum of the squared residuals, multiplied by -1. Note that is not the inverse but a "-" sign.

When comparing nested models, [according to wikipedia](http://en.wikipedia.org/wiki/Likelihood-ratio_test#Use_With_Nested_Models), the statistics *D* used the likelihood test is 2 times the difference of the log-likelihood functions for the two models.

However the code implements this:

```
lam = chisqr1 / chisqr0 # chisqr ~ 1/lik
D = -2 * np.log(lam)
```

That looks wrong to me. It should be:

```
D = 2*(chisqr1 - chisqr0)
```

(not sure about the sign). Anyway this is all based on logic and superficial information from Wikipedia.

newville:
For me, the aic and bic are the main points for this PR.  Do we even want the likelihood ratio method here?  If so, I'd imagine we want many similar tests like Hamilton's f test and so on.  These should probably be together and maybe not in ModelFit.  So I'm ok with disabling this....

For the specifics, yeah,  I agree with you. Don't most people just compare reduced chi square as the first comparison of two models?

Tillsten:
@tritemio -log(x) = log(1/x)

yoavram:
How does "reduced chi square" work? Does it take into account different in model complexity (=number of params) like the likelihood ratio test / aic does?

newville:
Hi Yoav,

   reduced chi_square = chi_square / (ndata - nvarys)

This can be used to assess a single fit: the value should be ~1 for a good
fit .  It can also as a way of comparing two models with different number
of variables: the better value for reduced chi_square is probably the
better model.

I can certainly believe that this is a bit simplistic for a statistician,
but it's very commonly used in the physical sciences.

--Matt

tritemio:
@Tillsten sure, I know.

What I meant is that:

    1/log(x) != -log(x)

tritemio:
My point is that the code does not reflect the wikipedia definition, that is also the reference in the docstring. @yoavram did derived the form of the likelihood test you implemented? Based on which definition? Do you have references?

tritemio:
@newville I agree with you on the use of reduced chi-square. My understanding is that the likelihhod test in principle allows to compute the *probabiltiy* that one model is better than the other. But if you use it without referring to the significance levels, than it is just the difference of 2 reduced chi-square values (the difference, not the ratio).

newville:
@tritemio, @yoavram, @Tillsten Can we please move this conversation to the mailing list?  The PR is accepted and closed, and the conversation now off-topic.      

A discussion of what statistics to calculate and report is definitely worth having, and getting the code correct for these is definitely important.    But this is the wrong place for this discussion. The conversation will be lost in a closed pull request. 

It's kind of infuriating to have to keep harping on this, but Github Issues is continually being misused.    It is tempting to consider turning off Issues completely.

PR #206, opened 2015-02-16
MAINT: ravel output from minimizer._residual, model._residual. Closes #205

### LINKED ISSUE #205: Error when fitting multi-dimensional data by using Model.fit.
Model.fit cannot fit multi-dimensional data and I get an error message (error: Result from function call is not a proper array of floats).

As mentioned in [FAQ page](http://lmfit.github.io/lmfit-py/faq.html#how-can-i-fit-multi-dimensional-data), this is due to need for converting residual to a one dimensional array.
So I think `Model._residual` should call `numpy.ndarray.flatten()` before returning.
```
 class Model(object):
 ...
     def _residual(self, params, data, weights, **kwargs):
         "default residual:  (data-model)*weights"
         diff = self.eval(params, **kwargs) - data
         if weights is not None:
             diff *= weights
-        return np.asarray(diff)  # for compatibility with pandas.Series
+        return np.asarray(diff).flatten()
```

andyfaff:
ravels output from Minimizer._residual and Model._residual, so that multidimensional fits can occur without the user having to flatten the data.

### DIFF
--- a/lmfit/minimizer.py
+++ b/lmfit/minimizer.py
@@ -289,7 +289,7 @@ class Minimizer(object):
         if callable(self.iter_cb):
             self.iter_cb(self.params, self.nfev, out,
                          *self.userargs, **self.userkws)
-        return out
+        return np.asarray(out).ravel()
 
     def __jacobian(self, fvars):
         """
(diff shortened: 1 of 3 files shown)


newville:
@andyfaff -- Thanks!

PR #209, opened 2015-02-18
Test

tespilla:
Added a missing factor in expgaussian function

### DIFF
--- a/lmfit/lineshapes.py
+++ b/lmfit/lineshapes.py
@@ -113,7 +113,7 @@ def expgaussian(x, amplitude=1, center=0, sigma=1.0, gamma=1.0):
     """
     gss = gamma*sigma*sigma
     arg1 = gamma*(center +gss/2.0 - x)
-    arg2 = (center + gss - x)/s2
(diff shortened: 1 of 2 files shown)


newville:
@tespilla  Thanks!   That looks right to me.  Any objections?

lmfit/lineshapes.py
```diff
@@ -113,7 +113,7 @@ def expgaussian(x, amplitude=1, center=0, sigma=1.0, gamma=1.0):
     """
     gss = gamma*sigma*sigma
     arg1 = gamma*(center +gss/2.0 - x)
-    arg2 = (center + gss - x)/s2
+    arg2 = (center + gss - x)/s2/sigma
```
tacaswell: A _very_ nitpicky styling thing, I prefer to see `(center + gss -x) / (s2 * sigma)`  as it is does not require understanding/remembering the left-to-right precedence order of python to be sure it is right.   

This might ever be a minor performance improvement as you only do one division broad cast across the input array.

PR #213, opened 2015-02-28
change default weighting in ModelFit.plot_() routines

newville:
This PR does:

1. add a `yerr` argument to the ModelFit.plot_() methods
2. change the default for `yerr` when `weights` are used from `1/self.weights**0.5`  to `1/self.weights`.
3. change the default labels to *not* include the model name, and instead move the model name to the plot title.

I also added docs for the plot methods to the rst doc.

### DIFF
--- a/doc/model.rst
+++ b/doc/model.rst
@@ -323,6 +323,7 @@ specifying one or more independent variables.
 
    See :ref:`model_param_hints_section`.
 
+
 :class:`Model` class Attributes
 ---------------------------------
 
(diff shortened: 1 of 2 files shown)


danielballan:
Look good to me. I can't recall why it was `**0.5` in the first place. Just a mistake?

newville:
@danielballan -- yeah, I think it was just a silly mistake.  Thanks to ronb2552@gmail.com for catching this!

Issue #215, opened 2015-03-04
IPython widets FutureWarning


tritemio:
In ipython 3, importing the widgets triggers a warning. lmfit detects the presence of the notebook and imports the widgets for its fancy fitting GUI.

However this results in a (relative obscure) warning every time lmfit is imported in a notebook (even if we do no intend to use the widgets), and even more annoying every time a python library imports lmfit.

     :0: FutureWarning: IPython widgets are experimental and may change in the future.

IMHO, there should be a way to import lmfit without widgets (thus avoiding the warning), especially for libraries that use `lmfit`. Would be OK to require a more convoluted import syntax in this case (and leave the default case as is).

danielballan:
Seems like importing the notebook fitter (where widgets are used) should
require a specific separate import.

newville:
@tritemio, @danielballan  +1 on reducing implicit imports of code that might not be used.  Probably just removing https://github.com/lmfit/lmfit-py/blob/master/lmfit/__init__.py#L49 would be enough.

tritemio:
Agree.

newville:
should be resolved wih #216

PR #216, opened 2015-03-05
remove import of ui from __init__.py

newville:
this addresses #215, and may give a slight improvement in import time.


tacaswell:
:+1:

danielballan:
Do the docs assume Fitter is already imported?

newville:
@danielballan I don't think there is really much documentation on Fitter.  What there is doesn't show an import....

PR #220, opened 2015-03-19
Added __division __ to lineshapes, fixes a bug with expgaussian.

Tillsten:
One could also change the `/2` to `/2.`, but this is better for py3.


Tillsten:
Also fixes a division bug with the shape below. But expgaussian still behaves strange, but that could be the parameterization.

newville:
Thanks!  I agree this is a good idea.

Issue #221, opened 2015-03-20
Runtime error: TypeError: context must be specified


tritemio:
In current master (b8e8a10e83156ff45493975c97a37af0a7f7699e), I get this error from an ipython notebook that works with 0.8.3.

I just dump the trace for now, I'll came back to it later:

```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-33-d9213941ff62> in <module>()
----> 1 ds.E_fitter.fit_histogram(mfit.factory_three_gaussians())

C:\Users\laser2002j\Anaconda\lib\site-packages\fretbursts-0.4rc9_9_g958ffd2-py2.7-win-amd64.egg\fretbursts\mfit.pyc in fit_histogram(self, model, pdf, **fit_kwargs)
    427         for i, data in enumerate(data_list):
    428             self.fit_res.append(
--> 429                 self.model.fit(data, x=self.hist_axis, **fit_kwargs)
    430                 )
    431             self.params.iloc[i] = pd.Series(self.fit_res[-1].values)

C:\Users\laser2002j\Anaconda\lib\site-packages\lmfit-0.8.3_54_gb8e8a10-py2.7.egg\lmfit\model.pyc in fit(self, data, params, weights, method, iter_cb, scale_covar, verbose, fit_kws, **kwargs)
    556 
    557         output = ModelFit(self, params, method=method, iter_cb=iter_cb,
--> 558                           scale_covar=scale_covar, fcn_kws=kwargs, **fit_kws)
    559         output.fit(data=data, weights=weights)
    560         output.components = self.components

C:\Users\laser2002j\Anaconda\lib\site-packages\lmfit-0.8.3_54_gb8e8a10-py2.7.egg\lmfit\model.pyc in __init__(self, model, params, data, weights, method, fcn_args, fcn_kws, iter_cb, scale_covar, **fit_kws)
    741         self.weights = weights
    742         self.method = method
--> 743         self.init_params = deepcopy(params)
    744         Minimizer.__init__(self, model._residual, params, fcn_args=fcn_args,
    745                            fcn_kws=fcn_kws, iter_cb=iter_cb,

C:\Users\laser2002j\Anaconda\lib\copy.pyc in deepcopy(x, memo, _nil)
    172             copier = getattr(x, "__deepcopy__", None)
    173             if copier:
--> 174                 y = copier(memo)
    175             else:
    176                 reductor = dispatch_table.get(cls)

[...]

PR #224, opened 2015-04-01
Limit deepcopy

newville:
This *should* address #221, but I think it will need careful testing.

In this approach, asteval keeps a list of items to *not* try to deepcopy() (stuff it will be able to import from Python or numpy later), and Parameters.__deepcopy__() uses this list.

### DIFF
--- a/lmfit/parameter.py
+++ b/lmfit/parameter.py
@@ -45,7 +45,11 @@ class Parameters(OrderedDict):
     def __deepcopy__(self, memo):
         _pars = Parameters()
         for key, val in self._asteval.symtable.items():
-            _pars._asteval.symtable[key] = deepcopy(val)
+            if key not in self._asteval.no_deepcopy:
+                try:
+                    _pars._asteval.symtable[key] = deepcopy(val, memo)
+                except TypeError:
+                    pass
         for key, par in self.items():
             if isinstance(par, Parameter):
                 name = par.name
--- a/lmfit/parameter.py
+++ b/lmfit/parameter.py
@@ -46,10 +46,7 @@ class Parameters(OrderedDict):
         _pars = Parameters()
         for key, val in self._asteval.symtable.items():
             if key not in self._asteval.no_deepcopy:
-                try:
-                    _pars._asteval.symtable[key] = deepcopy(val, memo)
-                except TypeError:
-                    pass
+                _pars._asteval.symtable[key] = deepcopy(val, memo)
         for key, par in self.items():
             if isinstance(par, Parameter):
                 name = par.name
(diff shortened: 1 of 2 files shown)


newville:
@tritemio @yoavram does this branch work for you?

yoavram:
I tested - no error with this branch in IPython notebook.

tritemio:
Tested, no errors.

lmfit/parameter.py
```diff
@@ -45,7 +45,11 @@ def __init__(self, asteval=None, *args, **kwds):
     def __deepcopy__(self, memo):
         _pars = Parameters()
         for key, val in self._asteval.symtable.items():
-            _pars._asteval.symtable[key] = deepcopy(val)
+            if key not in self._asteval.no_deepcopy:
+                try:
+                    _pars._asteval.symtable[key] = deepcopy(val, memo)
+                except TypeError:
+                    pass
```
tritemio: Since there is already the blacklist no_deepcopy, is the try-except needed?

newville:
@tritemio   I agree and removed the try / except TypeError -- any such remaining errors can be raised.  

I don't really love this solution, as it's really checking names of symbols to not deepcopy, regardless of the actual values.  OTOH, monkey-patching is unavoidable, and this solves the immediate problem.    Also, the asteval symbol table comes pre-loaded with many functions from numpy and python, so there's really no need to copy these symbols from one instance to another (unless they've been over-written).  So, I'm willing to say the previous behavior wasn't ideal and that this behavior isn't ideal either.   From a practical point of view, this seems to work better. 

Offhand, does anyone know how we might include "running all the tests from IPython" in the testing framework?

yoavram:
[pytest-ipynb](https://pypi.python.org/pypi/pytest-ipynb/0.1.1)? But I think that would require to write the tests again as notebooks. [This SoF](http://stackoverflow.com/a/23292713/1063612) answer explains how to convert `py` to `ipynb` so I think that in principle one can convert all the tests to notebook and run them. I never tried either..

tritemio:
Or we could have a simple notebook that run the tests programmatically from python similarly to this:

http://nbviewer.ipython.org/github/tritemio/FRETBursts/blob/master/notebooks/dev/tests/FRETBursts%20-%20Unit%20tests-debug.ipynb

and then execute the notebook from the command line.

The added benefit is that when a test fails the notebook gives you a live session to debug the error interactively (instead of having to recreate the error outside the testing framework).

Also, other peoples are working on the integration of notebook in tests frameworks and documentation. See this thread and following answers:

http://mail.scipy.org/pipermail/ipython-dev/2015-March/016177.html

We may have something more "official" in the following months. Focusing on supporting only ipython >= 3 would simplify things as the machinery to run notebooks from command line is already integrated.

newville:
@yoavram @tritemio   Thanks!   Ideally, we'd be able to run all tests with both Python and IPython as part of the standard test-suite, including with TravisCI.   Converting all the existing tests to IPython notebooks and running those sounds like a promising direction to pursue, as it could allow automating the tests.

PR #226, opened 2015-04-26
[WIP] Add custom target function for scalar_minimize

Tillsten:
This change adds an likelihood_fun kwarg to minimizer.scalar_minimize and minimizer.penalty function. If given, instead of minimizing `sum((residual)**2)` it will minimize likelihood_fun(residual) instead.
Note that the function is expected to return a scalar. 

Reason:
This allows for using a more robust likelihood (e.g. t-likelihood) than the normal distribution, which can be helpful if one has problems with outliers. This is also useful if the distrubtion of the data has a non-normal distrubtion, e.g. poisson for low photon counts. 

This is missing documention, also i am not to happy about the naming. Comments?

### DIFF
--- a/examples/example_stable.py
+++ b/examples/example_stable.py
@@ -1,0 +1,54 @@
+# -*- coding: utf-8 -*-
+"""
+Example using the t-likelihood for stable fitting.
+"""
+import numpy as np
+import scipy.special as sp
+import scipy.stats as stats
+import lmfit
+try:
+    import matplotlib.pyplot as plt
+    HAS_PYLAB = True
+except ImportError:
+    HAS_PYLAB = False
+
+np.random.seed(1)
+x = np.linspace(-5, 5, 30)
+
+# Setup example
+amp_erf = 5
+w = 0.3
+c = 1
+amp_sin = 2
+omega = 3
+
+y = amp_erf*sp.erf(x/w)+2*np.sin(x*omega)+c
+yn = y + np.random.randn(*y.shape)*1.5
+outliers = np.random.random_integers(0, x.size-1, x.size/5)
+yn[outliers] += np.random.randn(outliers.size)*10
+
+if HAS_PYLAB:
+    plt.style.use('fivethirtyeight')
+    plt.figure(figsize=(8, 5))
+    plt.plot(x, y, lw=3)
+    plt.plot(x, yn, 'o')
+
+def func(x, amp_erf, w, c, omega, amp_sin):
+    y_model = amp_erf*sp.erf(x/w) + amp_sin* np.sin(x*omega) + c
+    return y_model
+
+mod = lmfit.model.Model(func, ['x'])
+res = mod.fit(yn, x=x, w=1, c=1, omega=3, amp_sin=2, amp_erf=3)
+
+def logln(x):
+    "Returning the t-log-likehood of x with df=3"
+    return -stats.t.logpdf(x, df=2).sum()
+
+o = res.scalar_minimize(method='L-BFGS-B', likelihood_fun=logln)
+
+if HAS_PYLAB:
+    plt.plot(x, res.best_fit, 'k', lw=3)
+    plt.plot(x, res.eval(), 'r', lw=3)
+    plt.legend(['True function', 'w. noise+outliers',
+                'Fit', 'Robust Fit'], loc='upper left')
+
--- a/lmfit/minimizer.py
+++ b/lmfit/minimizer.py
@@ -278,7 +278,7 @@ class Minimizer(object):
         else:
             raise MinimizerException(self.err_nonparam)
 
-    def penalty(self, params):
+    def penalty(self, params, likelihood_fun=None):
         """
         Penalty function for scalar minimizers:
 
@@ -294,8 +294,10 @@ class Minimizer(object):
             objective function is an array, return the array sum-of-squares
         """
         r = self.__residual(params)
-        if isinstance(r, ndarray):
+        if isinstance(r, ndarray) and likelihood_fun is None:
             r = (r*r).sum()
+        elif isinstance(r, ndarray):
+            r = likelihood_fun(r)
         return r
 
     def prepare_fit(self, params=None):
@@ -407,7 +409,7 @@ class Minimizer(object):
         self.unprepare_fit()
         return
 
-    def scalar_minimize(self, method='Nelder-Mead', **kws):
+    def scalar_minimize(self, method='Nelder-Mead', likelihood_fun=None, **kws):
         """
         Use one of the scalar minimization methods from
         scipy.optimize.minimize.
@@ -456,7 +458,8 @@ class Minimizer(object):
                         options={'maxiter': 1000 * (self.nvarys + 1)})
         fmin_kws.update(self.kws)
         fmin_kws.update(kws)
-
+        penalty_func = lambda x: self.penalty(x, likelihood_fun)
+                
         # hess supported only in some methods
         if 'hess' in fmin_kws and method not in ('Newton-CG',
                                                  'dogleg', 'trust-ncg'):
@@ -485,9 +488,9 @@ class Minimizer(object):
             # in scipy 0.14 this can be called directly from scipy_minimize
             # When minimum scipy is 0.14 the following line and the else
             # can be removed.
-            ret = _differential_evolution(self.penalty, self.vars, **fmin_kws)
+            ret = _differential_evolution(penalty_func, self.vars, **fmin_kws)
         else:
-            ret = scipy_minimize(self.penalty, self.vars, **fmin_kws)
+            ret = scipy_minimize(penalty_func, self.vars, **fmin_kws)
 
         xout = ret.x
         self.message = ret.message


Tillsten:
Here the image of the included example to illustrate why this is useful:

img

lmfit/minimizer.py
```diff
@@ -294,8 +294,10 @@ def penalty(self, params):
             objective function is an array, return the array sum-of-squares
         """
         r = self.__residual(params)
-        if isinstance(r, ndarray):
+        if isinstance(r, ndarray) and likelihood_fun is None:
```
tacaswell: Maybe re-write this as

```python

if not isinstance(r, ndarray):
    return r

if likelyhood_fun is not None:
    return likelyhood_fun(r)
else:
    return np.sum(r*r)
```

examples/example_stable.py
```diff
@@ -0,0 +1,54 @@
...
+amp_sin = 2
+omega = 3
+
+y = amp_erf*sp.erf(x/w)+2*np.sin(x*omega)+c
+yn = y + np.random.randn(*y.shape)*1.5
+outliers = np.random.random_integers(0, x.size-1, x.size/5)
+yn[outliers] += np.random.randn(outliers.size)*10
+
+if HAS_PYLAB:
+    plt.style.use('fivethirtyeight')
```
tacaswell: nit-pick

```python
if HAS_MPL:
    with plt.style.context('figevthirtyeight'):
        fig, ax = plt.subplots()
        ax.plot(x, y, lw=3)
        ax.plot(x, yn, 'o')
```

and just use `ax.plot` and `ax.legend` in the second block.

Tillsten:
@tacaswell Thanks, will do.

 I still don't like my naming, maybe use `custom_lnfunc` or something else? Also
should we expand scalar_minize as i had done or just add another method?

andyfaff:
I'm -1 on this PR.  It's already possible to achieve this,  if the `userfcn` returns a scalar. Of course, if the `userfcn` returns an array then chi2 is automatically used. 
`userfcn` is already a penalty function, so with this PR we would be minimising the penalty function of a penalty function, which is adding another layer of complication.

I also note that there aren't any tests.

Tillsten:
@andyfaff Thanks for your feedback! 

At the moment there is no easy way to go from a data-modeling function to a likelihood. This would allow to use the same function for modling without rewriting, which i preferer to the current alternatives. Also the complication is very small, just two kwargs to methods.

 For test and docs, i will do that if there is a consensus to add the feature.

andyfaff:
On 27 April 2015 at 10:17, Till Stensitzki <notifications@github.com> wrote:

> At the moment there is no easy way to go from a data-modeling function to
> a likelihood.
>
I feel your pain. At the moment everything does have to be done in
`userfcn`.
If you only want to write one data modelling function then you have to have
a switch in your `userfcn` to swap between chi2 and likelihood.  This can
be accomplished by using an entry in the `fcn_kws` dictionary when the
`lmfit.Minimizer` is constructed.The `userfcn` could look like the
following pseudocode:

    def userfcn(params, *fcn_args, **fcn_kws):
            model_func = fcn_kws['model']
            model = model_func(params)

            ydata = fcn_args[0]
            resid = model - ydata
            if 'likelihood' in fcn_kws and fcn_kws['likelihood']:
                 return fcn_kws['likelihood'](resid)
            else:
                 return resid

lmfit/minimizer.py
newville: I'd recommend
```python
if isinstance(r, ndarray):
    if likelihood_fun is None: # use sum of squares as default
        likelhood_fun = lambda x: return (x*x).sum()
    r = likelihood_fun(r)
return r
```

newville:
Thanks @Tillsten    I'm +1 on this -- I like the idea of being able to easily switch between least-squares and a custom likelihood function with just a change to a ``minimize()`` argument, without having to change this in the objective function. 

If the objective function returns a scalar value, this "reduce residual array to some penalty scalar" isn't done at all, so existing code that calculates some likelihood value will work.   

Would be worth considering allowing the 'likelihood_function" (I think I prefer this name to "custom_lnfunc", but could be persuaded to call it anything) to be `None`, a callable, or one of a few names of  pre-defined, commonly used likelihood functions.   Perhaps

```python
      minimize(resid, ....., likelihood_function='least-squares')   # (r*r).sum()
      minimize(resid, ....., likelihood_function='log-likelihood')   #  logpdf(r).sum()
      minimize(resid, ....., likelihood_function='max-entropy')   #  (r*log(r)).sum()
```
Maybe that's more trouble than worth, but then again maybe it's just a documentation issue....

Anyway, I think this will be very useful.

lmfit/minimizer.py
tacaswell: As I understand it, this function gets called many times during the optimization?  Probably worth bench marking these method, I was worried about creating a new lambda every time through.

Tillsten: I seriously doubt that there are many use-cases where the overhead of lmfit plays a role, buf if proven otherwise i will revisit this.

Tillsten:
The reason why i dislike likelihood_function is because on should pass a objective function which should be minimized, and i somewhat doubt that someone what to minimize the likelihood  :).

newville:
@Tillsten   Ah, very good point.    The question (and Issue title) is really about how to reduce a residual array to a scalar for the scalar minimizers.  Maybe the keyword to `minimize()` should be something like `array2scalar` or `reduce` (though the sense is slightly different than Python's builtin) or something along those lines.... perhaps:

```python
 minimize(residual, ...., array2scalar='sum-of-squares')
 minimize(residual, ...., array2scalar='neg-log-likelihood')
```
That might still give the hint that the objective function should calculate something to be minimized.

The value could either be one of a few known string (in which case could have actual methods, not lambdas created per iteration), or a user-supplied callable.

Hopefully someone can think of a better name than `array2scalar`.

tritemio:
+1 for using reduce as keyword

Tillsten:
An alternative would be a function which takes a minimizer (or model) and returns one with the requested reduction fuction as an objective. This approach would have the advange that one could easily add parameters in this process, which is practial for MLE (e.g. adding the width of the distrubtion).

newville:
@Tillsten that's a really interesting suggestion.   I wonder if that might be handled similarly to the custom binary operation for convolution of 2 models?  That might need some thought (at least by me -- maybe you have it worked out already).

.

Tillsten:
@newville I will give it a run over weekend, i should have some code i can reuse. I agree with the concern of code-complexity. The minimizer object is already quite big and only made more complicated if it is really necessary.

newville:
@Tillsten  What's the status of this PR?  Should it be revived and updated to the current master?   I think even the simpler form -- adding something like a `reduce='neg-log-likliehood'` argument or something like it to `minimize()` -- would be very useful.

A richer approach of providing a callable would be interesting (and could, at least in principle, support arguments if using a partial function) but could wait if it's too much work.

Tillsten:
Sorry, i almost have no time at the moment. I actually tried to implent it on the Model-system, but i actually failed. Tbh i think the model code is quite complicated for what it does, but there are probly good reason for it. You are right, i will give the simpler version a try to get it included.

newville:
@Tillsten OK, that's understandable.  Anybody else have time to work on this, or even an opinion on whether we should (try to?) do this before working on a 0.9 release? 

I think this is completely separate from `Model` (which might be complicated, but does a lot of complex things to make it look easy), and that adding a "reduce" keyword to `scalar_minimize()` would be easy to propagate to `Model.fit()`.

andyfaff:
Just revisiting this PR in the light of #249, which proposes a Markov Chain Monte Carlo method for sampling the posterior probability distribution function. There the negative log-likelihood (NLL) is one of the terms that adds into the posterior PDF (the other being the log-prior probability).
The NLL has to be calculated tens of thousands of times there.  To do this one can use parallelisation, the default for which is multiprocessing. To use multiprocessing the object passed to Pool.map has to be pickleable. In that case lambda functions are not a good idea.

This PR and #249 should both consider each others needs w.r.t log-likelihood functionality.

andyfaff:
Nested functions can't be pickled either (making multiprocessing difficult)

newville:
@andyfaff   The PR here was about allowing custom ways to reduce a residual array returned by an objective function to a scalar value for the scalar minimizers.  Currently "sum-of-squares" is hard-wired in, and the aim here is to allow other ways to express a measure of "unlikelihood" that can be reduced. 

But the idea that log-likelihood is one of the possible ways to do that seems different to the aim of #249, no?  Or maybe I'm not understanding.

I think taking log-likelihood itself may not really need to be parallelized -- yeah, not creating a `lambda` function at each iteration is probably wise.   But, what needs parallelization is running the users objective function with different parameter values.   If one expects multiprocessing to handle whatever code the user writes in an objective function, the use of Pickle is a very serious limitation.  It might be fatal.  This is why I keep suggesting dill might a viable approach.

yoavram:
+1 for this PR, I really need a feature like this. 
The use-case is that I want to add a [regularization](https://en.wikipedia.org/wiki/Tikhonov_regularization) term to the fit, so something like:
```
def penalty(self, fvars):
        r = self.__residual(fvars)
        if isinstance(r, np.ndarray):
            r = r.sum()
        r += lam * np.linalg.norm([v.value for v in fvars])
        return r
```

As for pickling lamdba functions, [dill](http://pythonhosted.org/dill/) can pickle those:
```
>>> f = lambda x: x
>>> import pickle
>>> pickle.dumps(f)
_pickle.PicklingError: Can't pickle <function <lambda> at 0x00000000045F7488>: attribute lookup <lambda> on __main__ failed
>>> import dill
>>> dill.dumps(f)
b'\x80\x03cdill.dill\n_create_function\nq\x00(cdill.dill\n_unmarshal\nq\x01CN\xe3\x01\x00\x00\x00\x00\x00\x00\x00\x01\x00\x00\x00\x01\x00\x00\x00C\x00\x00\x00s\x04\x00\x00\x00|\x00\x00S)\x01N\xa9\x00)\x01\xda\x01xr\x01\x00\x00\x00r\x01\x00\x00\x00\xfa\x07<stdin>\xda\x08<lambda>\x01\x00\x00\x00s\x00\x00\x00\x00q\x02\x85q\x03Rq\x04c__builtin__\n__main__\nX\x08\x00\x00\x00<lambda>q\x05NN}q\x06tq\x07Rq\x08.'
>>> dill.loads(dill.dumps(f))(5)
5
```

newville:
@yoavram I agree that this would be a valuable addition.  I think it would not be a lot of effort to finish this off and get it to a working state, but that nobody has found the time to finish it.    Are you willing and able to finish this up?

newville:
@Tillsten this branch has some really useful code, but it hasn't been touched in well over a year.     I'd much rather see the proposed improvements added to the master branch, but it seems like it might require a fresh take from the current code base.   Can we either revive this code, bringing it up to date, or close the PR?

newville:
replaced by #387

Issue #197, opened 2015-01-21
Why are the models stateless?


Luindil:
Hi all,

i am currently writing an open source peak fitting software based on lmfit, pyqtgraph, numpy, spicy and pyqt4. The repository is https://github.com/Luindil/PeakThis. The idea is to use the model interface lmfit provides and add as many models to the graph by click defining the parameters (e.g. for a gaussian first click defines height and position, second click defines fwhm/sigma).

In the process of writing this software i find it unnecessary hard to work with the stateless models and to always use the parameter objects, especially for the resulting compound objects. I kind of have to always work around this implementation and try to keep the parameters in the models (i subclassed them).

In my opinion this stateless approach is not what i would like the models to behave. The use-cases i imagine are always including the parameter state saved in the model and then for example just calling eval(x) gives me the y-values. In any case the parameters are actually saved all the time as param hints in the model. So why not just use them? I think that would make the interface much easier to work with and then there is also no need for prefixes for same class model objects in composite models, because the composite model should know where which param was written too. In any case the parameter sigma/fwhm itself does not really make sense on its own since you always need also the peak shape to it, which is not saved in the parameter objects.
If somebody needs a specific compound model as the same starting model for several files fitting, it could just be created by deep-copying the starting model.

Can somebody explain me the rational and use-case for the current implementation of the model class? Why is there a need for the parameter objects? 

Best wishes
Clemens

danielballan:
Take a look at #56 and #57 for the genesis of `Model`.

You have described a situation where it indeed would be convenient to recall the previous fit. On the other hand, when a user is applying the same `Model` to many different data sets, it may be better for the `Model` to be "fresh" and identical each time. To address these two needs, we have added a `Fitter` class, a stateful elaboration of `Model.`

This layered approach to the design of lmfit allows users to choose the tool that is only as "smart" as they want it to be. We have avoiding packing *all* of this logic into the `Model` layer.

Take a look at `Fitter` and see if it fits your needs. There is also `NotebookFitter` and `MPLFitter`, which add a GUI, but the basic `Fitter` can be considered just a stateful `Model`.

As for the `Parameters` class: they let you specify bounds, and they can be held fixed in a convenient way. These useful features are not available in the underlying scipy routine.

Luindil:
Hei Dan, 

thank you for the response, i haven't seen the Fitter classes yet. I basically wrote the same one my own with some extra stuff...

[...]

PR #230, opened 2015-05-17
Add MinimizerResult

newville:
This is a fairly large PR, altering the Minimizer class so that:
 
1. The `params` passed in a `Minimizer.__init__()` are not changed.
2. The fit (either by the `minimize` function or the `Minimizer.minimize()`, `Minimize.leastsq()`, or `Minimize.scalar_minimizer` methods return a newly defined `MinimizerResult` object.  
3.  `MinimizerResult` contains fit information (status, messages), statistics (chi-square, aic, etc), and the `params` that were actually updated in the fit.   It's basically an empty container class that holds all the fit results. 

An important note is that in this scheme a new `params` -- a copy -- is made at the beginning of each fit. 

I believe this more or less addresses issues #174, #183, and begins to address #197.   However, this changes the API in many subtle ways.  Many tests, examples, and docs had to be changed.  For example, the interface for the calculation of confidence intervals had to be changed -- it now needs *both* a `Minimizer` object (to do subsequent fits) and a `MinimizerResult` (as a starting point for its calculations). 

All tests pass, though some needed tweaking.   The docs for this version are currently at http://cars9.uchicago.edu/software/python/lmfit_MinimizerResult/

I'm not sure this is perfect, but it does seem like an improvement to have `Minimizer` remain unchanged by a fit, and have all fit-dependent results in the returned value.   Since this is a fairly big change, and will definitely break some existing scripts (though the fixes are generally easy) this definitely needs a  boost in version number, and consensus on whether this is the right design.

This does not change the behavior of the `Model` class (and so, it is still stateless -- I toyed with this, but the PR is large enough as it is).   If we think this `MinimizerResult` is worth keeping, that seems like an important next topic to re-consider.   I don't have a strong opinion on whether `Model` should be changed, actually. 

This was a fairly long time in development, and I am in no rush to merge this without ample discussion.  So, if you have time, please read and try out the examples, and see if this is worth merging.


newville:
Any comments on this PR?   I think it's an improvement, but it will definitely break existing scripts (easy to fix, but not pain-free).  I'd greatly prefer some feedback on this, but it's been a month without any comments, and I think that silence has to be interpreted as "no objections".   Unless there are concerns, I'll plan on merging this in two weeks (June 28).

tacaswell:
I am :+1: on this in principle.  I skimmed the changes, but I don't know enough of the existing code base to have sensible comments on a short time scale.

gbouvignies:
If the minimization gets interrupted, e.g. KeyboardInterrupt, will it be possible to obtain 'params' values at the current iteration? Right now, as 'params' gets continuously updated, it is pretty straightforward to get the information.

newville:
@tacaswell Thanks! 

@gbouvignies Good question, and this probably needs documenting.  Right now a  `MinizerResult``  is created just prior to each fit (in ``prepare_fit()``), and held in ``self.result``, which is used as internal state during the fit.   At the end of a fit, this is the return value.  So this object (including ``result.params``) will hold the "active parameters", and are available after an interrupted fit.   Of course, this ``result`` is guaranteed to be over-written for each fit.

Issue #232, opened 2015-05-27
NotebookFitter and ipython 3.1


tritemio:
Seems like the NotebookFitter() is broken on current ipython notebook:

```
%matplotlib inline
import lmfit
import numpy as np
data = np.array([1,2,3,4,5,4,3,2])
x = np.arange(len(data))

fitter = lmfit.Fitter(data, x=x)
```

Raises this error:

```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-3-307f10613a3c> in <module>()
      5 x = np.arange(len(data))
      6 
----> 7 fitter = lmfit.Fitter(data, x=x)

C:\Users\laser2002j\Anaconda\lib\site-packages\lmfit-0.8.3-py2.7.egg\lmfit\ui\ipy_fitter.pyc in __init__(self, data, model, all_models, axes_style, data_style, init_style, best_style, **kwargs)
    217         super(NotebookFitter, self).__init__(data, model, axes_style,
    218                                              data_style, init_style,
--> 219                                              best_style, **kwargs)
    220 
    221     def _repr_html_(self):

C:\Users\laser2002j\Anaconda\lib\site-packages\lmfit-0.8.3-py2.7.egg\lmfit\ui\basefitter.pyc in __init__(self, data, model, axes_style, data_style, init_style, best_style, **kwargs)
    230         self.init_style = init_style
    231         self.best_style = best_style
--> 232         super(MPLFitter, self).__init__(data, model, **kwargs)
    233 
    234     def plot(self, axes_style={}, data_style={}, init_style={}, best_style={},

C:\Users\laser2002j\Anaconda\lib\site-packages\lmfit-0.8.3-py2.7.egg\lmfit\ui\basefitter.pyc in __init__(self, data, model, **kwargs)
     65         if model is None:
     66             model = ExponentialModel
---> 67         self.model = model
     68 
     69     def _on_model_value_change(self, name, value):

C:\Users\laser2002j\Anaconda\lib\site-packages\lmfit-0.8.3-py2.7.egg\lmfit\ui\basefitter.pyc in model(self, value)
    102         self.namefinder = NameFinder()
    103 
--> 104         self._finalize_model(value)
    105 
    106         self.guess()

C:\Users\laser2002j\Anaconda\lib\site-packages\lmfit-0.8.3-py2.7.egg\lmfit\ui\ipy_fitter.pyc in _finalize_model(self, value)
    240             for pw in self.param_widgets:
[...]

PR #236, opened 2015-07-05
Fixing Ipython Notebook widgets for Py3

MerlinSmiles:
This goes for issue  #232, how do I label that so they get connected?

This Pull requests fixes the broken widgets for the newest IPython Notebook 
Both py3 and py2 seem to work
There are still issues with overlapping widgets though, i have no idea on how to arrange them properly, seems they only have a width attribute...

### DIFF
--- a/lmfit/ui/ipy_fitter.py
+++ b/lmfit/ui/ipy_fitter.py
@@ -22,7 +22,7 @@ else:
     # as of IPython 3.x:
     from IPython.html.widgets import Dropdown
     from IPython.html.widgets import Button
-    from IPython.html.widgets import Box
+    from IPython.html.widgets import HBox as Box
     from IPython.html.widgets import FloatText
     from IPython.html.widgets import Checkbox
 
@@ -121,7 +121,8 @@ class ParameterWidgetGroup(object):
                         self.min_text, self.min_checkbox,
                         self.max_text, self.max_checkbox]
         display(box)
-        box.add_class('hbox')
+        if IPython.release.version_info[0] == 2:
+            box.add_class('hbox')
 
     # Make it easy to set the widget attributes directly.
     @property
@@ -199,9 +200,14 @@ class NotebookFitter(MPLFitter):
                 data_style={}, init_style={}, best_style={}, **kwargs):
         # Dropdown menu of all subclasses of Model, incl. user-defined.
         self.models_menu = Dropdown()
-        if all_models is None:
-            all_models = dict([(m.__name__, m) for m in Model.__subclasses__()])
-        self.models_menu.values = all_models
+        if IPython.release.version_info[0] == 2:
+            if all_models is None:
+                all_models = dict([(m.__name__, m) for m in Model.__subclasses__()])
+            self.models_menu.values = all_models
+        else:
+            if all_models is None:
+                all_models = [(m.__name__, m) for m in Model.__subclasses__()]
+            self.models_menu.options = all_models
         self.models_menu.on_trait_change(self._on_model_value_change,
                                              'value')
         # Button to trigger fitting.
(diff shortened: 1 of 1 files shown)


lmfit/parameter.py
```diff
@@ -102,6 +102,7 @@ def _update_param(name):
             if _updated[name]:
                 return
             par = self.__getitem__(name)
+            if not hasattr(par, '_expr_ast'):   par._expr_ast = None
```
tacaswell: Please do not compress if statements onto one line like this.

danielballan:
@MerlinSmiles Thanks for taking the time to fix this. Referring to the issue as you did is enough to create a reference between them. If you want #232 to be automatically closed when this PR is merged, use the phrase: closes #232.

lmfit/ui/ipy_fitter.py
```diff
@@ -10,21 +10,12 @@
 
 import IPython
 from IPython.display import display, clear_output
-# Widgets were only experimental in IPython 2.x, but this does work there.
```
danielballan: Why not keep 2.x compatibility?

lmfit/parameter.py
MerlinSmiles: Removed this again as it had nothing to do with the issue mentioned

lmfit/ui/ipy_fitter.py
```diff
@@ -116,12 +107,11 @@ def close(self):
         self.max_checkbox.close()
 
     def _repr_html_(self):
-        box = Box()
+        box = HBox()
         box.children = [self.value_text, self.vary_checkbox,
                         self.min_text, self.min_checkbox,
                         self.max_text, self.max_checkbox]
         display(box)
```
danielballan: By putting `box.add_class('hbox')` inside an `if` block like this we could keep compatibility with IPython 2.x.

```
if IPython.release.version_info[0] == 2:
    box.add_class('hbox')
```

MerlinSmiles:
@danielballan Ok, thanks, so I assume if i end this with the phrase it will work when it is merged at some point...

closes #323

lmfit/ui/ipy_fitter.py
MerlinSmiles: When I test this in an py2 environment it works fine with the widgets as they are. This is what I thought would stop py2 compatibility, this is a quick test:
```
%matplotlib inline
import numpy as np
import lmfit
from lmfit.ui import Fitter
data = np.array([1,2,3,4,5,4,3,2])
x = np.arange(len(data))/2.0

Fitter( x=x, data = data)
```

danielballan: We're talking about IPython 2/3, not Python 2/3, right?

MerlinSmiles: Oops, sorry totally missed that.
I'll revert those changes, but I cant test the widgets then though...

danielballan: I'm not following. IPython 2.x supports widgets. Ideally, NotebookFitter should work on both Python 2.7 and 3.x and IPython 2.x and 3.x. The widget API changed between IPython 2.x and 3.x. Later releases in the 2.x series support both 2-style and 3-style widgets, but early releases (e.g., 2.0) only support 2-style. That is why I imported things in two different ways.

MerlinSmiles: I must admit I didnt read carefully about the python Ipython issue.
I had the same problem as in the issue mentioned above, and fixed them for me. Here I'm using Python 3.4 and IPython 3.
Then I only checked with a Python 2.7 virtual env. if that would work, but still using Ipython 3 so I missed the IPython 2.x issue, as I thought it was without the I...
I'll move the Ipython 2 stuff back in.

danielballan: No problem! It takes some effort to cover all the possibilities. Thanks for taking the time to look into this.

tacaswell:
@MerlinSmiles You have to get the issue number right for the github magic to work ;)

MerlinSmiles:
@tacaswell I might be very confused at the moment :)

Issue #242, opened 2015-07-31
update parameter with set() method is inconsistent


newville:
Using `Parameter.set()` does not always work correctly, especially when switching a parameter back and forth between using `expr` and not.

Prompted by https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/lmfit-py/q2CMBqhwyWA/90tmoqPRCgAJ

Issue #237, opened 2015-07-05
Maybe something in Parameters update_constraint is broken


MerlinSmiles:
I don't really understand this error, and it is a bit hard to reproduce.

The error I get is:
```
/Users/Merlin/git/lmfit-py/lmfit/parameter.py in _update_param(name)
    107             if par._expr is not None:
    108                 par.expr = par._expr
--> 109             if par._expr_ast is not None:
    110                 for dep in par._expr_deps:
    111                     if dep in self.keys():

AttributeError: 'Parameter' object has no attribute '_expr_ast'
```
I get this error when I use the Ipython notebook and the embedded widgets, if I run this:
```
%matplotlib inline
import numpy as np
import lmfit
from lmfit.ui import Fitter
data = np.array([1,2,3,4,5,4,3,2])
x = np.arange(len(data))/2.0

Fitter( x=x, data = data)
```
and in the widget disable a 'vary' checkbox, then this error occurs when hitting the fit button.

I can make it disappear by adding
```
if not hasattr(par, '_expr_ast'):
    par._expr_ast = None
```
on line 104 in parameter.py

danielballan:
@newville Is it OK if we add ``self._expr_ast = None`` to the ``__init__`` method of ``Parameter``? That seems like the right place to "fix" this.

MerlinSmiles:
@danielballan this is already there.

danielballan:
Huh. Then this bug is confusing.

newville:
Yeah, that seems weird. @MerlinSmiles could you try printing out what the type for the 'par' variable is there.  I suspect something other than a Parameter might being passed in, somehow.   Like, it might be None here, as there are no actual parameters defined.    I also wonder if this might be related to the picklability problem we had a few months ago.  But probably having no parameters is the cause.... if that is the case, the message should definitely better!

tacaswell:
pickling smells right.  I did a bit of skimming to find where you could be creating instances with out going through `__init__` and pickle does that.

MerlinSmiles:
@newville Sure I can, I print ```par```, and ```par._expr_ast``` for the example above
```
par <Parameter 'decay', 1, bounds=[-inf:inf]>
par._expr_ast None
par <Parameter 'decay', 1, bounds=[-inf:inf]>
par._expr_ast None
par <Parameter 'amplitude', 1, bounds=[-inf:inf]>
par._expr_ast None
[...]

PR #243, opened 2015-07-31
Parameter set() fixes

newville:
This addresses issue #242, so that `Parameter.set()` works to set parameter attributes, even when already defined to have an `expr`.   It works in preliminary tests, but actual nose tests should be added.

It also adds the explicit setting of `_expr_ast=None` etc in order to address a problem with pickling and IPython uses, Issue #237


newville:
tests added, all tests passed, so I think this is simply a bug-fix, not breaking any expected behavior.  Will merge to master in ~2days (Monday) unless there are objections.

Issue #229, opened 2015-05-01
bad definition of pvoigt


Prestipino:
**Pseudo voigt** linefunction is defined as a the weighted sum of a gaussian and a lorentzian with the same fwhm. See  [http://en.wikipedia.org/wiki/Voigt_profile](http://en.wikipedia.org/wiki/Voigt_profile)
In the lmfit.lineshape.pvoigt pvoigt is defined as  the weighted sum of a gaussian and a lorentzian with sigma =eta

## Old code 
```python
def pvoigt(x, amplitude=1.0, center=0.0, sigma=1.0, fraction=0.5):
    """1 dimensional pseudo-voigt:
    pvoigt(x, amplitude, center, sigma, fraction)
       = amplitude*(1-fraction)*gaussion(x, center,sigma) +
         amplitude*fraction*lorentzian(x, center, sigma)
    """
    return ((1-fraction)*gaussian(x, amplitude, center, sigma) +
               fraction*lorentzian(x, amplitude, center, sigma))
```
## Possible corrected code
```python
def pvoigt(x, amplitude=1.0, center=0.0, fwhm=1.0, fraction=0.5):
    """1 dimensional pseudo-voigt:
    pvoigt(x, amplitude, center, fwhm, fraction)
       = amplitude*(1-fraction)*gaussion(x, center,fwhm) +
         amplitude*fraction*lorentzian(x, center, fwhm)
         
    """
    sigma_g=fwhm/(2*sqrt(2*log2))
    sigma_l=2*fwhm
    return ((1-fraction)*gaussian(x, amplitude, center, sigma_g) +
               fraction*lorentzian(x, amplitude, center, sigma_l))
```

newville:
@Prestipino  Hi!    OK, I think following the definition at  http://en.wikipedia.org/wiki/Voigt_profile#Pseudo-Voigt_Approximation  is the right thing to do.  

For consistency with the definitions of all the other lineshapes, I'd prefer to keep the parameter in the function called `sigma` (a PseudoVoigtModel could provide a parameter called `fwhm` derived from it.).   I think using the definition at the link, the value called `w` there maps to `sigma`.

PR #244, opened 2015-08-01
fix pseudo-voigt definition as per github issue #229

newville:
simple change to pseudo-voigt definition as per #229, so that the FWHM of the Gaussian and Lorentzian lineshapes are equal.  The `sigma` parameter is used un-altered for the Lorentzian, but altered for the Gaussian component.  The resulting FWHM of the sum is `2*sigma`.


newville:
merging this to master, as I'm pretty sure it is uncontroversial, and has no possible side-effects.

tacaswell:
attn @licode

newville:
@tacaswell @licode   actually, I'd recommend using the real Voigt function, not the pseudo-Voigt function.

PR #249, opened 2015-09-01
ENH: adding an emcee sampler to Minimizer

andyfaff:
This PR proposes to add the `emcee` Markov Chain Monte Carlo ensemble sampler to lmfit.  This is a Bayesian statistical technique for sampling the posterior distribution for the parameters. 
Amongst other things it is a robust way of examining parameter correlations, etc. Use of this method would require the user to install `emcee` and `pandas`.  However, if they are not installed then an Error is raised (no need to add extra dependencies for lmfit).
This method does not require upper and lower bounds for each parameter, but it would probably converge faster if it did.


newville:
@andyfaff  Wow, looks great!    Sorry I didn't respond to your earlier request -- I'm just getting back from travel and catching up on many things. I think it's a great idea to have this  `emcee` method in the `Minimizer` class, and I'm sure this will be very useful.  Some comments and points to consider:

1.  doc and worked simple-but-non-trivial example would be much appreciated.
2.  should probably include `nwalkers` in docstring.
3.  should the "1%" jitter parameter be a function argument?  Similar for the "15.8 / 84.2"?  BTW, should those be "32.6 / 67.4" to imply "1 sigma"?

4. would it be reasonable to make `flat_chain` (and possibly other parts of the result) attributes of the `MinimizerResult`, and then only return that 1 thing?   That would be my preference -- avoiding returning tuples is sort of the point of `MinimizerResult`, and one can then load that up with lots of potentially useful results.

5. The un-setting of bounds is *probably* appropriate here, but I worry about the (admittedly unlikely) event of someone providing an analytic Jacobian.   That is, it might be necessary to also "unset" the gradient scaling.  

One possible way would be to temporarily save all parameters bounds, set bounds_varying, but then set the actual parameter bounds to `-inf, inf` for the rest of the emcee calculation, and set them back when done.    I'm not 100% sure that is the right thing to do: if a user sets a bound, it is completely reasonable to assume that the residual/penalty function can never be called with values outside those bounds, which I think you ensure, but might be worth checking.  Anyway, it's probably the simplest thing to do.

lmfit/minimizer.py
```diff
@@ -499,6 +507,124 @@ def scalar_minimize(self, method='Nelder-Mead', params=None, **kws):
...
+        def lnprior(theta):
+            # stay within the prior specified by the parameter bounds
+            if (np.any(theta > bounds_varying[:, 1])
+                    or np.any(theta < bounds_varying[:, 0])):
+                return -np.inf
+            return 0
+
+        def lnprob(theta):
+            lp = lnprior(theta)
+            if not np.isfinite(lp):
```
Tillsten: Is that necessary?  -np.inf + 1 == -inf

lmfit/minimizer.py
```diff
@@ -499,6 +507,124 @@ def scalar_minimize(self, method='Nelder-Mead', params=None, **kws):
...
+                ub = np.inf
+            bounds_varying.append((lb, ub))
+
+        bounds_varying = np.array(bounds_varying)
+
+        self.nvarys = len(result.var_names)
+
+        def lnlike(theta):
+            # log likelihood
+            return -0.5 * self.penalty(theta)
```
Tillsten: I am still disagreeing with the likelihood, i am still thinking without weights the sigma is necessary to be included. See http://dan.iel.fm/emcee/current/user/line/

lmfit/minimizer.py
```diff
@@ -499,6 +507,124 @@ def scalar_minimize(self, method='Nelder-Mead', params=None, **kws):
...
+
+        def lnprob(theta):
+            lp = lnprior(theta)
+            if not np.isfinite(lp):
+                return -np.inf
+            return lp + lnlike(theta)
+
+        if ntemps > 1:
+            # jitter the starting position by scaled Gaussian noise (1% level)
+            p0 = 1 + np.random.randn(ntemps, nwalkers, self.nvarys) * 1.e-2
```
Tillsten: This line could be outside the if-statement. Maybe use know stdev if avaible?

lmfit/minimizer.py
```diff
@@ -499,6 +507,124 @@ def scalar_minimize(self, method='Nelder-Mead', params=None, **kws):
...
+            sampler = emcee.EnsembleSampler(nwalkers, self.nvarys, lnprob)
+
+        # burn in the sampler
+        p0 *= vars
+        for output in sampler.sample(p0, iterations=burn):
+            p0 = output[0]
+        sampler.reset()
+
+        # now do a production run
+        for output in sampler.sample(p0, iterations=steps - burn, thin=thin):
```
Tillsten: Maybe use sampler.run_mcmc instead?

Tillsten:
I think the bound handling is right way to go. The current emcee example could be updated to use the code here. My only real beef with code is the lnlike definition.

lmfit/minimizer.py
Tillsten: sry, ofc this comment was wrong.

Tillsten: For code clarity i would prefer to include l590 here:

    p0 = vars * (1 + np.random.randn(ntemps, nwalkers, self.nvarys) * 1.e-2)

Tillsten:
@newville The pentaly function is never called out of bounds.

lmfit/minimizer.py
newville: @Tillsten I agree that weighting should be included, but wouldn't that be part of what `penalty()` calculates? 

Perhaps support for scalar minimizers (and `penalty()`) needs to be handled separately from multivariate problems (and `_residual()`).  I may be biased, but scalar minimizers simply cannot be as good as multivariate minimizers for multivariate problems.   Reducing the residual to sum-of-square is deliberately throwing away useful information.

newville:
@Tillsten, @andyfaff   OK, I agree that with Andrew's defs for `lnprior()`, `lnprob()` (with explicit check for `inf`!) and `lnlike()` that the users objective cannot be called with out-of-bounds values.  

For bounds handling:  I agree that this is OK, except for the case of a user-supplied Jacobian, which will (I think) still "re-scale" the gradient (and so now be incorrect!).  I believe it will be simplest to set `par.min = par.max = None`, before the work of `emcee`, and then reset them before returning.

andyfaff:
> doc and worked simple-but-non-trivial example would be much appreciated.

My RST/Markdown skills are almost non-existent, but doc and examples are needed, I agree. What's not clear here, and indeed for the rest of the documentation, is how much of MCMC to explain.  One doesn't want to make it too complicated to use, but neither should it be a black box.  There are inevitably choices to be made regarding how the sampling is done (nwalkers, ntemps, etc). Should we just refer users to the `emcee` pages, or should we try and give detailed information?

> should probably include nwalkers in docstring.

Yes, missed that one. The `emcee` sampler needs quite a few walkers to sample properly. From the `emcee` website: "Walkers are the members of the ensemble. They are almost like separate Metropolis-Hastings chains but, of course, the proposal distribution for a given walker depends on the positions of all the other walkers in the ensemble. See Goodman & Weare (2010) for more details." It's best for `nwalkers >> nvarys`.

> would it be reasonable to make flat_chain (and possibly other parts of the result) attributes of the MinimizerResult <snip>?

Yes, that's reasonable.  How much documentation is required to explain what the chains represent?  Should we just refer them to the emcee documentation?

> The un-setting of bounds is probably appropriate here, but I worry about the (admittedly unlikely) event of someone providing an analytic Jacobian. That is, it might be necessary to also "unset" the gradient scaling.

The bounds are still set (changing lb+ub does not change the bounds in the `Parameter` instance), it's just that the internal parameter scaling is removed. I tried a couple of examples with the internal scaling transformations turned on, and the examples seemed to work.  What is not clear to me is whether the internal scaling in `lmfit` affects the sampling in `emcee`.  If the sampling is still done correctly then there would be no need to clear the parameter transformation.  I don't understand how someone providing an analytic jacobian would affect the emcee method, or subsequent operations using the `MinimizerResult`. The parameter transformation is setup whenever `prepare_fit` is called. So someone doing `emcee`, then `leastsq`, would not have any problems.  Does the user provided jacobian/gradient scaling affect the likelihood function?

> should the "1%" jitter parameter be a function argument? Similar for the "15.8 / 84.2"? BTW, should those be "32.6 / 67.4" to imply "1 sigma"?

The aim of the jitter is to give the walkers a small amount of variation about their initial position. The walkers soon spread out from this location. It's possible to give it as a function keyword, but I didn't want too many keywords to allow for future changes to the API.
The current parameter uncertainties are wrong.  By asking for 15.8 and 84.2 percentiles I'm calculating the +/- 1 sigma spread. I need to divide the current uncertainty by 2.  We need to remember that one of the reasons for doing this is because the probability distribution for the parameters may not be normally distributed, but could be skewed (or multimodal).  This is why having the chains is important.

lmfit/minimizer.py
andyfaff: Using `run_mcmc` would remove the for loop.  However, it still would need to be run twice, so that the burn samples could be discarded. Neither `run_mcmc` nor `sample` have a `burn` keyword. THe burn samples could be discarded from the chain in post processing, but the chain output from the serial and parallel samplers has different dimensionalities - one is 3D the other 4D. Which makes handling messier.

lmfit/minimizer.py
andyfaff: I had it that way originally. 
It is clarity vs having two lines of code.  No problem to change it.

lmfit/minimizer.py
andyfaff: The likelihood function was need to use is:

-\frac{1}{2}\sum_n \left[ \frac{(y_{obs} - y_{calc})^2}{s_n^2} + \ln (2\pi s_n^2) \right]
 (http://dan.iel.fm/emcee/current/user/line/)

lmfit instructs users to write an objective function that returns a (possibly weighted) residual.  The weighted residual vector is the first summand in the square brackets on the RHS.  The sum of the weighted residual vector is chi^2. chi^2 obviously varies as the model changes.
The second summand term,  \ln (2\pi s_n^2), is a constant and doesn't vary as parameters vary; unless one of the parameters is related to the noise term (as per the emcee example).
The `emcee` method only has access to `lmfit.penalty` or `lmfit.__residual`.  If I use `lmfit.penalty` (a scalar) I have to assume that it corresponds to the entirety of the likelihood function (assumes second summand is ignored), apart from multiplying it by -0.5.
If I use `lmfit.__residual` then the presumed behaviour should be:
* to square and sum if the output is a vector (this may under or overestimate depending on whether the second summand is included in the objective function output)
* to use as is if the output is a scalar
then to multiply by -0.5 as before.

One can get the exact likelihood function by saying that the user objective function has to return a scalar, which has included both summand terms correctly and has summed over all the points.

Unfortunately there is no way for the likelihood function to know what the objective function is returning (i.e. whether the user has weighted or not).

andyfaff: @Tillsten, so that I understand more fully, and with reference to the emcee equation for the log likelihood function, what terms do you think we're ignoring?

andyfaff:
On a related not, I was looking at the `scalar_minimize` code and there is the following line:

    _log_likelihood = result.ndata * np.log(result.redchi)

I'm not sure if this is correct.  likelihood is proportional to `exp(-0.5 * chi2)`, so log-likelihood would be something like `result.ndata * result.chi2`, i.e. without `np.log`.

lmfit/minimizer.py
Tillsten: I think two calls of run_mcmc are  clearer as most readers wont know that sample is a generator function and even with that it looks strange. But i trust your judgement.

lmfit/minimizer.py
Tillsten: @andyfaff 
I agree that the current code is indeed correct when the weights are supplied by the user. But if sigma is not know, shouldn't we marginalize over sigma by including it in the sampling? This is at least what every mcmc-least squares example i could find does (http://jakevdp.github.io/blog/2014/06/14/frequentism-and-bayesianism-4-bayesian-in-python/). But it is quite possible that i understand something wrong. So if you are sure, go right ahead!

Tillsten:
Maybe make nwalkers depened on nvars? Maybe use walkers_per_var instead?

Tillsten:
It would be really nice to haven an easy way to continue sampling after a run.

newville:
Sorry, it's challenging for me to keep up with the whole conversation.

On Tue, Sep 1, 2015 at 7:13 PM, Andrew Nelson <notifications@github.com>
wrote:

> doc and worked simple-but-non-trivial example would be much appreciated.
>
> My RST/Markdown skills are almost non-existent, but doc and examples are
> needed, I agree.
>

RST is very easy and the existing docs give plenty of examples.  I would
recommend documenting the `emcee` method in 'confidence.rst'.    And a
simple example would be very helpful.

What's not clear here, and indeed for the rest of the documentation, is how
> much of MCMC to explain. One doesn't want to make it too complicated to
> use, but neither should it be a black box. There are inevitably choices to
> be made regarding how the sampling is done (nwalkers, ntemps, etc). Should
> we just refer users to the emcee pages, or should we try and give
> detailed information?
>

Right -- documenting what the function arguments are is important, but you
don't have to explain MCMC.  Referring to the emcee pages is fine, and/or
pointing to general references (even wikipedia) for "what is MCMC" is
fine.   You can assume that anyone using lmfit, and reading more than one
page of documentation has some understanding of the basics.


> should probably include nwalkers in docstring.
>
> Yes, missed that one. The emcee sampler needs quite a few walkers to
> sample properly. From the emcee website: "Walkers are the members of the
> ensemble. They are almost like separate Metropolis-Hastings chains but, of
> course, the proposal distribution for a given walker depends on the
> positions of all the other walkers in the ensemble. See Goodman & Weare
> (2010) for more details." It's best for nwalkers >> nvarys.
>

That's fine -- and probably close to enough for the doc.  I was just
pointing out that the docstring was missing.

> would it be reasonable to make flat_chain (and possibly other parts of the
> result) attributes of the MinimizerResult ?
>
> Yes, that's reasonable. How much documentation is required to explain what
> the chains represent? Should we just refer them to the emcee documentation?
>

I think that's fine.  A brief description might be good too, probably
focusing on what data is held in the object rather than it's theoretical
underpinnings.


> The un-setting of bounds is probably appropriate here, but I worry about
> the (admittedly unlikely) event of someone providing an analytic Jacobian.
> That is, it might be necessary to also "unset" the gradient scaling.
>
> The bounds are still set (changing lb+ub does not change the bounds in the
> Parameter instance), it's just that the internal parameter scaling is
> removed. I tried a couple of examples with the internal scaling
> transformations turned on, and the examples seemed to work. What is not
> clear to me is whether the internal scaling in lmfit affects the sampling
> in emcee. If the sampling is still done correctly then there would be no
> need to clear the parameter transformation. I don't understand how someone
> providing an analytic jacobian would affect the emcee method, or subsequent
> operations using the MinimizerResult. The parameter transformation is
> setup whenever prepare_fit is called. So someone doing emcee, then leastsq,
> would not have any problems. Does the user provided jacobian/gradient
> scaling affect the likelihood function?
>

The analytic Jacobian does *not* affect the residual/cost/penalty
function.  My main concern is what the user's objective function sees when
called by this procedure, and what state the parameters are in when the
procedure is done, and I'm just trying to think through all the possible
ways things can go wrong.

After thinking about this some more, I think the approach you've taken
(leave the Parameter.min, .max in place, hijack `from_internal`) is best
approach.   I'm willing to say it might need to be revisited.

That is,  there is no reason to work in "internal" parameters for this code
-- it would probably be OK, but is a distraction.  A user's objective
*could in principle* use the bounds, (though I'm not sure I see an obvious
case where that would happen), so leaving those in place is nice. It's very
hard to imagine someone relying on the analytic Jacobian inside their own
objective function.     So, I think the "hijack from_internal") is a fine
approach.


> should the "1%" jitter parameter be a function argument? Similar for the
> "15.8 / 84.2"? BTW, should those be "32.6 / 67.4" to imply "1 sigma"?
>
> The aim of the jitter is to give the walkers a small amount of variation
> about their initial position. The walkers soon spread out from this
> location. It's possible to give it as a function keyword, but I didn't want
> too many keywords to allow for future changes to the API.
> The current parameter uncertainties are wrong. By asking for 15.8 and 84.2
> percentiles I'm calculating the +/- 1 sigma spread. I need to divide the
> current uncertainty by 2. We need to remember that one of the reasons for
> doing this is because the probability distribution for the parameters may
> not be normally distributed, but could be skewed (or multimodal). This is
> why having the chains is important.
>

Yes, I agree that non-normal is one of the main points.  It's just that it
should probably aim for "1 sigma" (68.3 not 67.4 percentile) or 3 sigma"
(99.7 percentile) instead of "84.2 percentile" (is that 1.414 sigma?).

lmfit/minimizer.py
andyfaff: I understand where you're coming from.  I'll try it out with `run_mcmc`

andyfaff:
> Maybe make nwalkers depened on nvars? Maybe use walkers_per_var instead?

emcee itself complains when nwalkers is too low compared to nvars.

andyfaff:
>It would be really nice to haven an easy way to continue sampling after a run.

I agree.  It's not clear to me how to create such a mechanism.

newville:
> It would be really nice to haven an easy way to continue sampling after a
> run.
>
> I agree. It's not clear to me how to create such a mechanism.
>
>
Could you include the `sampler` created in `emcee()` as part of the
`MinimizerResult`?  You then might consider breaking emcee into two parts
(`emcee_setup()` to setup and create the sampler,  and `emcee_run()` to do
the production run and organize results).  That way you could  allow the
user to send in an existing sampler to the "emcee_run()" method, allowing
them to do some sampling, investigate results, then continue sampling.


Re stderr: taking half of the "15.8", and "84.2" quantiles is clear, and
looks good to me.

andyfaff:
The sampler object is constructed using the `lnprob` nested function inside the `emcee` method of a `Minimizer` object.  Keeping the sampler presumably means keeping a reference for the `Minimizer` object.  I don't know what will happen if someone changes (e.g. changing which parameters are varied) the `Minimizer` object in-between calling `emcee` and asking the sampler to sample again. I think it's something I have to try and see what happens.
The solution may be to add a `continue` keyword (or something like that) to the emcee method that recognises that the sampling is supposed to keep going. However, one would still have to check that the Minimizer hasn't changed in the meantime (bounds were changed, number of varied parameters are changed, someone changed the userfcn, etc).

andyfaff:
With regards to the likelihood function including, or ignoring data uncertainties, I think that that's something that the user needs to take care of when they write the likelihood function.
I mentioned earlier that the logic would be:
1) userfcn returns a vector, assume it's weighted residuals. Sum and square to get chi2.
2) userfcn returns a float.  Assume that that's the likelihood function (although it's probably chi2).

Multiply 1 or 2 by -0.5.

At the end of the day, at this level, it's up to the user to write the correct function.  We can make things as easy as possible, but we can't account for everything.  How is our code supposed to know if the `__residual` function is uncertainty weighted or not - that's up to the user?

Tillsten:
The most common use case from user is probably returning a non weighted vector, where not sampling sigma is questionable and AFAIK wrong. I made [a notebook] based on jakes blogpost (https://gist.github.com/Tillsten/714c7edc704ed7b66d2f) to show the effect of a wrong fixed sigma, the second graph is with a fixed wrong sigma value. After seeing that i would even prefer to include a mandatory parameter, maybe `is_weighted` or `sample_sigma`.

Tillsten:
If we can`t reasonably guess the users intent, we should ask him. As above, my hunch is that
returning no weighted vectors is very common, and silently returning a wrong result is than the wrong thing to do.

Again, sorry for being obnoxious and thanks for your great work!

andyfaff:
> I made [a notebook] based on jakes blogpost

I'm interested in reading that, do you have a link?

Also, I note that your example has a non uniform prior. What's the
derivation of that? Should an emcee method have a keyword for a lnprior
function?

Tillsten:
The word notebook is a link. I have recheck my example, i don't remember using a non uniform prior.

Tillsten:
Ok, i checked my example, like your prior it just checks if the variables are in their bounds.

andyfaff:
Ok, i checked my example, like your prior it just checks if the variables
> are in their bounds.
>

return -1.5 * np.log(1 + beta ** 2) - np.log(sigma)


This looked as if the prior wasn't uniform.



-- 
Dr. Andrew Nelson


Tillsten:
Oh, i thought my you meant the example included in [lmfit](https://github.com/lmfit/lmfit-py/blob/master/examples/lmfit_and_emcee.py). Yeah, the prior in the nb is taken from blogpost. It makes sure that each incline of the line has the same probability. It should also not matter for point i am trying to make.

andyfaff:
@Tillsten, I can see where you're coming from, and I share your concerns.
Data uncertainties should be treated in an appropriate way.
Here are the cases I'm thinking about:

1. Objective function returns a single value.  At this point we have to
assume that they are either returning a correct loglikelihood or at the
very least returning a properly weighted chi2. If it's just chi2 then they
are ignoring the constant ln(2.pi.s_n) term.

2. Objective function returns a vector. At this point we either assume that
the user is a) returning (possibly weighted) residuals or b) the correct
log-likelihood for each point.

The first is easy to deal with, the choice has been taken out of our hands,
we have to assume the user is returning the correct log-likelihood. emcee
will multiply the return value by -0.5. Even if we had a sigma keyword then
there's nothing else we can do in this situation because everything has
already been summed.

The second is harder to deal with. If the vector is a correct
log-likelihood then all we should do is sum (and multiply by -0.5). If the
vector is straightforward residuals (a la the rest of lmfit) then one would
need to square, add on an data uncertainty term, sum and multiply by -0.5.
With 2a) we shouldn't do anything else because the user is already
returning the right thing. With 2b) it's possible for us to make
corrections for incorrect/non weighted data. But is the correction for a
fractional underestimation of sigma, or a constant underestimation of
sigma, etc?

At the moment with both 2a and 2b I am assuming that the vector is just
residuals, because the rest of lmfit works that way.  So the procedure is
to square, sum and multiply by -0.5.

In my opinion it's probably better to show the user how to make a correct
likelihood function in documentation.  If they know how to do this then
they can handle the incorrect sigma by amending their own objective
function. This is purely because we don't know what form the uncertainties
associated with their data should be.

Even if we did, it's still difficult to deal with.  Let's take the example
that the uncertainties are fractionally underestimated, a.la. the emcee
example. If the user gives us the weighted residuals _and_ and the sigma
uncertainty array then we can multiply the residuals by sigma to give
(model - data).  We then need to divide (model - data) by a corrected
uncertainty array.  However, to correct the uncertainty array we need to
know the model function which we don't have access to.
I'd put that one in the too hard basket.

I hope this makes sense. I think we both want to get to the same
destination.

andyfaff:
@Tillsten @newville please see https://gist.github.com/andyfaff/ea9e0ec69e378ab75952 for a non-linear regression example with `emcee`. I've beefed up the docstring, and will start writing the documentation based on this example. The example was based on the example in doc/confidence.rst.

Tillsten:
@andyfaff  i am in a conference this week so my feedback will wait till next week.

newville:
@andyfaff @Tillsten   There are several inter-woven conversations here, which makes it a bit difficult for me to keep up. The work here is good, but I think there are lingering questions, and that this will have to wait to merge into master until after 0.9.0 is released, hopefully by the end of the month.  

@andyfaff, for the idea of "continuing" a sampling,  I think it will be fine to include a `sampler` instance in a returned `MinimizerResult` -- yes, this references a particular minimization run (the "state"), but that's sort of the point of `MinimizerResult` anyway.

For the docs and examples, the RST doc needs a brief overview of what the `emcee` method does,  and a description of function arguments and returned values.  I think this  most naturally fits in `confidence.rst` for "better ways to estimate uncertainties", but since `emcee` is a method of `Minimizer`, it needs to be mentioned there as well (probably just linking to the more complete doc).   There should be a simple, non-trivial (python, not ipython notebook) example in the examples/ folder to show how it works (this can be incorporated into the docs too).   To be clear, these are both needed for the PR to be acceptable.

The IPython notebook in your gist is great, and I think that having a set of IPython notebook examples for lmfit would be fabulous.  For this particular case, it would be very useful to compare the results for the uncertainties with those from brute-force search of `conf_interval()`, and also from the automated estimate from `leastsq()`.   Maybe including timing results too?

I think it's important that the users objective function returns a quantity to be minimized, which may not necessarily be a correctly scaled chi^2 statistic, or "negative likelihood" or a "negative entropy" or **necessarily** anything.   If `emcee()` must make an assumption, it should be that the objective returns an unscaled chi^2 unless told otherwise.   It seems perfectly reasonable to have `emcee()` take a parameter to tell it what the objective function actually returns, and possibly also an `epsilon` parameter (`sigma` is used in too many related senses -- the uncertainty in the data should be `epsilon`) that the user can use to tell `emcee()` how to scale the return value.

Tillsten:
@andyfaff Notebook: looks nice so far! small nitpicking would include calling the error estimate first f and than directly call it s. Sadly there is no easy why to include a nb in sphinx yet, or did it change? I would volunteer to create the sphinx docs out of your nb if you are finished and dont want to do it.

For the uncertainty, i think we should follow the zen of python "Explicit is better than implicit" and
have a `is_scaled` parameter defaulting to False. In that case a new parameter sigma (or epsilon) is added the to Minimizer which will be included into the sampling.

newville:
@tillsten @andyfaff I believe ipython notebooks can be converted to RST with nbconvert, but I'm not sure if that can be done dynamically.

andyfaff:
I've been modifying the PR a little bit. Now possible to `reuse_sampler`,
and initialise the sampler from a given position. I've also written quite a
bit of the documentation. The emcee method is documented in fitting.rst,
and I've given a reasonably advanced example there. I've also written a
section in confidence.rst.  However, I didn't reproduce the example there,
I made a link back to fitting.rst.

I'm still thinking about the `is_scaled` keyword.  I'm not convinced that
it's the best way to go, but I'll formulate all that in a later email.

On 18 September 2015 at 02:08, Matt Newville <notifications@github.com>
wrote:

> @tillsten <https://github.com/tillsten> @andyfaff
> <https://github.com/andyfaff> I believe ipython notebooks can be
> converted to RST with nbconvert, but I'm not sure if that can be done
> dynamically.
>
> —
> <https://github.com/lmfit/lmfit-py/pull/249#issuecomment-141134720>.
>



-- 
Dr. Andrew Nelson


andyfaff:
Also forgot to say that I'm also working on a model selection example. For
example, what are the relative probabilities that a given dataset require
1,2,3,4...N Gaussians.

On 19 September 2015 at 18:24, Andrew Nelson <andyfaff@gmail.com> wrote:

> I've been modifying the PR a little bit. Now possible to `reuse_sampler`,
> and initialise the sampler from a given position. I've also written quite a
> bit of the documentation. The emcee method is documented in fitting.rst,
> and I've given a reasonably advanced example there. I've also written a
> section in confidence.rst.  However, I didn't reproduce the example there,
> I made a link back to fitting.rst.
>
> I'm still thinking about the `is_scaled` keyword.  I'm not convinced that
> it's the best way to go, but I'll formulate all that in a later email.
>
> On 18 September 2015 at 02:08, Matt Newville <notifications@github.com>
> wrote:
>
>> @tillsten <https://github.com/tillsten> @andyfaff
>> <https://github.com/andyfaff> I believe ipython notebooks can be
>> converted to RST with nbconvert, but I'm not sure if that can be done
>> dynamically.
>>
>> —
>> Reply to this email directly or view it on GitHub
>> <https://github.com/lmfit/lmfit-py/pull/249#issuecomment-141134720>.
>>
>
>
>
> --
> _____________________________________
> Dr. Andrew Nelson
>
>
> _____________________________________
>



-- 
Dr. Andrew Nelson


andyfaff:
[Model selection with `Minimizer.emcee`](https://gist.github.com/andyfaff/c51d04cec2a6454c1e1a). This takes advantage of multiprocessing to give a large speedup in the calculation time.

andyfaff:
Ping @newville, @Tillsten, I'd like to advance this PR. I'm not sure of what still needs to be done.

A.

newville:
@andyfaff  Sorry for letting this slip off the radar screen.  I think this is in very good shape, and have no objections to merging this.    Are there any other concerns left or is this ready to merge?

lmfit/minimizer.py
```diff
@@ -498,6 +530,294 @@ def scalar_minimize(self, method='Nelder-Mead', params=None, **kws):
...
+        self._lastpos = output[0]
+
+        # discard the burn samples and thin
+        chain = self.sampler.chain[..., burn::thin, :]
+        lnprobability = self.sampler.lnprobability[:, burn::thin]
+
+        flatchain = chain.reshape((-1, self.nvarys))
+
+        mean = np.mean(flatchain, axis=0)
+        quantiles = np.percentile(flatchain, [15.8, 84.2], axis=0)
```
Tillsten: Just a suggestion: maybe return also the 2-sigma quantiles? MCMC is most useful when the estimation from the covariance fails, and in these cases 2-sigma may not be 1-sigma*2. Alternative, take the quantils as a parameter But as the chain is returned anyways, it doesnt matter too much.

Tillsten:
@andyfaff I am still strongly in favour of including sigma (and scale by it) as a default. Nowhere else in lmfit we assume that the errors are correctly weighted, and i suspect many will use it wrong and underestimate their error dramatically. Else the PR is very nice work!

andyfaff:
I'd like to get Till's points dealt with before it's merged. You should be
aware that the log-likelihood is calculated by a new function called
`_lnpost`, which is not a method of `Minimizer`. This enables
parallelisation. `_lnpost` calls `userfcn` directly, it does not call
`Minimizer.penalty` or `Minimizer.__residual`.

lmfit/minimizer.py
andyfaff: Reflecting on this a little more, instead of using the mean value as the MLE we should probably use the median (50th percentile). I think this is a more robust statistic if the probability distribution is non-normal. If it is normally distributed than the median and mean should be the same. 

My hope is that the user plots the chain (using the corner package) to see the distribution of parameter values. The user can always calculate 2sigma (and non normality) from the chain. To that end, I prefer to stick with 1-sigma values for stderr. Also, I prefer not to add extra keywords...

andyfaff:
@Tillsten, @newville I've added an `is_weighted=True` keyword to the `emcee` method. This tells the method that iff `userfcn` returns a vector, then `userfcn` has already weighted the residuals correctly. Iff `userfcn` returns a vector and `is_weighted is False` then a data uncertainty is marginalised over during the sampling.
If `userfcn` returns a `float` then `is_weighted` is ignored because the user is assumed to be returning the actual log-posterior (or chi2).
Have a look at [fixing_sigma.ipynb](https://gist.github.com/andyfaff/33fc912f0c705a86874c), which is a version of the notebook that @Tillsten posted above. In it I demonstrate that I can reproduce the output of the original notebook, but via `Minimizer.emcee`.
I prefer `is_weighted` to default to `True`, in my `userfcn` I usually normalise the residuals by the uncertainties. Hopefully this is a good compromise.

Tillsten:
@andyfaff So much nice work, thank you! I still would prefer to default it to `True`, knowing some of my colleagues. Everything else is good as far as i checken. @newville I let you two decide on the default, else lets merge it!

newville:
@andyfaff @Tillsten   Yes, that's a very nice example, showing the power of the emcee() method!   That will be a great compliment to the confidence_interval code.

I'm fine with all of the recent choices, like preferring median to mean,  as it's probably more appropriate if you're here anyway.   Similarly, the default for `is_weighted` of `True` seems fine to me, as if you're using these methods you really ought appreciate the difference.

Can I request (if it's not already there -- I haven't looked at the rendered docs) an example / code-snippet showing how to take the emcee() results to extract 2- and 3-sigma values?  I think that would be a common request, though it probably shouldn't delay merging this.

Any objections to merging?

andyfaff:
There are a few more things I need to check before I think it should be merged (I looked at the doco, and some of the graphs don't come out the same as before, I need to pin that down).
 I'll also add the 2-3 sigma values in this PR.

andyfaff:
@Tillsten, @newville  I'd appreciate it if you could look at the jupyter notebooks I've just added to the examples directory. Can you follow them, are there any mistakes? (@Tillsten I don't think your emcee example works anymore).
Also, I'd appreciate it if you could read the `Minimizer.emcee` docstring to see if it all makes sense. I'm particularly interested in seeing if you understand all the instructions about creating a log-posterior function (in the notes section), and if the `is_weighted` and `float_behavior` keywords make sense. All constructive criticism welcomed.

If you're ok with it, then we're one step closer to a merge.

Tillsten:
@andyfaff At the end of the examples, I would add also print out the 1-sigma intervals.

newville:
@andyfaff  The notebook looks very nice.  

My overly picky suggestions would be to call the noise parameter 'noise' or 'epsilon' instead of 'f',  and to make the `lnprob` function be a little more obvious (maybe to me only), perhaps

```python
def lnprob(p):
       noise = p['noise']
       resid = (residual(p)/noise)**2 
       return -0.5 * np.sum(resid + np.log(2 * np.pi* noise**2))
```
but that's sort of just style.  I didn't see (perhaps missed on a quick read) what `float_behavior` is.  

Also, in the very, very nice corner plots, the "banana shape" implies non-Normal distribution, not high correlation.  Ellipticity implies correlation.

Anyway, I think this is all very nice.   I'll be happy to merge when you think it's ready.

andyfaff:
@newville @Tillsten I think I'm happy for this to be merged, assuming that others don't have any objections. I've been staring at this thing for such a long time and my vision is probably immune to the bugs now.
Is there a next release date in mind?

newville:
@andyfaff @Tillsten   I agree this is ready to be merged.  This is really nice work, making it very easy to do something that many people know can be an important approach to understanding uncertainties but is somewhat of a daunting task for non-experts to actually do.   I think this will be useful for many people.

If there are no objections, I'll merge this in about a day.

I don't have a strong opinion on the next release.  It would be OK with me to tag 0.9.3 after this merge, so that it has major changes of:
  * add emcee() method
  * speed up parameter._getval()
  * make asteval work with Py3.5.

I'm not sure what a time estimate for that would be....

Tillsten:
Fully agree on merging and making releases often. @andyfaff I think new features always come with a implied "provisional"-tag in a small project as ours, so if we find bug or annoyances we will fix it later.

Tillsten:
But it may be worth to support the new scipy 0.17 lm-solver in the next release too. I can't test it because i now fully depend on anaconda, which is still 0.16 on windows 64.

newville:
@Tillsten @andyfaff I think support for the new, very interesting, bounded LM code in scipy 0.17 could be kind of messy -- we'll need options added for whether to use it or MINUIT-style bounds.  In the spirit of release early+often, I'd suggest that to be the goal for 0.9.4

newville:
@andyfaff @Tillsten merged and tagged as 0.9.3rc1

PR #254, opened 2015-09-21
ENH: pickling Parameters

andyfaff:
This PR adds machinery to pickle Parameters, and adds tests to ensure that both Parameter and Parameters can be pickled.
The test is in tests/test_parameter.py.  A lot of the other tests regarding parameters (e.g. test_params_set) could possibly be rolled into this file to make things a little clearer.


newville:
@andyfaff  OK, that looks interesting....  I'd like to see more thorough testing of the re-created Parameters.  

I'm assuming this would really be needed in the context of a larger PR for using multiprocessing.  Is that correct?

My recollection was that we ran into a small problem in pickling and IPython (or, really zeromq), so we should probably test that too...

andyfaff:
Yes, it'll be needed for any parallelisation.  Parallelisation will
probably also require that the objective function, `Minimizer.__residual`
be pickleable. This will probably take the form of a `_Residual` class,
which will be initialised with everything the current `__residual` method
uses. It'll also have a `__call__` method so that the residuals can be
calculated.

With respect to more thorough testing, do you have any suggestions for what
kind of testing is required? It's certaintly not a problem to add a few
more `Parameter` into the mix, and to test it out with `expr`. I've pickled
in Jupyter (IPython) and there doesn't seem to be a problem.  I couldn't
answer for interprocess communication though.

On 22 September 2015 at 02:07, Matt Newville <notifications@github.com>
wrote:

> @andyfaff <https://github.com/andyfaff> OK, that looks interesting....
> I'd like to see more thorough testing of the re-created Parameters.
>
> I'm assuming this would really be needed in the context of a larger PR for
> using multiprocessing. Is that correct?
>
> My recollection was that we ran into a small problem in pickling and
> IPython (or, really zeromq), so we should probably test that too...
>
> —
> <https://github.com/lmfit/lmfit-py/pull/254#issuecomment-142029228>.
>



-- 
Dr. Andrew Nelson


newville:
@andyfaff 

> Yes, it'll be needed for any parallelisation.  Parallelisation will probably also require that the objective function, `Minimizer.__residual`  be pickleable. This will probably take the form of a `_Residual` class, which will be initialised  with everything the current `__residual` method uses. It'll also have a `__call__` method so that the residuals can be calculated.

I'm bothered by the idea of needing to alter to a Python-only library so that it can be used with pickle and multi-processing.  Adding extra junk like `__reduce__` and `__setstate__` just to satisfy pickle's poor implementation is bad enough.    Pickle should be fixed, or multiprocessing should be fixed to use a decent object serialization process.  My understanding is that dill makes many improvements to pickle in this regard, but perhaps it's not enough.

Refactoring `Minimizer.__residual` to be a separate class makes no sense to me.   Does that even help make pickling easier?  How?    If `Minimizer.__residual` is not picklable, why would '_Residual.__call__` be?   What would be the requirements for the user's objective function?  Doesn't that need to be picklable too?   How do you ensure that?  To be clear, these are not rhetorical questions, and will need to be answered clearly.

I understand that asteval does some tricks that might need tweaking for Pickle to work.  I might be persuaded to make such changes.  But I'm reluctant to refactor `Minimizer` for pickle.  It would be better to fix multiprocessing so that `Minimizer.minimze` can be run on multiple processors.

andyfaff:
The problem is that multiprocessing has to pickle objects to distribute them among processes. The `pickle` module is used by the `multiprocessing` module to achieve this, and it does have faults.

I tried using `dill`, and `Parameters` was not pickleable by that module either.  Use of `dill` would have required the `pathos` module for multiprocessing as well, two extra dependencies.

`__reduce__` and `__setstate__` are magic methods designed exactly for this purpose. A common example is pickling an object that contains an open system resource like a file handle.  You can't just copy those, they have to be dealt with in a specific way. Addition of these methods enables `Parameters` to be pickled.  There are **no** `asteval`, or any `asteval` related attributes, that require tweaking for this to work.

With regards to `Minimizer.__residual`, this is supplied as the `func` argument to the parallel map implementation. This method needs to be pickleable in order to distribute across different processes. In order to pickle the method the entire `Minimizer` object that contains it will need to be pickled.  I haven't tried it yet, but I suspect I'll run into difficulties.  This is because Minimizer is a complex object that also contains the sampler that is asking for the pickling to be done. I've already checked that a [_Residual](https://gist.github.com/andyfaff/17e2920e762d3fdb833c) class can be pickled.  I made a small branch in which I replaced the `Minimizer.__residual` method with a callable attribute, `self.__residual = _Residual()`, and all the lmfit tests passed (it's around 30 minutes work).  Use of a `_Residual` class would mean that `Minimizer` does not have to be pickled.

The users objective function does need to be pickleable.  If it's not pickleable than an exception will be raised when the objective function is sent to `Pool.map`. We can't ensure that, apart from mentioning it in the documentation.  

If there is an aversion to refactoring Minimizer to use a _Residual class, then would you prefer to see the emcee related work in its own Sampler class?

andyfaff:
So i tried a few different approaches to parallelising emcee today, it's quite straightforward and a lot simpler than the discussion above made it sound. It won't involve refactoring Minimizer.  userfcn won't get called through __residual, but through a function instead.
The Parameters still have to be pickled though.

andyfaff:
@newville I've checked that `Parameters` can be pickled in an IPython session.  I wouldn't know how to go about doing such a test in a nose script.

Would you like me to add any extra tests, or investigate anything else, for this PR?  The included tests show that `Parameters` can be pickled with these changes, and that the unpickled object is equal to the original object.
The first value returned from the `__reduce__` method tells the pickler how to reconstruct a `Parameters` instance when unpickling. The last value returned is the argument that has to be supplied to `__setstate__` when the object is unpickled.  This is a list of the constituent `Parameter` objects. `__setstate__` simply uses that list to add the parameters back in one by one.

     def __reduce__(self):
         """
         Required to pickle a Parameters instance.
         """
         items = [self[k] for k in self]
         return self.__class__, (), items
 
     def __setstate__(self, state):
         """state is list of parameters"""
         for param in state:
             self.__setitem__(param.name, param)

andyfaff:
PR has been updated in light of #261. Does anyone have feedback on this PR? For example, does anyone want me to add specific pickling tests?  If possible I'd like to get the PR merged in the near future.

newville:
@andyfaff  Looks fine to me.   I think more elaborate tests would be nice, but it's clear that the majority of the basic functionality is covered by the test.     If I don't hear other concerns, I'll merge this later today.

PR #255, opened 2015-09-22
Fix ipyfitter

danielballan:
This is a belated fix for the NotebookFitter on IPython 4.0. I believe it should also work correctly on IPython 2.x and 3.x, but I am struggling to test it just this moment -- some unrelated lapack error is blocking my progress.


newville:
@danielballan  Thanks!  I think this should be 0.9.0rc2 and be nearly ready for release.

PR #258, opened 2015-10-03
MAINT: .gitignore

andyfaff:
`.idea/` is a directory for Pycharm project files that is normally stored in the root of a repository. It keeps showing up whenever I want to add and commit files to my fork, I'd like to ignore those files.
Also `.DS_Store` are files that OSX uses to store preferences for folder views.  There are a plethora of them on OS X machines.  Could we also ignore those too?

Finally, git has suddenly decided to change the line endings in:
THANKS.txt,
examples/doc_model1.py
examples/doc_model2.py
lmfit/_version.py

They were CRLF, but have been changed to LF. I think this was because of changes made to .gitattributes in a recent commit.


newville:
@andyfaff, @Tillsten   Thanks!  

I would agree to add (almost) anything to .gitignore.  I also recommend `git status -uno`  (I keep an alias for this in $HOME/.gitconfig) and `git status --ignored` too.

For line-endings, I'm willing to take the full blame, and defer to any consensus view.   I develop on and commit from Windows, Mac, and Linux and use multiple editors.  I'm totally used to and mostly ignore line ending differences in text files.

Should we change `.gitattributes` to have `* text=auto`?  Honestly, I have no preference.

Tillsten:
Automate everything! Also auto-strip trailing spaces if possible.

andyfaff:
I vote for sticking with LF. However changing the git attribute for the
line ending just got me in a whole heap of trouble. As soon as git ingests
a file it marks it as modified, even if I haven't opened it. This made it a
problem to change branches.

tritemio:
I also had some trouble with line endings with recent lmfit master. I also develop on many OS and always stick to auto conversion of line-ending to \n under Windows. I'm not sure which option it it, as I just use the recommended defaults when installing git under windows.

So, as @andyfaff I would stick to LF.

newville:
Actually, as I think about this more, using `autocrlf` seems like a very bad idea, and do much prefer enforcing using LF.   Is a one-time, large cleanup to make all text files strictly eol=lf worth doing?

Probably like many of you, I use git from several OSes on the *same* repository with drives mapped by NFS or CIFS.   So, I think that "autocrlf" can not possibly make sense unless one only uses a copy of a repository from a single OS.

tritemio:
For your convenience:
https://git-scm.com/book/en/v2/Customizing-Git-Git-Configuration#Formatting-and-Whitespace

@newville, I agree, sources should only have LF. I don't know which one is the best way to enforce it. Maybe just using the defaults will be good enough. Most people on Windows will have the right global settings (default): converting CRLF to LF on commit and LF to CRLF on checkout. 

I don't know if a setting in .gittatributes can enforce this behavior consistently on all platforms.

tacaswell:
Part of this is making sure your editors respect existing line-endings of opened files.

andyfaff:
The git client will see git attributes and automatically make sure that the
correct line endings are applied. My problem was that the line endings
changed as soon as I rebased, without me knowing.

newville:
@tritemio, @andyfaff    As I think about this more, I think that auto-converting file-endings is not the right thing to do (either as `autocrlf=True` or `autocrlf=input`).   The data in a repository are not OS dependent, and IMHO the programs interacting with the repository should not be either.   One can (and I do, really!)  work with the same copy of a repository from different OSes (say, over CIFS-mounted or WebDAV/Dropbox shares).  So, I don't fully agree that git ensures "the correct" line endings are applied.  Rather, it ensures that the line-endings corresponding to the OS used to access the repository  are applied.   But since multiple OSes can access the same repository, such "automatic" line endings will vary.   If one is working with exactly one OS per repository,  it might be OK.  But I think it is not a good default behavior.  

Like I said, I'm completely used to line endings being mixed and I don't find "line endings changed" messages to be particularly bothersome.   So, I don't actually have a very strong opinion on this.    But, I'm slightly in favor of the explicit `text eol=lf` and `autocrlf=False`.

tacaswell:
Line ending thrashing makes forensics via git blame difficult.

About once or twice a month I do this sort of thing to track down _why_ the code is the way it is in matplotlib. Knowing the context of a code change can be very informative.

Issue #259, opened 2015-10-05
Speed issue in v0.9.1


licode:
I did a test to compare new version and v0.8.3 as the following example.

```
def gaussian(x, amp, cen, wid):
    return amp * np.exp(-(x-cen)**2 /wid)

t1 = time.time()
gmode = None

for i in range(300):
    prename = 'g'+str(i)+'_'
    gmode1 = lmfit.Model(gaussian, prefix=prename)
    if gmode is None:
        gmode = gmode1
    else:
        gmode += gmode1
        
t2 = time.time()
p = gmode.make_params()
t3 = time.time()
copy.deepcopy(p)
t4 = time.time()

print(t2-t1)
print(t3-t2)
print(t4-t3)
```

result for v0.9.1 is
```
20.6462230682
1.95246195793
1.74573016167
```

while result for v0.8.3 is
```
0.400651931763
0.0140390396118
0.0666930675507
```

Any reason why the new version is slow in building model and creating parameters? I checked that lots of deepcopy are used in order to keep input parameters separated from output parameters. 

I wrote some complex model to do fitting with many gaussian functions, but right now it becomes extremely slow with the new version. Anyway we can turn this deepcopy off based on user's requirement? I can work on this if we care more about the speed issue.

newville:
@licode   Hmm,  I'm not sure, but I'm also not sure it's only due to deepcopy, but I suspect (but have not confirmed) that this slowdown probably happened around #214.    

That is, Parameters() got more expensive to create, and each Model creates an OrderedSet() for its parameter names.   I think both of these are contributing.   But, deepcopy is definitely suspicious too!

I'm on travel at the moment, and just don't have time to look into this further.    Let us know what you figure out. 

Perhaps we need a longer release candidate time scale.  We're clearly missing important problems.

tacaswell:
@licode You should use the `%prun` ipython magic to investigate what changed.

tritemio:
@licode @tacaswell another very useful tool is `line_profiler`. Pip install it an it enables `%lprun` magic to have line-by-line performance stats.

andyfaff:
I've been looking at overhead in Parameter._getval() recently. Testing
indicates that the total time required to obtain parameter values is a
significant fraction (~1/3) of the time it takes to calculate the objective
function.
It might be worth you looking there as well.

On 6 October 2015 at 11:51, Antonino Ingargiola <notifications@github.com>
wrote:

[...]

PR #261, opened 2015-10-07
MAINT: speedup Parameter.__deepcopy__ and asteval.__init__

andyfaff:
Recent testing (#259) has revealed that `deepcopy` of Parameters is very slow. There are a couple of reasons for this.  Firstly, there was a significant period of time taken for determining symbols added by the user in _asteval.symtable, compared to when the interpreter is initialised.  This problem was solved using `set.difference`.
Secondly, there was a serious performance penalty in __deepcopy__ because whenever a `Parameter` is added to `_pars` the Parameters are forced to undergo a `update_constraints` via `__setitem__`. If 200 parameters are added then for each of the 200 parameters that you add you have to call `update_constraints`. `update_constraints` iterates through every parameter in a loop. This loop gets longer and longer everytime a Parameter is added. This problem has been solved by adding an attribute `_dont_update_constraints`.  Whenever lots of parameters are going to be added at once this should be set to `True`, once the addition has finished this should be set to `False` and a manual call of `update_constraints` should be made. This is a _lot_ more efficient.
Thirdly, the `asteval.Interpreter` construction has been made more efficient via dictionary comprehensions.

### DIFF
--- a/lmfit/parameter.py
+++ b/lmfit/parameter.py
@@ -34,7 +34,6 @@ class Parameters(OrderedDict):
 
     add()
     add_many()
-    add_many_parameters()
     dumps() / dump()
     loads() / load()
     """
@@ -61,8 +60,6 @@ class Parameters(OrderedDict):
         _pars._asteval.symtable.update(unique_symbols)
 
         # we're just about to add a lot of Parameter objects to the newly
-        # created object. Temporarily switch off updating_constraints().
-        # This causes a O(N**2) slowdown.
         parameter_list = []
         for key, par in self.items():
             if isinstance(par, Parameter):
@@ -77,7 +74,7 @@ class Parameters(OrderedDict):
                 param.expr = par.expr
                 parameter_list.append(param)
 
-        _pars.add_many_parameters(parameter_list)
+        _pars.add_many(*parameter_list)
 
         return _pars
 
@@ -92,15 +89,13 @@ class Parameters(OrderedDict):
         par._expr_eval = self._asteval
         self._asteval.symtable[key] = par.value
 
-        self.update_constraints()
-
     def __add__(self, other):
         "add Parameters objects"
         if not isinstance(other, Parameters):
             raise ValueError("'%s' is not a Parameters object" % other)
         out = deepcopy(self)
         params = other.values()
-        out.add_many_parameters(params)
+        out.add_many(*params)
         return out
 
     def __iadd__(self, other):
@@ -108,25 +103,23 @@ class Parameters(OrderedDict):
         if not isinstance(other, Parameters):
             raise ValueError("'%s' is not a Parameters object" % other)
         params = other.values()
-        self.add_many_parameters(params)
+        self.add_many(*params)
         return self
 
     def update_constraints(self):
-        """update all constrained parameters, checking that
-        dependencies are evaluated as needed.
         """
-        if self._dont_update_constraints:
-            return
-
+        Update all constrained parameters, checking that dependencies are
+        evaluated as needed.
+        """
         _updated = []
         def _update_param(name):
-            """update a parameter value, including setting bounds.
+            """
+            Update a parameter value, including setting bounds.
             For a constrained parameter (one with an expr defined),
             this first updates (recursively) all parameters on which
             the parameter depends (using the 'deps' field).
             """
             # Has this param already been updated?
-            # if this an expression dependency, it may have been
             if name in _updated:
                 return
             par = self.__getitem__(name)
@@ -178,41 +171,32 @@ class Parameters(OrderedDict):
         Parameters
         ----------
         parlist : sequence
-        A sequence of tuples, each containing at least the name. The order in
-        each tuple is the following:
-            name, value, vary, min, max, expr
+            A sequence of tuples, or a sequence of `Parameter` instances. If it
+            is a sequence of tuples, then each tuple must contain at least the
+            name. The order in each tuple is the following:
+
+                name, value, vary, min, max, expr
 
         Example
         -------
         p = Parameters()
+        # add a sequence of tuples
         p.add_many( (name1, val1, True, None, None, None),
                     (name2, val2, True,  0.0, None, None),
                     (name3, val3, False, None, None, None),
                     (name4, val4))
 
+        # add a sequence of Parameter
+        f = Parameter('name5', val5)
+        g = Parameter('name6', val6)
+        p.add_many(f, g)
         """
-        parameter_list = []
         for para in parlist:
-            param = Parameter(*para)
-            parameter_list.append(param)
-        self.add_many_parameters(parameter_list)
-
-    def add_many_parameters(self, parameter_list):
-        """
-        Convenience method for adding many Parameter items
-
-        Parameters
-        ----------
-        parameter_list : sequence of lmfit.Parameter objects
-        """
-        try:
-            self._dont_update_constraints = True
-            for param in parameter_list:
-                if isinstance(param, Parameter):
-                    self.__setitem__(param.name, param)
-        finally:
-            self._dont_update_constraints = False
-            self.update_constraints()
+            if isinstance(para, Parameter):
+                self.__setitem__(para.name, para)
+            else:
+                param = Parameter(*para)
+                self.__setitem__(param.name, param)
 
     def valuesdict(self):
         """
@@ -302,6 +286,7 @@ class Parameters(OrderedDict):
         dump(), loads(), json.load()
         """
         return self.loads(fp.read(), **kws)
+
 
 class Parameter(object):
     """
--- a/lmfit/parameter.py
+++ b/lmfit/parameter.py
@@ -485,10 +485,17 @@ class Parameter(object):
 
     def _getval(self):
         """get value, with bounds applied"""
-        if (self._val is not nan and
-            isinstance(self._val, uncertainties.Variable)):
+
+        # Note assignment to self._val has been changed to self.value
+        # The self.value property setter makes sure that the
+        # _expr_eval.symtable is kept updated.
+        # If you just assign to self._val then
+        # _expr_eval.symtable[self.name]
+        # becomes stale if parameter.expr is not None.
+        if (isinstance(self._val, uncertainties.Variable)
+            and self._val is not nan):
             try:
-                self._val = self._val.nominal_value
+                self.value = self._val.nominal_value
             except AttributeError:
                 pass
         if not self.vary and self._expr is None:
@@ -498,25 +505,22 @@ class Parameter(object):
         if not hasattr(self, '_expr_ast'):
             self._expr_ast = None
         if self._expr_ast is not None and self._expr_eval is not None:
-            self._val = self._expr_eval(self._expr_ast)
+            self.value = self._expr_eval(self._expr_ast)
             check_ast_errors(self._expr_eval)
 
         if self.min is None:
             self.min = -inf
         if self.max is None:
-            self.max =  inf
+            self.max = inf
         if self.max < self.min:
             self.max, self.min = self.min, self.max
-        if (abs((1.0*self.max - self.min)/
+        if (abs((1.0 * self.max - self.min)/
                 max(abs(self.max), abs(self.min), 1.e-13)) < 1.e-13):
             raise ValueError("Parameter '%s' has min == max" % self.name)
         try:
-            if self.min > -inf:
-                self._val = max(self.min, self._val)
-            if self.max < inf:
-                self._val = min(self.max, self._val)
+            self.value = max(self.min, min(self._val, self.max))
         except(TypeError, ValueError):
-            self._val = nan
+            self.value = nan
         return self._val
 
     def set_expr_eval(self, evaluator):
--- a/lmfit/parameter.py
+++ b/lmfit/parameter.py
@@ -64,22 +64,21 @@ class Parameters(OrderedDict):
         # we're just about to add a lot of Parameter objects to the newly
         # created object. Temporarily switch off updating_constraints().
         # This causes a O(N**2) slowdown.
-        _pars._dont_update_constraints = True
-
+        parameter_list = []
         for key, par in self.items():
             if isinstance(par, Parameter):
-                name = par.name
-                _pars.add(name, value=par.value, min=par.min, max=par.max)
-                _pars[name].vary = par.vary
-                _pars[name].stderr = par.stderr
-                _pars[name].correl = par.correl
-                _pars[name].init_value = par.init_value
-                _pars[name].expr = par.expr
-                _pars._asteval.symtable[name] = par.value
+                param = Parameter(name=par.name,
+                                  value=par.value,
+                                  min=par.min,
+                                  max=par.max)
+                param.vary = par.vary
+                param.stderr = par.stderr
+                param.correl = par.correl
+                param.init_value = par.init_value
+                param.expr = par.expr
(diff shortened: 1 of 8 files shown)


lmfit/parameter.py
```diff
@@ -38,15 +40,32 @@ class Parameters(OrderedDict):
...
+        sym_from_construction = set(_pars._asteval.no_deepcopy)
+        sym_unique = sym_in_current.difference(sym_from_construction)
+        unique_symbols = {key: deepcopy(self._asteval.symtable[key], memo)
+                          for key in sym_unique}
+        _pars._asteval.symtable.update(unique_symbols)
+
+        # we're just about to add a lot of Parameter objects to the newly
+        # created object. Temporarily switch off updating_constraints().
+        # This causes a O(N**2) slowdown.
+        _pars._dont_update_constraints = True
```
tacaswell: I might be better (without really understanding this code) to put a bit more state into `_pars` and have it keep track if if it is 'stale', and only trigger the update when a call needs to know what the constraints are.

tacaswell: and if you are going to do it this way I would suggest giving `_pars` a context manager so that if something errors out inside, it will fall back to a safe state.

andyfaff: I understand what you're saying with respect to triggering an update. The only places where this *could* be done is in `Parameter._getval` or the value setter `Parameter.data`.  However, this is a very bad idea as both of those methods have to perform very efficiently during a minimization because the parameter values are constantly being changed and extracted. It's much better to take the hit where it is I think.

andyfaff:
@tacaswell, what do you think of the latest commit?

andyfaff:
Travis fails due to #262

newville:
@andyfaff  Thanks so much for this!  That's a huge amount of work and progress.    I'm on travel at an attention-intense review meeting so can't comment in as much detail as I'd like, or try this out at the moment.  Offhand, I'm not sure why #262 would crop up from this, but I'm sure it's solvable.

andyfaff:
The problem is latent in the master and only becomes obvious when you query
the fwhm value.

newville:
@andyfaff Please fix `test_change_prefix` to not fail so that this PR passes all tests.  

Please replace `OrderedSet` with `set` in model.py.

andyfaff:
> Please fix test_change_prefix to not fail so that this PR passes all tests.

Do you mean mark the test as a knownfail, or do you want me to fix the prefix change itself? The first I can do, the second will be difficult because I don't know the `Model` code.

andyfaff:
> Please replace OrderedSet with set in model.py.

Do you mean replace `self._param_names = OrderedSet()` with `self._param_names = set()`?

newville:
@andyfaff  For the test, I would say to change it to do nothing but pass -- we'll change the test very soon when we make `prefix` immutable, but that might mean more discussion about what to do about copying parameter hints.   I still don't understand why your changes would prompt it to fail now, but the test wasn't really useful anyway.    But, of course, we're not going to merge a PR that fail tests, so the test has to be fixed to move forward.

For OrderedSet():  I know this has been suggested before. Yes: replace all instances of `OrderedSet()` with `set()` in `model.py`.   That will actually give a much larger absolute speed-up for @licode's original report, reducing the 20 seconds for peak creation to around 2 seconds.   If you don't want to do this, it's fine, we'll do it separately.

andyfaff:
Working on the OrderedSet problem now, I *may* be able to fix it.

andyfaff:
The latest timing for the sample problem in #259 is:

    0.38925909996032715
    0.031076908111572266
    0.02474808692932129

Notes: 

* `Model.param_names` is no longer an `OrderedSet`
* `test_change_prefix` in test_model.py will emit a warning when running tests because of an issue with changing model prefix and parameter constraints.

danielballan:
Just chiming in to appreciate these efforts. Nice work, @andyfaff!

andyfaff:
@newville, the problem manifests itself now because I call `update_constraints` in `Parameters.add_many_parameters`. BTW, `add_many_parameters` is a new method, are you happy with the name?

danielballan:
@andyfaff Is it possible we could defer `update_contraints` in a more automatic way? Say, by computing the constraints in a property that gets marked as stale when a parameter is added? That way this performance-related implementation detail does not have to leak into and complicate the user API.

andyfaff:
@danielballan thank you for making me think twice. The reason why there was a large performance hit experienced in #259 was because `Parameters.update_constraints` was being called (in `Parameter.__setitem__` every time a Parameter was being added.  The `update_constraints` method is expensive, and when you add 300 parameters that takes a lot of time.

The solution I took was to add an attribute (`_dont_update_constraints`) in `Parameters` that tells the `update_constraints` method to do nothing. When the (new) `Parameter.add_many_parameters` is called it temporarily sets that flag whilst adding all the parameters, then clears it once they've all been added. Then `update_constraints` is called. Thus, when adding N parameters it only gets called once, rather than N times.

The alternative approach (that I didn't consider until our question prompted me) would be to remove `update_constraints` from `Parameters.__setitem__`. Thus, you could add as many parameters as you wanted without having to call that (relatively expensive) method. A flag could be set to say that the `Parameters` object is stale because Parameter's have been added.  The user of the `Parameters` object would then be responsible for calling `update_constraints` before use. (deepcopying and pickling Parameters are two key locations where lots of parameters are added).

I guess it's a design question. Would people prefer to have the `Parameters` constraints automatically evaluated when `Parameter` objects are added, or would they prefer to manually `update_constraints` as and when the `Parameters` object is used?

Either way I think the `add_many_parameters` method is a good addition.

newville:
@andyfaff @danielballan   Thanks!  This is really great work, and also a non-trivial amount of work diving deep in the code.    Now, if you'll allow some nit-picking and comments:

1.  A minor point:  I think it's easier to read names that are generally positive.  So `dont_update_constraints`   could be `delay_update_constraints`.  

2. Then again, I kind of think this whole mechanism isn't necessary at all.   That is, I think we do not want setting a parameter value to automatically trigger "update_constraints()" (even though, as you say, it is currently called in  `__setitem__()`).   Parameters really only needs to be consistent within a single fit iteration / step.  I think saying that Parameters always has to be consistent is just unnecessary, and that we  should probably never have `update_constraints()` called internally.  I could be persuaded to leave it as is (and I'm not sure it is needed for *this* PR), but I think that #259 points out that this is probably a bad idea, and I think I favor just not calling `update_constraints()` until it s absolutely necessary -- really, not until ready to really call an objective function.

3.  For `add_many_parameters()`, I think I would recommend this to be "semi-private", as it's not clear it is useful outside the Parameters class.  For the public API, wouldn't one just use "+" and really on `__add__()`/`__iadd__()`?   Internally,  it's used 3 places,  but if `update_constraints()` was not called in `__setitem__()`, it would amount to

```python
    def add_many_parameters(parameters):
          for par in parameters:
                if isinstance(par, Parameter):
                    self.__setitem__(par.name, par)
```
and the test of "am I a Parameter" does occur in `__setitem__()`, so it could just be
 
```python
    def _add_params(params):
          for par in params:  
                self.__setitem__(par.name, par)
```
Anyway, I think this would mostly simplify `Parameters`, which is probably good.   I'm not exactly certain about not automatically calling `update_constraints()` inside of `Parameters`, and would certainly want to hear what you and others think,  and not would not want it to delay this PR too much.    

Does *not* calling `update_constraints()` until necessary seem reasonable to you all?

andyfaff:
> For add_many_parameters(), I think I would recommend this to be "semi-private", as it's not clear it is useful outside the Parameters class. For the public API, wouldn't one just use "+" and really on __add__()/__iadd__()?

I think it's a natural pattern to want to add lots of `Parameter` instances to a `Parameters` object. It's a pattern that I use in my own code that uses lmfit, and it's a pattern that's used internally. I prefer it to `add_many`.  Besides one can't use `__add__` and `__iadd__` because they're for adding two `Parameters` instances, not adding a `Parameter` to `Parameters` (at least not without some modification).

Everyone agrees that `update_constraints` needs to be called before the objective function is called. It should only be called once per objective function evaluation, no more.

For me, on the whole I'm probably in favour of having the `Parameters` constraints as up to date as possible. (I use the `Parameters` class outside a specific minimization task to display parameters in a Qt GUI). The question I have is why wouldn't you want it as up to date as possible? I'm guessing that the answer would be that calling `update_constraints` is expensive.  My answer, in turn, would be that a Parameters instance is created once, but then that single instance is reused many times when calling the objective function.
For example, in the lifetime of a `Parameters` object `update_constraints` may be called once when it's created (by adding a list of `Parameter`), but then 200 000 times when the objective function is evaluated. The cost of having Parameter constraints up to date before a minimization takes place is well amortized over the total cost of minimization.
The flip side to my argument may be multiprocessing. If you want to parallelise for a CPU intensive operation you have to use multiprocessing. If you want to use multiprocessing you need to pickle the objective functions (the code that calls the objective functions) and `Parameters` instances. In the process of pickling/unpickling you probably wouldn't want to call `update_constraints` because it would have to be called again when injecting values into each `Parameter` (in a method like __residual).  I'd need to do some profiling to see if this is really a problem though.

newville:
On Tue, Oct 13, 2015 at 10:04 PM, Andrew Nelson <notifications@github.com>
wrote:

> For add_many_parameters(), I think I would recommend this to be
> "semi-private", as it's not clear it is useful outside the Parameters
> class. For the public API, wouldn't one just use "+" and really on *add*
> ()/*iadd*()?
>
> I think it's a natural pattern to want to add lots of Parameter instances
> to a Parameters object. It's a pattern that I use in my own code that
> uses lmfit, and it's a pattern that's used internally. I prefer it to
> add_many. Besides one can't use __add__ and __iadd__ because they're for
> adding two Parameters instances, not adding a Parameter to Parameters (at
> least not without some modification).
>

Right,  __add__ / __iadd__ are for combining two Parameters objects.

But `add_many_parameters()` adds a *list* of Parameter objects which seems
a little strange to me.  A Parameters is an ordered sequence of Parameter
objects, so when do you need (or have) a list of Parameter objects that is
not better handled as a Parameters instance?   I'm not sure we really need
two ways to add a sequence of Parameter objects to an existing Parameters
instance.

What would you say to having `Parameters.add()` allow its first argument to
be either a string or a Parameter, in which case it could at that existing
Parameter instead of making a new one?   That would allow you to add an
existing Parameter, though one at a time (and `add_many()` would then allow
you to add many, though with a relatively clunky syntax compared to list
comprehension).

In short, I think the existing API could encompass the desire to add an
existing Parameter without another `Parameter.add_*()` method.

> Everyone agrees that update_constraints needs to be called before the
> objective function is called. It should only be called once per objective
> function evaluation, no more.
>
> For me, on the whole I'm probably in favour of having the Parameters
> constraints as up to date as possible. (I use the Parameters class
> outside a specific minimization task to display parameters in a Qt GUI).
> The question I have is why wouldn't you want it as up to date as possible?
> I'm guessing that the answer would be that calling update_constraints is
> expensive. My answer, in turn, would be that a Parameters instance is
> created once, but then that single instance is reused many times when
> calling the objective function.
>
It's not so much a question of creation being expensive (well, after the
bug fixes!), but whether every time a Parameter value is set that requires
re-running `Parameter.update_constraints()`.   In current master, if you
have

```python
     params = Parameters()
     params.add('a', 0.0, vary=True)
     params.add('b', 2.0, vary=True)
     params.add('d', expr='a+b')
```

doing

```python
     params['a'].value = 1.0
```

does *not* cause `d` to be updated.    I think that would be necessary if
"as up to date as possible" is the expectation.

For example, in the lifetime of a Parameters object update_constraints may
> be called once when it's created (by adding a list of Parameter), but
> then 200 000 times when the objective function is evaluated. The cost of
> having Parameter constraints up to date before a minimization takes place
> is well amortized over the total cost of minimization.
>
 The parameters have to be up-to-date at each iteration.   I think the
point for me is that if you have

```python
     params = Parameters()
     params.add('a', 0.0 vary=True)
     params.add('b', 2.0, vary=True)
     params.add('c', 99.0, vary=True)
     params.add('d', expr='a+b')
```

In the next fit iteration, you're basically going to re-assign the values,
with something like:

```python
     params['a'].value = 0.1
     params['b'].value = 2.1
     params['c'].value = 99.8
     # maybe do:
     # params.update_constraints()
     call_users_objective_function(params, ....)
```

and the question is:  Should the value of 'd' be updated when 'a' is set,
**and** when 'b' is set, **and** when 'c' is set, or just once, before the
users function is called?   If keeping things as up to date as possible, it
sort of has to be called every time, even thought the objective function
will never see the partial set of values (and so 'd'  never really needs to
hold the value of '2.1', reflecting 'a=0.1, b=2.0', only '2.0' or '2.2').

Triggering `update_constraints()` on each assignment would ensure that

```python
>>> params['a'].value = 1
>>> params['b'].value = 1
>>> print(params['d'].value)   # does not currently work!
2
>>> params['b'].value = 2
>>> print(params['d'].value)  # does not currently work!
3
```

but will be a performance hit for combining models or adding Parameters
(each addition would trigger `update_constraints`) and for running fits
(`update_constraints` will be run Nvariables times per iteration instead of
once).    Or the code has to have special cases, and so get harder to
follow and maintain.

> The flip side to my argument may be multiprocessing. If you want to
> parallelise for a CPU intensive operation you have to use multiprocessing.
> If you want to use multiprocessing you need to pickle the objective
> functions (the code that calls the objective functions) and Parameters
> instances. In the process of pickling/unpickling you probably wouldn't want
> to call update_constraints because it would have to be called again when
> injecting values into each Parameter (in a method like __residual). I'd
> need to do some profiling to see if this is really a problem though.
>

I don't think Pickle or multiprocess make any difference.  If you have
constraint expressions, you have to call `update_constraints()` to get
up-to-date values.

I'm pretty sure that each iteration and call to `__residual(params)` has to
be done by a single process, and I'm not highly optimistic about
multiprocessing+fitting with Pickle. But that's probably a separate topic.

Cheers,

--Matt

andyfaff:
How about changing `add_many`, I like the following:

    def add_many(self, *parlist):
        """
        parameter_list = []
        for para in parlist:
            if isinstance(para, Parameter):
                self.__setitem__(para.name, para)
            else:
                param = Parameter(*para)

> It's not so much a question of creation being expensive (well, after the bug fixes!)

`update_constraints` is still very expensive. It takes up approximately 30-40% of the runtime (calculated for the emcee PR). I'll put up the profiles somewhere, perhaps some optimisation can be done there.

> but whether every time a Parameter value is set that requires re-running `parameter.update_constraints()`.

I appreciate the point you're making. I would like that 'as up to date as possible', I have some ideas of how that could be approached. However, I certainly wouldn't want that if it created a performance penalty.

I can change the PR to modify `add_many` relatively simply, I'll remove `update_constraints` from __setitem__ while I'm at it.  Then hopefully we should be good to go??

andyfaff:
Please see revision. 

*`add_many_parameters` has been removed, that functionality has been incorporated in `add_many`. 
* `update_constraints` has been removed from `Parameters.__setitem__`. It's the users responsibility to call this method before using the Parameters, otherwise all the constraints may be stale.
* The previous item required are several changes to test_algebraic_constraint, test_algebraic_constraint2 and test_params_set.

All tests pass.

newville:
@andyfaff   OK, I'll merge this.  I might then make a few tweaks of my own, including allowing `Parameter.add()` to take a `Parameter`, but Thanks!

I still do not see the use case for "list of Parameter objects" instead of "Parameters", but I guess there's not much harm in that.

Issue #265, opened 2015-10-09
BUG: parameters not being updated properly


andyfaff:
With current master there is a bug when setting parameter values, if there are constraints and limits.  Take the following example below.
`a` is always supposed to be twice `b` and it has no limits. `b` is clamped between 0 and 20.  Whenever I set `b` to a new value, and `b` is within its limits, `a` is recalculated properly. If I set `b` outside it's limits, then `b` is clamped to `b.max` correctly. It follows that `a` should be twice `b.max`.  However, this isn't the case.  I have a fix for this, it'll be included in #261

    import numpy as np
    from lmfit import Parameter, Parameters
    from numpy.testing import assert_, assert_almost_equal

    p = Parameters()

    p['a'] = Parameter('a', 10, True)
    p['b'] = Parameter('b', 10, True, 0, 20)

    p['a'].expr = '2 * b'
    assert_almost_equal(p['a'].value, 20)

    p['b'].value = 15
    assert_almost_equal(p['b'].value, 15)
    assert_almost_equal(p['a'].value, 30)

    p['b'].value = 30
    assert_almost_equal(p['b'].value, 20)
    assert_almost_equal(p['a'].value, 40)

    ---------------------------------------------------------------------------
    AssertionError                            Traceback (most recent call last)
    <ipython-input-6-dd8c6e7c13e7> in <module>()
          1 p['b'].value = 30
          2 assert_almost_equal(p['b'].value, 20)
    ----> 3 assert_almost_equal(p['a'].value, 40)

    /Users/anz/miniconda3/envs/dev3/lib/python3.4/site-packages/numpy/testing/utils.py in assert_almost_equal(actual, desired, decimal, err_msg, verbose)
        508         pass
        509     if round(abs(desired - actual), decimal) != 0:
    --> 510         raise AssertionError(_build_err_msg())
        511 
        512 

    AssertionError: 
    Arrays are not almost equal to 7 decimals
     ACTUAL: 60
     DESIRED: 40

andyfaff:
I'll also include this test in the PR

newville:
@andyfaff   Yes,  you have to run `p.update_constraints()` to guarantee that constrained parameters have up-to-date values.  That is potentially expensive, so within a fitting loop, it doesn't really make sense to run that on every assignment -- it just needs to be run once after assigning all variable parameters values.  Since "within a fitting loop" is the main point, that is the default behavior.

[...]

PR #266, opened 2015-10-09
make sure update_constraints() is called

newville:
make sure update_constraints() is called in residual function before calling users objective function.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -4,6 +4,7 @@ include requirements.txt
 exclude *.pyc core.* *~ *.pdf
 recursive-include lmfit *.py
 recursive-include tests *.py *.dat
+recursive-include NIST_STRD *.dat
 recursive-include doc *
 recursive-exclude doc/_build *
 recursive-exclude doc *.pdf
--- a/lmfit/minimizer.py
+++ b/lmfit/minimizer.py
@@ -254,6 +254,7 @@ class Minimizer(object):
             params[name].value = params[name].from_internal(val)
         self.result.nfev = self.result.nfev + 1
 
+        params.update_constraints()
         out = self.userfcn(params, *self.userargs, **self.userkws)
         if callable(self.iter_cb):
             abort = self.iter_cb(params, self.result.nfev, out,


andyfaff:
Before you merge this I think there may be another way of doing this, one that works for a Parameters instance in isolation.

In `Parameter._getval` there are a couple of places where `Parameter._val` is assigned. The relevant lines being:

    self._val = self._expr_eval(self._expr_ast)

and
    
    if self.min > -inf:
        self._val = max(self.min, self._val)
    if self.max < inf:
        self._val = min(self.max, self._val)

My suggestion is that these lines are changed to:

    self.value = self._expr_eval(self._expr_ast)

and

    if self.min > -inf:
        self.value = max(self.min, self._val)
    if self.max < inf:
        self.value = min(self.max, self._val)

(Actually wouldn't

    self.value = max(self.min, min(self._val, self.max))

be faster to use here). The reason being that assigning to self.value also updates the `Parameters._asteval.symtable`. So when the value of a dependent parameter is next retrieved it'll be updated.  The real test is more complicated chained expressions.  SImilar to the example I gave, but when:

    a.expr = `2 * b`
    b.expr = `2 * c`

and you then change `c`.

andyfaff:
I agree with `update_constraints` being expensive. You would only want to call that once per iteration.

andyfaff:
So I just repeated the test outlined in #265, and it still fails even if `update_constraints` is called beforehand.  I think that `update_constraints` still needs to be called in `__residual`, but that my suggestions are the only way to fix the problem.

newville:
@andyfaff   OK, I think there are a lot of interrelated issues and PRs, and the level of confusion is quite high.   I can't tell whether your 2nd post amends or replaces the 1st.     Are you still recommending the suggestions in the first PR?

The 3rd post is about #265 is wrong.  I will respond there.

I still don't understand why tests that work in master are failing in #261, and the discussion of `Model.prefix` and `Model.copy()` can't really go forward unless it is mixed with that PR, or the whole thing is reworked.   I'd like to aim for a little more care and clarity and try to solve one problem at a time.

newville:
@andyfaff  we can still think about whether we should be setting `.value`, but for now I just wanted to clean up the actual bug, so we can move on the bigger and better things (like #261 !)

andyfaff:
So why hadn't the bug reared itself in the wild? Can a test be added for
that?

newville:
@andyfaff I *believe* that the fact that the value of an expression-based parameter might be "stale" (say, evaluated with the previous iteration of its dependencies) is not significant as all the values actually converge.  I'm not 100% certain of this, and definitely seems better to me to explicitly update the constraints.   

Updating the asteval symbol table for each new parameter value seems fine, but it doesn't actually remove the need for `update_constraints()`, which takes care nested dependencies (or tries very hard to do so).

PR #267, opened 2015-10-11
Added moffat function MoffatModel and appropriate documentation.

Stuermer:
A Moffat is commonly used in astronomy to describe point spread functions. However, the profile is generally useful for lines that have larger wings than a Lorentzian.

Not yet implemented: Better guess parameters.
FWHM expression in the Model.

I coudln't find out how to correctly add the FWHM expression in this case. As written in the documentation it is \sigma 2 \sqrt{2^{1/\beta}}-1.

It's the first time I use github, sorry if I did something wrong.

### DIFF
--- a/doc/builtin_models.rst
+++ b/doc/builtin_models.rst
@@ -147,6 +147,26 @@ where :math:`\sigma_g = {\sigma}/{\sqrt{2\ln{2}}}` so that the full width
 at half maximum of each component and of the sum is :math:`2\sigma`. The
 :meth:`guess` function always sets the starting value for ``fraction`` at 0.5.
 
+
+:class:`MoffatModel`
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. class:: MoffatModel(missing=None[, prefix=''[, name=None[, **kws]]])
+
+a model based on a `Moffat distribution function
+<https://en.wikipedia.org/wiki/Moffat_distribution>`_, the parameters are
+``amplitude`` (:math:`A`), ``center`` (:math:`\mu`),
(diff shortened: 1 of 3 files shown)


newville:
@Stuermer Thanks (and sorry for delay in responding).   Of course, I'd never heard of this one, so am very excited to try it out!  In fact, I was just looking at some X-ray images today that looked like it had extra long tails.

To set the FWHM, you can put a parameter hint in `MoffatModel.__init__()`.     The existing examples are probably too simple to be a good guide, but this should be (are you sure that is right?  It looks somehow like you might have a misplaced brace, and that it might be 
```latex
 2\sigma\sqrt{ 2^{1/\beta} -1}
```

in which case you could have

```python
self.set_param_hint('fwhm', expr="2*%ssigma*sqrt(2**(1.0/%sbeta)-1)" % (self.prefix, self.prefix))
```

(well, unless I got that wrong).   That is, you can "just" set the expression for FWHM, and be sure to allow for prefixes for the parameter names.

Stuermer:
@newville  You're right of course, I had a misplaecd brace. I fixed it in the documentation and added the FWHM expression to the moffatModel. It seems to work !  I didn't understand the syntax regarding the prefix, but now it is obsious. Thanks for pointing it out.
I use the Moffat to fit fabry-perot etalon lines in a laser spectroscopy experiment. Due to several reasons the lineshape is not a perfect lorentzian. A pseudovoigt/voigt also works well, but the moffat seem to beat them in terms of speed (which is somewhat critical in my application).  I'm glad it might be helpful for you, too.

newville:
@Stuermer  Thanks!

PR #268, opened 2015-10-11
Example how to use analytic jacobians together with built-in models.

Stuermer:
For a Gaussian and a Lorentzian analytic jacobians are implemented. Resulting fits show an improvement in the number of evaluations needed to find a solution.


newville:
Thanks!

PR #269, opened 2015-10-15
Conda recipe

yoavram:
> Disclaimer: I am not sure if this will interest the lmfit community, but I needed it for my own project and thought that I'd share. If it's of no interest, feel free to close without merging. It will still be a good documentation of how to do this.

I created a conda recipe for lmfit using the [PyPi skeleton](http://conda.pydata.org/docs/build_tutorials/pkgs.html#build-a-simple-package-with-conda-skeleton-pypi) (on Windows, these commands must run with administrator privilege and after installing `conda install conda-build unxutils`):
```
>>> conda skeleton pypi lmfit --output-dir conda-recipe
```
To build the conda pakcage, from the root folder of the lmfit repo:
```
>>> conda build conda-recipe
```
To install, after building:
```
>>> conda install --use-local lmfit
```
Then check it works:
```
>>> python -c "import lmfit;print(lmfit.__version__)"
0.9.1
```
I uploaded the package I built to anaconda.org: https://anaconda.org/yoavram/lmfit. 
It should work for Python 3.4 for all platforms supported by conda (I also used `conda conda convert -f --platform all`), so one should be able to install lmfit with:
```
>>> conda install -c https://conda.anaconda.org/yoavram lmfit
```

### DIFF
--- a/conda-recipe/bld.bat
+++ b/conda-recipe/bld.bat
@@ -1,0 +1,8 @@
+"%PYTHON%" setup.py install
+if errorlevel 1 exit 1
+
+:: Add more build steps here, if they are necessary.
+
+:: See
+:: http://docs.continuum.io/conda/build.html
+:: for a list of environment variables that are set during the build process.
--- a/conda-recipe/build.sh
+++ b/conda-recipe/build.sh
@@ -1,0 +1,9 @@
+#!/bin/bash
+
+$PYTHON setup.py install
+
+# Add more build steps here, if they are necessary.
+
+# See
+# http://docs.continuum.io/conda/build.html
+# for a list of environment variables that are set during the build process.
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -1,0 +1,67 @@
+package:
+  name: lmfit
+  version: "0.9.1"
+
+source:
+  fn: lmfit-0.9.1.tar.gz
+  url: https://pypi.python.org/packages/source/l/lmfit/lmfit-0.9.1.tar.gz
+  md5: 61c7ec515d324663b9fb6ed59d0568e3
+#  patches:
+   # List any patch files here
+   # - fix.patch
+
+# build:
+  # noarch_python: True
+  # preserve_egg_dir: True
+  # entry_points:
+    # Put any entry points (scripts to be generated automatically) here. The
+    # syntax is module:function.  For example
+    #
+    # - lmfit = lmfit:main
+    #
+    # Would create an entry point called lmfit that calls lmfit.main()
+
+
+  # If this is a new build for the same version, increment the build
+  # number. If you do not include this key, it defaults to 0.
+  # number: 1
+
+requirements:
+  build:
+    - python
+    - setuptools
+    - numpy
+    - scipy
+
+  run:
+    - python
+    - numpy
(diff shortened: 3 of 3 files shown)


newville:
@yoavram  I don't use conda recipes much myself, and am slightly hazy on whether (and why) `pip install` would be inappropriate for anaconda.

Anyone else have an opinion?

FWIW, 0.9.1 has known problems, and we'll probably push a 0.9.2 early next week, so a 0.9.1-specific recipe will probably be need to be replaced soon.   Should the recipe be "latest version"?

tritemio:
@yoavram I'm swamped at the moment. I did a recipe for 0.8.3 https://github.com/tritemio/lmfit-conda.recipe. Can you check if there is any difference?

@newville A conda package is convenient if you have other conda packages depending on lmfit. This way you install/update all with a single command and you declare dependencies in a single place (conda manages well hard dependencies like pytables on all platforms). Also, building a package for a platform will run the tests on that platform, potentially catching issues. Not a huge advantages but it can be handy.

newville:
@tritemio  Ah, good point that other conda packages could then rely on conda to install lmfit.   That's  a compelling reason.   

Having no conda experience, I'm not sure I would be able to maintain a complicated recipe, but this doesn't look too bad.   One question for the conda recipe writers:  Is it better to grab the package from pypi (as @yoavram does) or from a github tag (as @tritemio did)?  Is there a way for a recipe to grab "latest github release"?  I think that would be preferable, and would imagine others would want that too.

Anyone have any objections to adding this?

tritemio:
@newville what I do with my own conda packages is grabbing the latest version on master from github and adding the number of commits since latest tag as build number (the latest number in the package name). This way the most recent commit has the priority version-wise (unless you rebase). It is not exactly kosher but works for me. Example on how to do it:

https://github.com/tritemio/FRETBursts/blob/master/conda.recipe/meta.yaml

And, BTW, that package has lmfit as dependency.

yoavram:
I think that it becomes even more useful if it's incorporated into the CI, so for example, travis-ci could be set to build the package (instead of doing `python setup.py install`, do `conda build conda-recipe & conda install --use-local lmfit`) and then at the deploy stage, deploy to anaconda.org (formerly known as binstar).
Another alternative is to use [anaconda.org CI](http://docs.anaconda.org/examples.html#ContinuousIntegrationRunABuildOnGitPush), I'm not sure what's better and easier.

newville:
@yoavram  Sorry to be dense, but I'm not sure I follow.  What is "it" that becomes more useful if it's incorporated into the CI?  You mean the conda recipe becomes more useful?  

I'm not sure I see a reason to stop using Travis CI or change the travis script, but I could be missing something.

yoavram:
Yes, the recipe is more useful if its used by Travis because otherwise deploying those conda packages becomes tedious, as one needs to setup an environment for each python version. I'm working on setting Travis to build and deploy my own project as a conda package; if im successful, I would be happy to do the same process for lmfit.

newville:
@yoavram I'll merge this as is, and perhaps we can go on from there.  I don't quite understand the point of using anaconda.org CI or changing the Travis CI script. It's not that I'm necessarily opposed to such an idea, but I think you have to  make a case about why it would be better *for lmfit*.

Issue #263, opened 2015-10-08
API change for conf_interval()


tritemio:
In lmfit 0.8.3 we could do:

```
fit_res = model.fit(y, x=x)
ci = lm.conf_interval(fit_res)
```

This raises an error in lmfit 0.9.1. I couldn't find any mention to a changed API for `conf_interval` in the docs.

We should either add a note to the docs or restore the old API.

newville:
@tritemio   definitely needs a doc change.  Sorry I forgot to do this.  The API had to be changed to work with the new `MinimizerResult`.

tritemio:
@newville no need to apologize. How do we call `lm.conf_interval` with results from a  model fit in lmfit 0.9?

licode:
probably you can see https://github.com/lmfit/lmfit-py/issues/251

tritemio:
@licode, my question here is how to compute `lm.conf_interval` when using `lm.Model` to fit, like in the first example in this issue.

newville:
@tritemio  actually @licode  gave you a very good hint. 

```python
ci = lm.conf_interval(fit_res, fit_res)
```
is probably what you want.  I suppose in  that sense, we could allow the second argument to be `None` as long as the first argument had `chisqr` and `params` attributes.....   Or, well, probably adding a `conf_interval` method to `ModelFit` is simpler....

tritemio:
@newville, ok thanks, it wasn't obvious to me without looking at the implementation. 

Anyway, if it's only an implementation detail I would prefer maintaining the old API, less broken scripts.

newville:
@tritemio it's not quite an implementation detail that 

```python
ci = lm.conf_interval(fit_res, fit_res)
```
works.   The 1st arg of `conf_interval` needs to be something that subclasses `Minimizer`, so that `.leastsq()` can be run.  The 2nd arg needs to be something that mimics a `MinimizerResult` in that it has the updated `params` (with best values and uncertainties), and has a `chisqr`.     It's sort of accidental that `ModelResult` satisfies both these needs.    Then again, it was sort of accidental that 
`conf_interval(fit_res)` worked in 0.8.3.

[...]

Issue #262, opened 2015-10-07
Bug when model prefix is changed.


andyfaff:
When making changes to `Model.make_params` I came across the following bug. It's based on `test_change_prefix` in `test_model.py`. It seems that when the prefix is changed the hints aren't updated to account for this.

    >>> from lmfit.models import GaussianModel
    >>> mod = GaussianModel(prefix='b')
    >>> mod.prefix = 'c'
    >>> params = mod.make_params()
    >>> names = params.keys()
    >>> all_begin_with_c = all([n.startswith('c') for n in names])
    >>> all_begin_with_c
    True
    >>> print(params)

    NameError
       <_ast.Module object at 0x108ecfc88>
                  ^^^
    name 'bsigma' is not defined
    ---------------------------------------------------------------------------
    NameError                                 Traceback (most recent call last)
    <ipython-input-4-73ac4b936803> in <module>()
    ----> 1 print(params)

    /Users/anz/miniconda3/envs/dev3/lib/python3.4/reprlib.py in wrapper(self)
         22             repr_running.add(key)
         23             try:
    ---> 24                 result = user_function(self)
         25             finally:
         26                 repr_running.discard(key)

    /Users/anz/miniconda3/envs/dev3/lib/python3.4/collections/__init__.py in __repr__(self)
        200         if not self:
        201             return '%s()' % (self.__class__.__name__,)
    --> 202         return '%s(%r)' % (self.__class__.__name__, list(self.items()))
        203 
        204     def __reduce__(self):

    /Users/anz/miniconda3/envs/dev3/lib/python3.4/site-packages/lmfit/parameter.py in __repr__(self)
        374         if self.name is not None:
        375             s.append("'%s'" % self.name)
    --> 376         sval = repr(self._getval())
        377         if not self.vary and self._expr is None:
        378             sval = "value=%s (fixed)" % (sval)

    /Users/anz/miniconda3/envs/dev3/lib/python3.4/site-packages/lmfit/parameter.py in _getval(self)
        452         if self._expr_ast is not None and self._expr_eval is not None:
        453             self._val = self._expr_eval(self._expr_ast)
    --> 454             check_ast_errors(self._expr_eval)
        455 
        456         if self.min is None:

    /Users/anz/miniconda3/envs/dev3/lib/python3.4/site-packages/lmfit/parameter.py in check_ast_errors(expr_eval)
         19     """check for errors derived from asteval"""
         20     if len(expr_eval.error) > 0:
[...]

PR #270, opened 2015-10-19
Model conf_interval

newville:
This started as a PR to add  `Model.conf_interval` to address #263, but sort of morphed into a general PR for 0.9.2, trying to address #262, better deal with #265, and even do some cleanup related to #261.

Features of this PR include

 * `Model.conf_interval()` and `Model.ci_report()` added.
 * `ci_report()` (for all confidence intervals) now shows differences from best-fit value by default, 
and has options of `with_offset=True, ndigits=5` to control whether the best-fit value is offset from other interval values, and how many digits to show.  It's also now in the documentation.
 * make `Model.prefix` immutable,  make `Model.copy()` raise `NotImplemented`.
 * `Model._param_names` is now a plain list (was `OrderedSet`, then `set`), which even further improves the performance of @licode #259 example.
 * `Parameters.add()` can take an existing `Parameter`.

Sorry, it is kind of a disjointed set of ideas, and so is totally fair to comment on these individually.

Anyway, I think this would make 0.9.2.    Comments, complaints, suggestions greatly appreciated!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ python:
     - 3.4
 
 before_install:
-    - wget http://repo.continuum.io/miniconda/Miniconda-latest-Linux_x86_64.sh -O miniconda.sh
+    - wget http://repo.continuum.io/miniconda/Miniconda-latest-Linux-x86_64.sh -O miniconda.sh
     - bash miniconda.sh -b -p $HOME/minicoda
     - export PATH=$HOME/miniconda/bin:$PATH
     - hash -r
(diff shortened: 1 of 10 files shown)


newville:
If there are no objections, I'll merge and tag this as 0.9.2 in a day or so.

andyfaff:
The tests for my package to work with master.

newville:
OK, this now addresses #271.   Also, the TravisCI scripts were apparently in need of updating, and are now better.

@tritemio, @andyfaff can you verify that this is good for you?   If so, and if there are no other issues or objections, I'll merge this, then merge @tritemio's test, and call that 0.9.2.

newville:
OK, this time for sure!  Planning to merge late Monday (US Central time) unless more mistakes found.

PR #274, opened 2015-10-27
Model conf interval (round2)

tritemio:
Add remaining commit.


newville:
Thanks!  This time for sure!

Issue #281, opened 2015-12-05
AttributeError: 'Call' object has no attribute 'starargs'


flyhunter:
I'm using expressions for parameters and I'm getting an error. My code is:

```python
x = [1723, 1773, 1823, 1523, 1773, 1033.03078, 1042.98077, 1047.90937, 1053.95899, 1057.94906, 1063.13788, 1075.74218, 1086.03102]
y = [0.79934, -0.31876, -0.46852, 0.05, -0.21, 11.1708, 10.31844, 9.73069, 9.21319, 9.12457, 9.05243, 8.66407, 8.29664]

def VFT(T, ninf=-3, A=5e3, T0=800):
 return ninf + A/(T-T0)

vftModel = Model(VFT)
vftModel.set_param_hint('D', vary=False, expr=r'A*log(10)/T0')
result = vftModel.fit(y, T=x)
print(result.fit_report())
```

The error reads:

     - Adding parameter for hint "D"
     - Adding parameter "ninf"
     - Adding parameter "A"
     - Adding parameter "T0"
     - Adding parameter "D"
    AttributeError
       <_ast.Module object at 0x7efded66e6d8>
          ^^^
    'Call' object has no attribute 'starargs'

    AttributeError:  in expr='<_ast.Module object at 0x7efded66e6d8>'

However, everything runs smoothly when I run the above code without using "log" in the expression:

```python
vftModel.set_param_hint('D', vary=False, expr=r'A/T0')
```

I'm using Python3.5 in a fully updated Arch Linux (as of 5 Dec 2015) with the latest lmfit (0.9.2).

newville:
@flyhunter  thanks!  This looks like a good hint:

https://github.com/alecthomas/importmagic/commit/4f26647461165a6ffa163f4d745bf52d71bb1027

newville:
Should be addressed with #282

PR #282, opened 2015-12-06
Py35 updates

newville:
This addresses #281, fixing use of function calls in constraint expressions for Py3.5
Also includes a couple small code cleanups.

### DIFF
--- a/lmfit/asteval.py
+++ b/lmfit/asteval.py
@@ -622,7 +622,7 @@ class Interpreter:
 
     def on_call(self, node):
         "function execution"
-        #  ('func', 'args', 'keywords', and 'starargs', 'kwargs' in py < 3.5)
+        #  ('func', 'args', 'keywords'. Py<3.5 has 'starargs' and 'kwargs' too)
         func = self.run(node.func)
         if not hasattr(func, '__call__') and not isinstance(func, type):
             msg = "'%s' is not callable!!" % (func)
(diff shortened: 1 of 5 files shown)


lmfit/asteval.py
```diff
@@ -622,25 +622,27 @@ def on_raise(self, node):    # ('type', 'inst', 'tback')
 
     def on_call(self, node):
         "function execution"
-        #  ('func', 'args', 'keywords', 'starargs', 'kwargs')
+        #  ('func', 'args', 'keywords', and 'starargs', 'kwargs' in py < 3.5)
```
Tillsten: left ',' before and

Tillsten: else it looks all good as far as i can tell.

newville:
I'll merge this in about a day unless there are objections.

Also: I'm not sure why Travis CI didn't like Python 3.5 (no dateutil module?).   Tests do pass for Python3.5 for me.  Hopefully that can be figured out soon.

tacaswell:
What does lmfit use dateutil for?  

It looks like the meta-data on pypi won't even let you install it via pip.  I guess you could do a source install...

newville:
@tacaswell  Thanks -- good point.   dateutil is not used.  It's probably just cruft from some earlier travis recipe.  I'll try without it.

newville:
@tacaswell  that worked ... thanks!

PR #286, opened 2015-12-22
ENH: speeding up update_constraints

andyfaff:
Some speed improvements in `Parameters.update_constraints`

### DIFF
--- a/lmfit/parameter.py
+++ b/lmfit/parameter.py
@@ -141,7 +141,9 @@ class Parameters(OrderedDict):
         Update all constrained parameters, checking that dependencies are
         evaluated as needed.
         """
-        _updated = [name for name,par in self.items() if par._expr is None]
+        requires_update = {name for name, par in self.items()
+                           if par._expr is not None}
+        updated_tracker = set(requires_update)
 
         def _update_param(name):
             """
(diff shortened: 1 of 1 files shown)


newville:
@andyfaff  Thanks.  This seems like it's passing all the tests, so should be OK to merge.   But I have a couple of questions: 

* are you sure the checks in lines 159 to 161 are unnecessary?   I can believe they are, but there have been subtle edge-cases in the past and some of what looks like cruft is actually needed.   And it's hard to see this as dominating performance.

* how much of a speed-up is this?

andyfaff:
I don't think the checks are needed because it should be all taken care of
in `Parameter._getval`, which is called when retrieving the value
attribute.
I've put the script I used to profile at:
https://gist.github.com/andyfaff/a0158e26e663b9308899

The changes in this PR speed the `update_constraints` method up by 46%
(from 0.0463 s to 0.0249 s). This is out of a total of 0.19 s (unmodified).
So it's about a 10% speedup.
These changes become more significant if `Parameter._getval` is modified.

newville:
@andyfaff   Sorry for the delay on this and for not paying very close attention to these speed problems.  The whole of the 'Get Value and Update Constraints' method is kind of a mess.   So, thanks very much for taking this on.

My inclination is to merge this and #287, but I'm slightly concerned about unintended and untested consequences, especially with respect to using lmfit.Parameters in a non-fitting context?    That is, I've only done a little bit of playing with this branch.   Also, should we include some timing benchmark scripts to avoid regressions?    I don't think that is required for these PRs, but might be useful.

OK to merge?

andyfaff:
The scripts I used to profile these functions were ipython notebooks. I'll
try and find them again. I'm not sure what would be required to automate.
I think #287 is worth taking a closer look at before it's merged.




On 17 January 2016 at 04:00, Matt Newville <notifications@github.com> wrote:

> Merged #286 <https://github.com/lmfit/lmfit-py/pull/286>.
>
> —
> <https://github.com/lmfit/lmfit-py/pull/286#event-517000141>.
>



-- 
Dr. Andrew Nelson


newville:
@andyfaff  These both (#286 and #287) look OK to me, though I can believe further tweaks might be helpful.  I decided to merge #287 too, mostly to avoid getting too many separate branches going and because we're definitely going to want these improvements.

PR #287, opened 2015-12-23
ENH: Streamline Parameter._getval

andyfaff:
These changes speed up the `Parameter.value` attribute retrieval by a significant amount.

The following gists 
(unstreamlined _getval)
https://gist.github.com/andyfaff/a0158e26e663b9308899
and
(streamlined _getval)
https://gist.github.com/andyfaff/a3f4fc311a2c97c630fa

outline that the overhead from _getval is reasonably large. These modifications halve the time required to evaluate the `userfcn` in this example. A total speedup of ~31% should be achievable in Minimizer.__residual.

### DIFF
--- a/lmfit/parameter.py
+++ b/lmfit/parameter.py
@@ -335,9 +335,9 @@ class Parameter(object):
     vary : bool
         Whether the Parameter is fixed during a fit.
     min : float
-        Lower bound for value (None = no lower bound).
+        Lower bound for value (None or -inf means no lower bound).
     max : float
-        Upper bound for value (None = no upper bound).
+        Upper bound for value (None or inf means no upper bound).
     expr : str
         An expression specifying constraints for the parameter.
     stderr : float
@@ -347,7 +347,7 @@ class Parameter(object):
         Of the form `{'decay': 0.404, 'phase': -0.020, 'frequency': 0.102}`
     """
     def __init__(self, name=None, value=None, vary=True,
-                 min=None, max=None, expr=None):
+                 min=-inf, max=inf, expr=None):
         """
         Parameters
         ----------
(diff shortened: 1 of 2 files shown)


andyfaff:
I've since confirmed that a combination of #286 and #287 speedup `Minimizer.__residual` by 37%, for the example script given above. 

(master branch was 0.326 s, both PR's were 0.206 s)

newville:
@andyfaff thanks for this.  I've been off-line for awhile and will be again most of the next week.  I doubt I'll have time to look at this in detail for several weeks, and it seems like it might need a more careful review than I get git it right now.

The `_getval()` and related code is certainly messy, and I think grew several barnacles in trying to make Parameter useful outside of the context of the `Minimizer` class.   It could probably do with a more thorough review / redesign.   

I don't really understand why `min` and `max` need to be properties. That is, accessing or setting these shouldn't need to trigger a calculation, unless I'm missing something subtle.  They should just be attributed, I think, with default of +/-inf and a value of `None` being replaced appropriately as soon as possible (say, in `__init__` or `setup_bounds`).

andyfaff:
No worries about any time delay, it's better to take time and get things right.

The general idea behind these changes is that having too many `if` statements can add up and degrade performance (as per the investigations outlined above).

This PR streamlines `_getval` by removing as many of those checks as possible. Currently several of the checks examine `self.max` and `self.min`, making sure that the limits make sense (e.g. if `self.min` is `None`, then make the lower limit `-np.inf`).  In the amended version accessing the limits does not require any calculation, apart from returning the value. However, by using a property setter (e.g. `set_min`) then you can ensure that the limits make sense.
The property setters ensure that `min < max` and that `self.min != self.max`, and if you try to set a limit of `None`, then either `-np.inf` or `np.inf` is used as the limit.  If you just have `min` and `max` as attributes then you can't make those checks.
For example the user could take an already initialised `Parameter` and decide to remove a limit:

    >>> a = Parameter('a', 10, True, -10, 10)
    >>> # now decide to remove lower limit without going through Parameter.set
    a.min = None

With the existing code you would then get:

    >>> print(a.min)
    None

Whereas with the new code you would get:

    >>> print(a.min)
    -np.inf

PR #290, opened 2016-01-19
PEP8: various PEP8 and style fixes

andyfaff:
various PEP8 changes. (Pycharm suggests that they're PEP8 problems). @newville, if this isn't important then feel free to close PR.


newville:
@andyfaff Thanks!  Will merge.  FWIW, I'm OK with preferring PEP8 (even if for no other reason than having a consistent formatting model).   I don't always remember to hold to it myself, probably reflecting the slow death of old habits.

PR #294, opened 2016-01-28
MAINT: improve Parameters pickling support

andyfaff:
Peter Metz reports a problem with unpickling a `Parameters` instance. The problem can be reproduced with:

    import lmfit
    import pickle
    p = lmfit.Parameters()
    p.add('var_0', value=1)
    p.add('var_1', value=2)
    p['var_0'].expr = 'var_1'
    pkl = pickle.dumps(p)
    q = pickle.loads(pkl)

The Error arises because `var_0` has an expr that depends on `var_1`. When `p` is unpickled the unpickling process tries to retrive the value of `var_1` from the symtable, but it doesn't yet been added to the symtable.
Solution: add the symbols to the symtable first, then add the parameters.

### DIFF
--- a/lmfit/parameter.py
+++ b/lmfit/parameter.py
@@ -162,17 +162,24 @@ class Parameters(OrderedDict):
 
         Parameters
         ----------
-        state : list
-            state[0] is a dictionary containing symbols that need to be
-            injected into _asteval.symtable
-            state[1:] are the Parameter instances to be added
-         is list of parameters
+        state : dict
+            state['unique_symbols'] is a dictionary containing symbols that
+            need to be injected into _asteval.symtable
+            state['params'] is a list of Parameter instances to be added
         """
-        # first add all the parameters
+        # first update the Interpreter symbol table. This needs to be done
+        # first because Parameter's early in the list may depend on later
+        # Parameter's. This leads to problems because add_many eventually leads
+        # to a Parameter value being retrieved with _getval, which, if the
+        # dependent value hasn't already been added to the symtable, leads to
+        # an Error. Another way of doing this would be to remove all the expr
+        # from the Parameter instances before they get added, then to restore
+        # them.
+        self._asteval.symtable.update(state['unique_symbols'])
+
+        # then add all the parameters
         self.add_many(*state['params'])
 
-        # now update the Interpreter symbol table
-        self._asteval.symtable.update(state['unique_symbols'])
 
     def update_constraints(self):
         """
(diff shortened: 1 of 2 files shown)


newville:
@andyfaff  OK, thanks!  This is a 1 line change, plus an added test, and all tests pass.  I think I'm supposed to be very grateful and merge this, and I think I would for most cases.    I'm willing to merge, but this conversation and issue seems wrong to me.  I'm willing to be out-voted on this, so please chime in, but here is my view:

Do we actually care if Parameters can be pickled?  Because, I'm sure we could find more problems with pickling if we looked for like 10 more minutes. 

Now, we'd probably all like for multi-processing fitting to magically "work" so that fitting could make use of multiple cores.   Unfortunately, the python multiprocessing library uses pickling, so it seems natural to be looking at any problem with pickling objects as standing in the way of our being able to use multiprocessing. 

Except, that pickle is simply broken.   The number of things that cannot be pickled is huge and important, and if multiprocessing *requires* pickle, multiprocessing involving any complex object is broken, more or less by design. So we could try to find all the problems that might arise in pickling Parameters or Models or whatever.  Or we could use another approach.  Perhaps dill will help.  Then again, pickling parameters is only part of the problem.

Here's my suggestion.  First, we accept and merge this PR.   Next, we make a 'fix_multiprocessing' branch, and the next time someone finds a flaw with pickling any lmfit object or wants to work on fitting using multiprocessing, they can work on that branch all they want.  When they get fitting with multiprocessing to work, we'll happily merge that into master.   In the meantime, we take no more changes for problems with pickling or multiprocessing on the main branch.

Sorry if this seems less than grateful to @andyfaff  -- I am not ungrateful, really!  I just think the approach is kind of a waste of effort.  Comments welcome.

andyfaff:
I think this bug is more of an oversight on my part, and is less to do with
any deficiencies with pickling in python. Bugs will inevitably be present
in any codebase and this PR is a reflection of that.

I'm more than happy to answer questions related to pickling and unpickling
from the user community, and to maintain the code supporting
pickling/unpickling of Parameter(s).

There will be Parameters that can't be pickled, such as those containing
lambda functions for expressions. But I think the (un)pickling code is
actually pretty neat, aside from the bug we've just knocked on the head.

newville:
@andyfaff   I merged this.  I think it's fine to view this a bug, but it's going to be easy to find Parameters instances that cannot be pickled.  I'm reluctant to say that we really want to support the idea that Parameters can be pickled.   In general, they cannot pickled, though it sometimes works.

OTOH, one could certainly save Parameters via a json representation (or similar) and restore them in another Python session or send them via some socket-based protocol to another running process.    I think we can promise that such a save()/load() mechanism works reliably and any cases that don't work are fixable and worth fixing.  I don't see that we can really say that with Pickle.

Issue #296, opened 2016-02-04
Unexpected behavior of Parameters.copy() -- re-ordering or arguments in Parameters.__init__()


tronsgaard:
I am relatively new to lmfit and Python in general, but I think I might have stumbled upon a small error in parameters.py:

These few lines of code triggers a mysterious AttributeError in lmfit-0.9.2, Python 3.4.3:
```
from lmfit import Parameters
p1 = Parameters()
p2 = p1.copy()
p2.add('foo')
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
/usr/local/lib/python3.4/dist-packages/lmfit/parameter.py in _getval(self)
    546         try:
--> 547             self.value = max(self.min, min(self._val, self.max))
    548         except(TypeError, ValueError):

TypeError: unorderable types: float() < NoneType()

During handling of the above exception, another exception occurred:

AttributeError                            Traceback (most recent call last)
<ipython-input-5-b0051757b071> in <module>()
----> 1 p2.add('foo')

/usr/local/lib/python3.4/dist-packages/lmfit/parameter.py in add(self, name, value, vary, min, max, expr)
    198         else:
    199             self.__setitem__(name, Parameter(value=value, name=name, vary=vary,
--> 200                                              min=min, max=max, expr=expr))
    201 
    202     def add_many(self, *parlist):

/usr/local/lib/python3.4/dist-packages/lmfit/parameter.py in __setitem__(self, key, par)
     83         par.name = key
     84         par._expr_eval = self._asteval
---> 85         self._asteval.symtable[key] = par.value
     86 
     87     def __add__(self, other):

/usr/local/lib/python3.4/dist-packages/lmfit/parameter.py in value(self)
    557     def value(self):
    558         "The numerical value of the Parameter, with bounds applied"
--> 559         return self._getval()
    560 
    561     @value.setter

/usr/local/lib/python3.4/dist-packages/lmfit/parameter.py in _getval(self)
    547             self.value = max(self.min, min(self._val, self.max))
    548         except(TypeError, ValueError):
--> 549             self.value = nan
    550         return self._val
    551 

/usr/local/lib/python3.4/dist-packages/lmfit/parameter.py in value(self, val)
    565         if not hasattr(self, '_expr_eval'):  self._expr_eval = None
    566         if self._expr_eval is not None:
--> 567             self._expr_eval.symtable[self.name] = val
    568 
    569     @property

[...]

PR #297, opened 2016-02-04
Change the order of arguments in the signature of Parameters.__init__()

tronsgaard:
**Abstract:** This pull request switches the order of the arguments in `Parameters.__init__()`, ensuring that `asteval=None` is interpreted as a keyword argument.

**Reason:** The following code results in an AttributeError:
```
from lmfit import Parameters
p1 = Parameters()
p2 = p1.copy()
p2.add('foo')
```
```
/usr/local/lib/python3.4/dist-packages/lmfit/parameter.py in _getval(self)
    546         try:
--> 547             self.value = max(self.min, min(self._val, self.max))
    548         except(TypeError, ValueError):

TypeError: unorderable types: float() < NoneType()

During handling of the above exception, another exception occurred:

AttributeError                            Traceback (most recent call last)
<ipython-input-5-b0051757b071> in <module>()
----> 1 p2.add('foo')

/usr/local/lib/python3.4/dist-packages/lmfit/parameter.py in add(self, name, value, vary, min, max, expr)
    198         else:
    199             self.__setitem__(name, Parameter(value=value, name=name, vary=vary,
--> 200                                              min=min, max=max, expr=expr))
    201 
    202     def add_many(self, *parlist):

/usr/local/lib/python3.4/dist-packages/lmfit/parameter.py in __setitem__(self, key, par)
     83         par.name = key
     84         par._expr_eval = self._asteval
---> 85         self._asteval.symtable[key] = par.value
     86 
     87     def __add__(self, other):

/usr/local/lib/python3.4/dist-packages/lmfit/parameter.py in value(self)
    557     def value(self):
    558         "The numerical value of the Parameter, with bounds applied"
--> 559         return self._getval()
    560 
    561     @value.setter

/usr/local/lib/python3.4/dist-packages/lmfit/parameter.py in _getval(self)
    547             self.value = max(self.min, min(self._val, self.max))
    548         except(TypeError, ValueError):
--> 549             self.value = nan
    550         return self._val
    551 

/usr/local/lib/python3.4/dist-packages/lmfit/parameter.py in value(self, val)
    565         if not hasattr(self, '_expr_eval'):  self._expr_eval = None
    566         if self._expr_eval is not None:
--> 567             self._expr_eval.symtable[self.name] = val
    568 
    569     @property

AttributeError: 'Parameters' object has no attribute 'symtable'
```

Interestingly, the old parameters object has ended up in `p2._asteval`:
```
p2._asteval.add('foo')
print(p2._asteval)
    Parameters([('foo', <Parameter 'foo', -inf, bounds=[-inf:inf]>)])
```
The error seems to be caused by the definition of `Parameters.__init__()`:
```
    def __init__(self, asteval=None, *args, **kwds):
        super(Parameters, self).__init__(self)
        self._asteval = asteval

        if asteval is None:
            self._asteval = Interpreter()
        self.update(*args, **kwds)
```
The keyword argument `asteval=None` is being interpreted as a positional argument, so that when an existing Parameters object is passed to the constructor as the first argument, it is seen as the `asteval` parameter. This is indeed what the .copy() method does.
```
[/usr/lib/python3.4/collections/__init__.py:211]

    def copy(self):
        'od.copy() -> a shallow copy of od'
        return self.__class__(self)
```

### DIFF
--- a/lmfit/parameter.py
+++ b/lmfit/parameter.py
@@ -73,7 +73,7 @@ class Parameters(OrderedDict):
     dumps() / dump()
     loads() / load()
     """
-    def __init__(self, asteval=None, *args, **kwds):
+    def __init__(self, *args, asteval=None, **kwds):
         super(Parameters, self).__init__(self)
         self._asteval = asteval
 


tronsgaard:
Not sure why this patch works in Python3 and fails in Python2 though..

newville:
@tronsgaard I reopened #296. The issue exists until the solution is found, not when a PR is created.

I agree that copying Parameters seems broken.   In fairness, though, your example use of 'Parameters.copy()' is pretty silly.   I won't say that copying Parameters is not needed, but can you give a more realistic reason why this should do something (and what is that something)?    Anyway, my guess is that a better solution is probably to add an explicit `copy()` method:

```python
    def copy(self):
        return self.__deepcopy___(None)
```

but I have not tested that thoroughly.

tronsgaard:
@newville Sorry about the confusion - should the discussion take place in the issue then?

And yes, my minimalistic example is pretty silly - it was just meant to illustrate the behavior.

I stumbled upon `.copy()` when I was wrapping an lmfit model and corresponding parameters inside a custom class in a piece of code that does repeated fitting of almost identical data. Since I was using version 0.8.x, I wanted to create a (deep) copy of the pre-defined initial parameters before passing them to the model. Shortly after, I upgraded to lmfit-0.9.x and got the error message above - and after looking closer, I realized that .copy() was in fact only doing a shallow copy.

With lmfit>=0.9 I probably don't need to copy parameters at all. In case I need compatibility with lmfit<0.9, I could use `copy.deepcopy()` or perhaps a direct call to `.__deepcopy__()`.

However, as a user, I would expect the methods that are visible within a class to be working (e.g. when using Tab-completion in IPython). If not fixing the `.__init__()` somehow, I suggest overwriting `.copy()` in Parameters to raise a NotImplementedError and also hide it from the user.

From a compatibility point-of-view, there might also be users relying on the .copy() method returning a shallow copy as in previous versions.

newville:
@tronsgaard  if the discussion is about the issue, the discussion should be with the issue.  If the discussion is about the proposed PR, it should be with the proposed PR.  If it's a general question and/or you are not sure, it should be on the mailing list.   But, certainly having the conversation in two places is a bad idea.    Is that too obvious?  

My preference would be for you to close both of these and start over.

tronsgaard:
Okay, let's start over then

PR #303, opened 2016-02-15
explicitly set Parameter.min/max to -/+ inf, avoid properties and extra function calls

newville:
This more clearly makes Parameter.min/max defaults to be `-/+inf`, and more quickly and assertively changes `None` to `inf`.    This allows avoiding properties for parameter bounds.  It makes a small but noticeable improvement in speed and also (I think) cleans up the code by moving many checks for min/max of `None`. 

With the example benchmark code in `example/benchmark_fit.py` (very similar to gist posted by @andyfaff), I get a runtime of 0.878 seconds with this code, compared to 0.937 seconds with the master branch, and 0.962 seconds with ba6b3e4be7b34f01d2f5badf21b2d1f5703ac14c  (just before speedups to `Parameter._getval()` from @andyfaff).   That's with Anaconda Python 2.7.10 on a macbook pro.
  
It seems handy to have the benchmark example to check for any possible regressions in runtime.


examples/benchmark_fit.py
```diff
@@ -0,0 +1,52 @@
+import numpy as np
+import time
+import cProfile, pstats
```
andyfaff: I don't think that cProfile exists on Python3.

lmfit/parameter.py
```diff
@@ -458,52 +458,34 @@ def set(self, value=None, vary=None, min=None, max=None, expr=None):
...
         """make sure initial bounds are self-consistent"""
         # _val is None means - infinity.
+        if self.max is None:
+            self.max = inf
+        if self.min is None:
+            self.min = -inf
         if self._val is not None:
-            if self.max is not None and self._val > self.max:
+            if self.min > self.max:
+                self._min, self._max = self.max, self.min
```
andyfaff: shouldn't this be:

    self.min, self.max = self.max, self.min

lmfit/parameter.py
```diff
@@ -458,52 +458,34 @@ def set(self, value=None, vary=None, min=None, max=None, expr=None):
...
+            if isclose(self.min, self.max, atol=1e-13, rtol=1e-13):
+                raise ValueError("Parameter '%s' has min == max" % self.name)
+
+            if self._val > self.max:
```
andyfaff: L481-484, remove two `if` clauses

    self._val = max(self.min, min(self.max, self._val))

lmfit/parameter.py
```diff
@@ -458,52 +458,34 @@ def set(self, value=None, vary=None, min=None, max=None, expr=None):
...
         # _val is None means - infinity.
+        if self.max is None:
+            self.max = inf
+        if self.min is None:
+            self.min = -inf
         if self._val is not None:
-            if self.max is not None and self._val > self.max:
+            if self.min > self.max:
+                self._min, self._max = self.max, self.min
+            if isclose(self.min, self.max, atol=1e-13, rtol=1e-13):
```
andyfaff: This `isclose` check needs to happen whenever `self.min` or `self.max` is changed. The only way to ensure that is to use properties! I suspect the majority of changing bounds does not go through the `set` method, they are changed directly, i.e. `par.min = 5`.
Otherwise the `self.from_internal` and `self.scale_gradient` can have divide by zero's, e.g. L571.

andyfaff:
Having `self.min` and `self.max` as properties has a lot of advantages:

1. One can ensure that max and min are always set to valid values.
2. By having properties there is an explicit approach to setting each value. The setting only occurs in one place. By removing the properties it's possible to change the attributes anywhere in the code, I think some robustness is lost.
2. The `isclose` test can be done whenever `min` and `max` are changed.

I don't think there's a large speedup by removing them, although there are extra lines.

lmfit/parameter.py
```diff
@@ -458,52 +458,34 @@ def set(self, value=None, vary=None, min=None, max=None, expr=None):
...
-        if max is not None:
-            self.max = max
+        if min is None:
+            min = -inf
+        if max is None:
+            max = inf
+        self.min = min
+        self.max = max
 
     def _init_bounds(self):
```
andyfaff: `_init_bounds` is only called once, when the object is constructed. These valuable checks should be carried out whenever `min` or `max` are changed.

lmfit/parameter.py
```diff
@@ -458,52 +458,34 @@ def set(self, value=None, vary=None, min=None, max=None, expr=None):
...
-        self._max = val
-        if self.min > self.max:
-            self._min, self._max = self.max, self.min
-        if isclose(self.min, self.max, atol=1e-13, rtol=1e-13):
-            raise ValueError("Parameter '%s' has min == max" % self.name)
-
-    def get_min(self):
-        return self._min
-
-    def set_min(self, val):
```
andyfaff: The checks in `set_min` should be carried out whenever `min` is changed. How will this be done if the properties are removed?

lmfit/parameter.py
```diff
@@ -458,52 +458,34 @@ def set(self, value=None, vary=None, min=None, max=None, expr=None):
...
+        elif self._expr is None:
             self._val = self.min
-        elif self.max is not None and self._expr is None:
-            self._val = self.max
         self.setup_bounds()
 
-    def get_max(self):
-        return self._max
-
-    def set_max(self, val):
```
andyfaff: The checks in `set_max` should be carried out whenever `max` is changed. How will this be done if the properties are removed?

newville:
@andyfaff   Yes, there's a typo at line 477.  I do believe cProfile is included in Python 3, though I could be wrong.   For using min/max as propertied, I could be persuaded either way.  I find that the speedup in removing properties for min/max are approximately the same as those you made when streamlining `_getval()` a couple of months ago  (be0076f1fa7183d21c30d86c376016aa651099ce).    While I would normally agree that changes of runtime smaller than 10% are not very important, you did for #286 and #287, although I think you said you found a larger speedup for those 2 changes was ~30%, while I see ~10% or less.  Not sure why that is.

Using properties for min/max is definitely more code.  Checking that the bounds are well-ordered and not equal, doesn't seem like it solves a very common problem to me.  

But also, setting a `min` to the same value as `max` **does** result in the `ValueError` being raised when using the Parameters instance in a fit.   This is because, when used by `Minimizer()`, a copy (deepcopy) is made just prior to the fit so that  `_init_bounds()` is in fact run:

```python 
    params = Parameters()
    params.add('amp',   value= 1.0, min=0.0, max=10.0)
    params['amp'].min=1.0
    params['amp'].max=1.0
    params.add('decay', value= 0.0, min=-1.0, max=10.0)
    params.add('shift', value= 0.0, min=-np.pi/2., max=np.pi/2)
    params.add('omega', value= 1.0, min=-10.0, max=10.0)
    out = minimize(fcn2min, params, args=(x, data)) 
....

  File "/Users/Newville/Codes/lmfit-py/lmfit/parameter.py", line 477, in _init_bounds
    raise ValueError("Parameter '%s' has min == max" % self.name)
ValueError: Parameter 'amp' has min == max
```

The error is raised at `minimize()`, not when the value is set.  I think that's OK.  This check won't happen when using Parameters outside the context of a fit.  Then again, we never promised that was supported.  

It's kind of hard to argue with code that passes all tests with fewer lines of code and a slightly faster runtime.  OTOH, if there are tests that *should* be added that make the code fail, then slower-but-correct is clearly better.     Are there tests that should be added?

newville:
@andyfaff any thoughts on this?   I would be willing to abandon the PR if you still object. I do think it has a marginal improvement in speed.   Either way, it might be useful to include the benchmarking script.

andyfaff:
I think a PR that sets up benchmarking for the project would be very helpful. `asv` is commonly used for such purposes. 
What I was particularly concerned with a while back was making sure that whenever the optimizer requests evaluations, then that code is executed as quickly as possible. For example, with `Minimizer.leastsq` the `scipy.leastsq` code calls `Minimizer.__residual`. Whatever is run from `Minimizer.__residual` should run as fast as possible. In `lmfit.emcee` it wouldn't be unusual for the user function to be evaluated on the order of 1e6 times or more, whereas `leastsq` would only require on the order of a couple of hundred function evaluations. I figured out from line profiling that (amongst others) `Parameter._getval` could be streamlined. I was less concerned about the cost of setup and teardown of a minimization, as that's negligible. I did look into the cost of properties vs cheap attributes - but I didn't consider it an issue.
I've just [done an `emcee` run as a test](https://gist.github.com/andyfaff/c75fa35b16a15d865a09) (using examples/lmfit_emcee_example.ipynb) and the total run time (TRT) was 440 s. Of that:

- 250 s (57% TRT) was spend in `_getval`. Less than half of TRT was spent on objective function calculation!
- 154 s (35% TRT) was spent on line [L615](https://github.com/lmfit/lmfit-py/blob/master/lmfit/parameter.py#L615), which is where the parameter value is clipped. 
- 19 s (5% TRT) is spent retrieving the min/max values on L615, the overwhelming majority of that line is using the `min` and `max` functions. Note, I don't know how much time would be used retrieving a bog standard attribute.
- 30 s (7% TRT) is spent on [L596](https://github.com/lmfit/lmfit-py/blob/master/lmfit/parameter.py#L596), testing if the value is an `uncertainties` variable.

So removing these properties would give a maximum of 5% speedup. In comparison figuring out how to clip a value to min/max in an efficient way might give 35% speedup!
On balance I'd probably leave the properties in.

newville:
@andyfaff  I agree about using `asv` -- it looks like exactly the right tool.  I'd have to look into how to set that up -- have you used this package much before?  Could you set this up for lmfit?

I'm OK with including a long emcee-using run in a canonical benchmark, but that should probably also include simpler uses, as with `leastsq`.  I would have thought that for any objective function,  N_residual / (N_getval * N_variables)  would be fairly constant (for N_residual = # of calls to _residual(), N_getval = # of calls to Parameter._getval(), and N_variables = # of variables).   

That said, I would not be surprised if run-time was dominated by getting values for bounded parameters (as compared to un-bounded values). It might be nice to measure that accurately.

For speeding up `_getval()`,  I'd suggest comparing

```python
self.value = max(self.min, min(self._val, self.max))
```

with

```python
v = self._val
if v > self.max: v = self.max
if v < self.min: v = self.min
self.value = v
```

In my more crudely measured benchmarks I also see a small (~5%) speedup this change.  

I understand calling that a small effect, but I also think that taking a 5% improvement when possible is not a bad idea.   Have you compared this PR with master using asv?

newville:
@andyfaff  I made this change (away from `min(maxval, max(minval, val))`), and also added asv benchmarking tools, and two benchmarks to compare different commits, including this branch with master.    I put the benchmarking files in `asv_benchmarking`.   More tests could certainly be added.

An example output at http://millenia.cars.aps.anl.gov/software/python/lmfit/asv/run1/ , though we can include such results in the main github.io doc.   

Basically, for a simple fit see a 5% to 10% improvement in a simple fit for the changes in this PR, with eliminating the calls to `min()` and `max()` helping a little bit.  I believe any improvement here will likely be noticeable as the number of variables and function evaluations goes up.

The results using `conf_interval()` fit is less clearly a win, and the results are noisier. I'm not sure why this is, but my guess is it spends less of its time doing `Parameter._getval()`.

I ran into odd errors with `scalar_minimize()` and the CG method, though only with Python3.  I removed this method from some of the tests, though maybe that should be investigated further.

I didn't add a diffev example to the benchmark.   Do you have a suggestion for one to include?

asv_benchmarking/README.md
```diff
@@ -0,0 +1 @@
+Benchmarking with asv (air-speed-velocity)
```
andyfaff: I think it needs some remarks on how you start it going, and how you create the webpages, etc.

asv_benchmarking/asv.conf.json
```diff
@@ -0,0 +1,73 @@
...
+    "project_url": "http://lmfit.github.io/lmfit-py",
+
+    // The URL or local path of the source code repository for the
+    // project being benchmarked
+    "repo": "http://github.com/lmfit/lmfit-py/",
+
+    // List of branches to benchmark. If not provided, defaults to "master"
+    // (for git) or "tip" (for mercurial).
+    "branches": ["master", "param_minmax_inf"], // for git
+    // "branches": ["tip"],    // for mercurial
```
andyfaff: I don't think we'd want the `param_minmax_inf` branch hardcoded in there. Can this just be the master branch, and whatever branch is currently being worked on?

asv_benchmarking/asv.conf.json
```diff
@@ -0,0 +1,73 @@
+{
```
andyfaff: Instead of the `/asv_benchmarking` dir could this be in `/benchmark`

andyfaff:
I like the addition of `asv` to the project (see line notes above). Although it's more work I think the `asv` changes would be better off in a separate PR, because the asv work is pretty much orthogonal and is worth adding straightaway.
It's nice to look at the graphs and see the asv times drop around mid-October when the astspeed up PR's went in.
I'll create a few benchmarks for `differential_evolution` and `emcee` rightaway. We can use those to compare master and this branch. If this branch is noticeably faster then I'd be +1 for this PR to be merged. Whilst I create the benchmarks can you decide whether you'll create a separate `asv` PR, or you'll keep it in here?

newville:
@andyfaff  I'm very new to asv, and had its docs open the whole time I was working on this.  Improving the README.md to give a quick description of how to use asv is a great idea.

I agree we probably don't want any branch except `master` hard-coded in, but it is sometimes going to be useful to compare branches.   FWIW, I went with 'asv_benchmarking' (as opposed to 'benchmark') to make it clear it was asv-specific.   It seems to have many dedicated files and directories, so it probably can't coincide very well with other tools.  

In any event, I think all the evidence is that these changes to Parameter._getval() are justified.

lmfit/parameter.py
```diff
@@ -611,10 +595,10 @@ def _getval(self):
                     self.value = self._expr_eval(self._expr_ast)
                     check_ast_errors(self._expr_eval)
 
-        try:
-            self.value = max(self.min, min(self._val, self.max))
-        except(TypeError, ValueError):
-            self.value = nan
+        v = self._val
+        if v > self.max: v = self.max
```
andyfaff: I'd change these lines to:

    if self.min < self._val < self.max:
        pass
    else:
        self._val = max(self.min, min(self._val, self.max))
    self.value = self._val

I like the range check form instead of two ifs. It's also faster.

newville: It's the use of `min()` and `max()` functions that are the problem, not how
the test is done.

On Sun, Feb 21, 2016 at 9:22 PM, Andrew Nelson <notifications@github.com>
wrote:

> In lmfit/parameter.py
> <https://github.com/lmfit/lmfit-py/pull/303#discussion_r53583085>:
>
> > @@ -611,10 +595,10 @@ def _getval(self):
> >                      self.value = self._expr_eval(self._expr_ast)
> >                      check_ast_errors(self._expr_eval)
> >
> > -        try:
> > -            self.value = max(self.min, min(self._val, self.max))
> > -        except(TypeError, ValueError):
> > -            self.value = nan
> > +        v = self._val
> > +        if v > self.max: v = self.max
>
> I'd change these lines to:
>
> if self.min < self._val < self.max:
>     pass
> else:
>     self._val = max(self.min, min(self._val, self.max))
> self.value = self._val
>
> I like the range check form instead of two ifs. It's also faster.
>
> —
> <https://github.com/lmfit/lmfit-py/pull/303/files#r53583085>.
>



-- 
--Matt Newville <newville at cars.uchicago.edu> 630-252-0431

newville:
merged to master. but changes may not be finalized,especially for asv tests.

andyfaff:
I think it's quite rare (during fitting) that the parameters are outside
the limits. So on this case it's going to be the speed of the `if` tests
that is the rate limiting step.
The range test `if not (min < val < max):` is faster than two separate if
tests.

newville:
@andyfaff  I'd say to benchmark it, or use the `dis` module.    

FWIW, I think there has to be two tests, and that
```
if not(min < val < max)
```
is equivalent to 
``` 
if not (min < val and val < max)
```
and that
``` 
if val < min:  
    val = min
elif val > max
    val = max
```
is going to be about the same bytecode for the tests, but never call external functions.

Issue #308, opened 2016-02-29
conf_interval fails when fit was done with bounded/linked parameters.


wakass:
## Submitting an Issue
#
# 0. If you have not submitted an issue to LMFIT before, you
#    *MUST* read
#    https://github.com/lmfit/lmfit-py/blob/master/.github/Contributing.md
#    Do *NOT* submit an Issue without reading this. Use the Mailing list
#    https://groups.google.com/group/lmfit-py for questions about LMFIT.
#
# 1. Include an example that shows the Issue or explain why you did not.
#    It is OK to use gist or give a link to examples or data.
#    It is OK to use or point to Jupyter notebooks.

The following script:
import numpy as np
from numpy import sqrt,pi,exp,arange
import matplotlib.pyplot as plt
import lmfit as lm

def gaussian(x, amp, cen, wid):
    "1-d gaussian: gaussian(x, amp, cen, wid)"
    return (amp/(sqrt(2*pi)*wid)) * exp(-(x-cen)**2 /(2*wid**2))

def n_gaussians(x,pars,n_peaks=1):
    total = 0.
    for i in np.arange(n_peaks)+1:
        a = pars['amp_%d'%i].value
        w = pars['wid_%d'%i].value
        c = pars['cen_%d'%i].value
        total = total + gaussian(x,a,c,w)
    return total

def resid(pars,x, data=None,eps=None,model=n_gaussians,**kwargs):    
    model = model(x,pars,**kwargs)

    if data is None:
        resid = model
    else:
        resid = model - data 
    return resid

y = np.load('linecut.dat.npy')*1e4
x = np.linspace(0,3.,y.shape[0])

n_peaks = 4

p= lm.Parameters()
p.add_many( ('amp_1',  4.0, True,0.5, 6.,None),
            ('cen_1', 0.7, True, 0, .85,  None),
            ('wid_1', 0.11875854684504183, False, None, None,  None),

            ('amp_2',  None, False, None, None,'amp_1'),
            ('cen_2', .9, True, .85, 1.4,  None),
            ('wid_2', None, False, None, None,  'wid_1'),

            ('amp_3',  4.0, True,0.5, 6.,None),
            ('cen_3', 1.1, True, .95, 1.5,  None),
            ('wid_3', 0.24352453256655024 , True, 0.1, .3,  None),

            ('amp_4',  None,False,None,None,'amp_3'),
            ('cen_4', 1.7, True, 1.6, 2.5,  None),
[...]

PR #310, opened 2016-02-29
allow para.value to be plain float or np.float64

newville:
This addresses #308, and an earlier messages on mailing list.

In `confidence_interval()`, a Parameter value (in `para.value`) may be either a plain `float` or a `numpy.float64`.  Plain `floats` do not have a `copy()` method.  

This PR simply takes `float` of the Parameter value for the starting value.


Tillsten:
looks good

PR #316, opened 2016-03-14
Rudimentary support for complex models

gitj:
This pull request allows models to return complex values. A simple, but reasonably sensible residual function is provided.

Currently all tests pass _except_ the test_data_alignment test which is relevant for pandas Series data. I'm not yet sure how to address this, but I will attempt to figure out a solution.

The present implementation is slightly cryptic to gain efficiency because it occurs at one of the most performance critical portions of the code: the evaluation of the residual function.

### DIFF
--- a/examples/complex_resonator_model.ipynb
+++ b/examples/complex_resonator_model.ipynb
@@ -1,0 +1,424 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "metadata": {
+    "collapsed": false
+   },
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "Populating the interactive namespace from numpy and matplotlib\n"
+     ]
+    }
+   ],
+   "source": [
+    "%pylab inline"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 2,
+   "metadata": {
+    "collapsed": true
+   },
+   "outputs": [],
+   "source": [
+    "np.random.seed(123) # make this notebook reproduceable. "
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 3,
+   "metadata": {
+    "collapsed": true
+   },
+   "outputs": [],
+   "source": [
+    "import lmfit"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Following Khalil et al. (http://arxiv.org/abs/1108.3117), we can model the forward transmission of a microwave resonator with total quality factor $Q$, coupling quality factor $Q_e$, and resonant frequency $f_0$ using:\n",
+    "\n",
+    "$$ S_{21}(f) = 1-\\frac{Q\\,Q_e^{-1}}{1+2jQ(f-f_0)/f_0} $$.\n",
+    "\n",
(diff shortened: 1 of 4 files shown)


gitj:
Latest commit gets all tests passing by using .ravel() before .view() to deal with pandas Series.

newville:
@gitj   Thanks -- this looks good.   A couple suggestions:

1.  I think "is this data complex?" would be better as    `np.any(np.iscomplex(data))`
2.  I think casting complex to float might be better as  `data.view(np.float)`.   That is, I think your `data_dtype` has to be `np.float` (ie, double precision floating point) no matter what the Complex data type actually was.

I think similar tests would be appropriate for weights, but I think your handling of "complex data + real weights" is the right approach.

Anyway, thanks!

gitj:
@newville Regarding your suggestions:
1. I am fine with np.any(np.iscomplex(data)), it is certainly more readable. However, the reason I chose the implementation I did is for efficiency:
x = np.arange(10000,dtype='complex')

`In [7]: %timeit (x.dtype.char in np.typecodes['Complex'])`
`10000000 loops, best of 3: 179 ns per loop`
versus
`In [8]: %timeit np.any(np.iscomplex(x))`
`100000 loops, best of 3: 18.8 µs per loop`

This might matter for fitting data with lots of points and where many residual evaluations are required. But maybe that's too esoteric a use case to worry about.

2. If the user passes in data with dtype = np.complex64, then data.view(np.float) will not be correct (it needs to be data.view(np.float32). The current implementation handles all complex types.
However, since the code actually looks at the dtype of `diff`, not data directly, then it will probably be casted to np.complex128 anyway since the model will probably return that. But potentially a user could define a model that returns complex64 too to save memory.

So overall I would agree with your suggestions if readability of the code is more important than absolute performance (for point 1) and generality (for point 2). I do think it's more correct to properly deal with dtypes other than np.complex128 and np.float64 in general.

Let me know if I've convinced you one way or another and I can update the pull request.

newville:
@gitj Hm, the timing is persuasive for the slightly less readable form.  I would suggest adding a comment on this '# np.iscomplex(data) too slow here !' 

For np.float, I believe the residual array needs to be float (aka float64) and cannot be float32 or other precision.  That is, I think this won't work with complex64 (but I could be wrong....) in which case explicitly casting to double seems preferable to "native".

gitj:
@newville 
That's a good point: the underlying scipy.optimize codes almost certainly need things to be np.float anyway. In that case things can be greatly simplified: The only usable complex type to check for is np.complex, and that can be directly viewed as np.float. So the only difference from your original suggestion is replacing np.any(np.iscomplex(data)) with a check of data.dtype is np.complex. For more generality, we could cast from other precisions to float64, complex128, but it is probably not worth it. I was worried about the .view trick not working because of mismatched precision, but if we only check for np.complex, it is a non issue. I'll update the pull request soon.

newville:
@gitj Great, I agree with using `data.dtype == np.complex` and then `view(np.float)`.

gitj:
OK, the new, simpler implementation is here and tests pass. I'd eventually like to add support for complex data in the plotting and other convenience functions, but I think that will have to be anohter pull request.

newville:
@gitj  Looks great.  Anyone have objections to merging?

gitj:
I've added a notebook demonstrating the use of this feature.

newville:
@gitj  Great, thanks for the example.  Unless I hear objections, I'll merge this in about a day.

Issue #317, opened 2016-03-16
Pretty printing of params_hints


tritemio:
The `Model` attribute `param_hints` is a nested `OrderedDict` which prints, by default, in a hard to read format:

```
OrderedDict([('p1_sigma',
              OrderedDict([('min', 0.01), ('value', 0.03), ('max', 0.2)])),
             ('p1_fwhm', OrderedDict([('expr', '2.3548200*p1_sigma')])),
             ('p2_sigma',
              OrderedDict([('min', 0.01), ('value', 0.03), ('max', 0.2)])),
             ('p2_fwhm', OrderedDict([('expr', '2.3548200*p2_sigma')])),
             ('p1_center',
              OrderedDict([('value', 0.3), ('max', 0.6), ('min', -0.1)])),
             ('p2_center',
              OrderedDict([('value', 0.85), ('max', 1.1), ('min', 0.5)])),
             ('p1_amplitude', OrderedDict([('value', 1), ('min', 0.01)])),
             ('p2_amplitude', OrderedDict([('value', 1), ('min', 0.01)])),
             ('br_amplitude', OrderedDict([('value', 0.0001), ('min', 0)])),
             ('br_center1', OrderedDict([('min', 0), ('expr', 'p1_center')])),
             ('br_center2', OrderedDict([('min', 0), ('expr', 'p2_center')])),
             ('br_sigma1', OrderedDict([('min', 0), ('expr', 'p1_sigma')])),
             ('br_sigma2', OrderedDict([('min', 0), ('expr', 'p2_sigma')]))])
```

A cleaner pretty printing would make inspection much faster, especially with many parameters. Compare the following output to the previous for example:

```
Name             Value      Min      Max    Expr    
p1_sigma          0.03     0.01      0.2    
p1_fwhm                                     2.3548200*p1_sigma
p2_sigma          0.03     0.01      0.2    
p2_fwhm                                     2.3548200*p2_sigma
p1_center          0.3     -0.1      0.6    
p2_center         0.85      0.5      1.1    
p1_amplitude         1     0.01             
p2_amplitude         1     0.01             
br_amplitude    0.0001        0             
br_center1                    0             p1_center
br_center2                    0             p2_center
br_sigma1                     0             p1_sigma
br_sigma2                     0             p2_sigma
```

The function to produce the above output is:

```python
def print_params_hints(p, spacing=8):
    name_len = max(len(s) for s in p)
[...]

PR #318, opened 2016-03-17
Add method to pretty-print Model.param_hints

tritemio:
Addresses issue #317.


lmfit/model.py
```diff
@@ -11,8 +11,6 @@
 from .printfuncs import fit_report, ci_report
 from .confidence import conf_interval
 
-from collections import MutableSet
-
 try:
```
tritemio: MutableSet was not used

newville:
@tritemio Very nice!

Issue #319, opened 2016-03-17
Undefined name 'long' in parameter.py


tritemio:
It may be a false positive, but my linter signals an error about using `long()` which is undefined on python3 in:

https://github.com/lmfit/lmfit-py/blob/0.9.3rc2/lmfit/parameter.py#L680

newville:
@tritemio Good point.  I wonder if there is something in `six` to handle this.... I didn't see it immediately.

tritemio:
Using the future package (which we already use) we could do:

```
# Python 2 and 3
from builtins import int
bigint = int(1)
```

http://python-future.org/compatible_idioms.html?highlight=long#long-integers

andyfaff:
six is vendored in scipy, so we know it's going to be installed.

tritemio:
Right, but future is already a dependence and results in more pythonic (i.e. python-3-like) code.

newville:
@tritemio using `builtins.int` looks like a good solution.  Are you willing to make a PR?

tritemio:
@newville I can make a PR. I don't know what is the `__long__` method used for? Apparently is never called because we never had a problem so far.

newville:
@tritemio Yeah, we could just remove this method.

newville:
completed

PR #321, opened 2016-03-18
Remove long() and use builtins.int() for py2/3 comp

tritemio:
Addresses #319

### DIFF
--- a/lmfit/parameter.py
+++ b/lmfit/parameter.py
@@ -2,6 +2,7 @@
 Parameter class
 """
 from __future__ import division
+from builtins import int
 from numpy import arcsin, cos, sin, sqrt, inf, nan, isfinite
 import json
 from copy import deepcopy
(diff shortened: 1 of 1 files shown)


newville:
@tritemio    Hmm, looks like `from builtins import int` is **not** valid Python2....  But if you're removing the `__long__()` method anyway, do we need this import?

tritemio:
It is valid python 2. It is our travis environment that does not install the package `future` which provides `builtins`.

I left the import because we call `int()` and with the `builtins` import we would use long integers on both python 2 and 3. Just consistency.

tritemio:
Ok, the problem is that `future` currently **is not** a dependency for lmfit. My mistake, I though it was.

Either with add `future` as a dependency (which is a very nice way to achieve python2/3 code) or we ignore the inconsistency (of having different ints in python 2 and 3) and remove the `from builtins import int`

newville:
@tritemio Great, thanks!

PR #322, opened 2016-03-18
Add future as dependency

tritemio:
This PR adds [`future`](http://python-future.org/) as a dependency. 

The `future` package provides a few importables: 
-  `builtins` (only on python2, which provides python3-style objects)
- `past`
- `future`

See:
- http://python-future.org/compatible_idioms.html

For the ints issue #319 (and #321) see:
- http://python-future.org/compatible_idioms.html#long-integers

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,7 @@ before_install:
     - conda info -a
 
 install:
-    - conda create -q -n test_env python=$TRAVIS_PYTHON_VERSION numpy scipy pandas matplotlib nose
+    - conda create -q -n test_env python=$TRAVIS_PYTHON_VERSION numpy scipy pandas matplotlib nose future
     - source activate test_env
     - pip install emcee
     - python setup.py install
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,3 @@
 numpy>=1.5
 scipy>=0.13
+future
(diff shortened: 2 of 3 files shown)


newville:
@tritemio   I think I'm -1 on this.  There actually is not a real problem with the existing code that this solves.  Why not just remove `__long__()` and move on? 

Like @andyfaff  suggests, the `six` package seems more prevalent at least in scipy world.   
Like, how many incompatible compatibility libraries can there be?

tritemio:
Yep, the problem at hand does not seem to have any implications. It is just different int types implicitly used in python 2 and 3. But it principle it may create issues depending on where these ints are used.

Six is more prevalent because is the older solution. Future came later but in most cases allows to write python3 code in python2. Six adds a shim instead. When dropping support for python 2 it will be mostly a matter of removing future.

Anyway, I wanted just to bring it to the table. But feel free to close it without merging.

newville:
@tritemio I'd say we don't need to add this dependency.   The vast majority of the code here is both Python2 and Python3.   We don't even rely on the `six` package.    (To be completely fair, asteval.py does test `sys.version_info` for differences in the `ast` module).    

So why not just remove `__long__()` (and `__int__()` for that matter)?   This will still allow
 
```python
>>> p = Parameter('x', 3.24)
>>> int(p)
3
>>> p = Parameter('x', 3.2e25)
>>> int(p)
31999999999999999463129088L   # python2
``` 
presumably because int() will use the  integer result of `float(p)`, which is defined....

The current PR #321 is failing, so can't be merged.  And it's failing on an import that is actually

Issue #323, opened 2016-03-19
Parameter could / should have __array__() method


newville:
The discussions of #319, #321, #322 reminded me of the original motivation for `Parameter.__float__()`, which is that
```python
>>> p  = Parameter('x', 3.9)
>>> float(p)
3.9
>>> p + 1
4.9
```

but unfortunately,  
```python
>>> import numpy as np
>>> np.linspace(-1, 1, 101)*p
```
is an "object array", and so essentially useless.

**But**:  I believe that if `Parameter` has an `__array__()` method, simply

```python 
    def __array__(self):
         return np.array(float(self._getval()))
```

Then this will work to give a proper ndarray.  

That would mean that all the examples and uses would be able to drop the explicit access of the `.value` attribute, so that 
```python
def fcn2min(params, x, data):
    """ model decaying sine wave, subtract data"""
    amp = params['amp'].value
    shift = params['shift'].value
    omega = params['omega'].value
    decay = params['decay'].value

    model = amp * np.sin(x * omega + shift) * np.exp(-x*x*decay)
    return model - data
```
could become (for example)

```python
def fcn2min(params, x, data):
    """ model decaying sine wave, subtract data"""
    shift = params['shift'] 
    omega = params['omega'] 
    decay = params['decay']

    model = params['amp'] * np.sin(x * omega + shift) * np.exp(-x*x*decay)
    return model - data
```

which seems much nicer to me...  I have not tested this in any detail.  Any objections to pursuing this idea?

tritemio:
+1

andyfaff:
Can one do this for `Parameters` objects? e.g. instead of

    params = np.array([parameters[param].value for param in parameters],
float)

could you do (or something like it):

    params = np.array(parameters)

newville:
@andyfaff  I'm not 100% sure, but I think this might be possible.  Of course, that removes the dictionary-ness, but sometimes that's OK.

andyfaff:
>
> The dict-ness is still there if you want it, but it might  be a clean way
to extract all the numerical parameter values into an array.
The array form is very useful as well, a lot of peoples cost functions
would have originally been programmed in this way becasue that's the form
that the scipy minimizers accept.

newville:
@andyfaff  Yeah, I understand, though (perhaps unsurprisingly) I think it would be better if the scipy minimizers had a better data structure for parameter values than a simple array -- the order actually has no inherent meaning.

[...]

PR #320, opened 2016-03-17
Improve pretty printing of Parameters

tritemio:
Similar to #318.

`Parameters` has a `pretty_print()` which prints with the following style:

```
Parameters({
    'p1_sigma': <Parameter 'p1_sigma', 0.03, bounds=[0.01:0.2]>, 
    'p1_fwhm': <Parameter 'p1_fwhm', 0.0706446, bounds=[-inf:inf], expr='2.3548200*p1_sigma'>, 
    'p2_sigma': <Parameter 'p2_sigma', 0.03, bounds=[0.01:0.2]>, 
    'p2_fwhm': <Parameter 'p2_fwhm', 0.0706446, bounds=[-inf:inf], expr='2.3548200*p2_sigma'>, 
    'p1_center': <Parameter 'p1_center', 0.3, bounds=[-0.1:0.7]>, 
    'p2_center': <Parameter 'p2_center', 0.85, bounds=[0.5:1.1]>, 
    'p1_amplitude': <Parameter 'p1_amplitude', 1, bounds=[0.01:inf]>, 
    'p2_amplitude': <Parameter 'p2_amplitude', 1, bounds=[0.01:inf]>, 
    'br_amplitude': <Parameter 'br_amplitude', 0.0001, bounds=[0:inf]>, 
    'br_center1': <Parameter 'br_center1', 0.3, bounds=[0:inf], expr='p1_center'>, 
    'br_center2': <Parameter 'br_center2', 0.85, bounds=[0:inf], expr='p2_center'>, 
    'br_sigma1': <Parameter 'br_sigma1', 0.03, bounds=[0:inf], expr='p1_sigma'>, 
    'br_sigma2': <Parameter 'br_sigma2', 0.03, bounds=[0:inf], expr='p2_sigma'>, 
    })
```

This follows a style typical for object representations in python (`__repr__`) but is hard to read for a user.

This PR modifies `pretty_print()` in backward-compatible way  (not removing arguments) in order to produce the following output:

```
Name               Value        Min        Max       Vary    Expr      
br_amplitude      0.0001          0        inf       True    None
br_center1           0.3          0        inf      False    p1_center
br_center2          0.85          0        inf      False    p2_center
br_sigma1           0.03          0        inf      False    p1_sigma
br_sigma2           0.03          0        inf      False    p2_sigma
p1_amplitude           1       0.01        inf       True    None
p1_center            0.3       -0.1        0.7       True    None
[...]

PR #325, opened 2016-03-20
Rename argument `spacing` to `colwidth` for consistency with Paramete…

tritemio:
Rename `Model.print_param_hints()` argument `spacing` to `colwidth` for consistency with `Parameters.pretty_print()`.

### DIFF
--- a/lmfit/model.py
+++ b/lmfit/model.py
@@ -242,22 +242,22 @@ class Model(object):
             else:
                 warnings.warn(self._invalid_hint % (key, name))
 
-    def print_param_hints(self, spacing=8):
+    def print_param_hints(self, colwidth=8):
         """Pretty-print parameters hints.
 
-        Argument `spacing` is column width except for first and last columns.
+        Argument `colwidth` is column width except for first and last columns.
         """
         name_len = max(len(s) for s in self.param_hints)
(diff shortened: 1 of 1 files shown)


newville:
@tritemio  That change seems OK to me.  But, I should also say that I noticed that neither of the `pretty_print()` methods added with #318 or #320 are actually documented.   Sorry for not noticing this earlier, but that really should be fixed.

tritemio:
I just realized we don't use sphinx's autodoc to generate the API docs automatically! We should definitely switch, maintaining two separate documentation for each function/object will not scale and will end up the two going out-of-sync.

The new print functions are already documented in their docstrings. Should I copy and paste this documentation in the docs files for now?

newville:
@tritemio I'd be OK with using autodoc with sphinx.  My preference would be to use the napoleon extension, though I'd be open to better suggestions.  At one point awhile ago we had trouble with IPython in the docs, and it would be nice to fix that.

I think the switch of the docs and maybe docstrings would be a fair amount of work.  Since we've been steadily adding small features and fixes for awhile, and have a few potentially larger changes (such as #323) in the queue,  I suggest getting 0.9.3 ready for release soon (either with or without docs for the pretty_print methods pasted in -- your choice) and postponing any doc refactoring to the next release.

PR #326, opened 2016-03-23
Docs fixes

tritemio:
This PR enables autodoc, napoleaon and mathjax sphinx extension and fix all warnings during build.

Seems like all the docs are correctly generated, so this can be a good starting point for a gradual switch to autodoc generated API docs.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ sandbox/
 .idea/
 .DS_Store
 .ipynb_checkpoints/
+doc/html
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -65,6 +65,7 @@ copyright = u'2014, Matthew Newville, The University of Chicago,  Till Stensitzk
 # built documents.
 #
 # The short X.Y version.
+sys.path.insert(0, os.path.abspath('../'))
 try:
     import lmfit
     release = lmfit.__version__
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -24,13 +24,18 @@ sys.path.append(os.path.abspath(os.path.join('.')))
 # coming with Sphinx (named 'sphinx.ext.*') or your custom ones.
 from extensions import extensions
 
+extensions = [
+    'sphinx.ext.autodoc',
+    'sphinx.ext.napoleon',
+    'sphinx.ext.mathjax',
+    ]
+
 try:
     import IPython.sphinxext.ipython_directive
     extensions.extend(['IPython.sphinxext.ipython_directive',
                        'IPython.sphinxext.ipython_console_highlighting'])
 except ImportError:
     pass
-
 
 intersphinx_mapping = {'py': ('http://docs.python.org/2', None),
                        'numpy': ('http://docs.scipy.org/doc/numpy/', None),
@@ -173,4 +178,3 @@ latex_documents = [
    'Non-Linear Least-Squares Minimization and Curve-Fitting for Python',
    'Matthew Newville, Till Stensitzki, and others', 'manual'),
 ]
-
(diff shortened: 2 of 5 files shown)


tritemio:
The latest commit shows that we can use autodoc to document one method (`Parameters.pretty_print`), while the rest of the class documentation is still manually generated.

newville:
@tritemio   Hm, That mixes a lot of separate changes.   Many of the fixes are very good, so thanks!  

But, some of the changes seem wrong to me:  1) there should not be a `doc/html` folder at all, and 2) I don't like the idea of using `sys.path.insert(0, '..')` in `conf.py` -- I think the docs should get the version from a working. installed environment.

For mixing autodoc and hand-coded rst, I guess that's one approach, but it encourages laziness and messiness.   I think we might have to be a little more rigorous and strict about documenting modules going forward.  Like, autodoc is OK, but I think there are things that should go in the main docs that are different from what should appear in the doc-strings.    Writing a doc string and 
```
    .. automethod:: mymethod
```

in the rst file makes is not really a substitute for documentation.

newville:
@tritemio All that said, I'll merge as is.  Thanks!

tritemio:
The `sys.path.insert(0, os.path.abspath('../'))` is needed because it does not make sense to use the RST files from current folder and the docstrings from the installed version of lmfit. The two can be very different and you end up generating an inconsistent documentation.

With `sys.path.insert(0, os.path.abspath('../'))`, instead, you always generate all the documentation from the version checked out in current folder, which is the right thing to do IMHO. Plus, since we are using versioneer, the lmfit version shown in the documentation will refer the exact commit.

BTW, I didn't mention it before, the command I use to generate the docs is:

```
$ sphinx-build -a -b html . html/
```

and it is run from inside the docs folder in lmfit source tree.

newville:
@tritemio 

> The sys.path.insert(0, os.path.abspath('../')) is needed because it does not make sense to use the 
> RST files from current folder and the docstrings from the installed version of lmfit. The two can be 
> very different and you end up generating an inconsistent documentation.

I think it is better to build and install (in a virtualenv if needed) the library before building the docs, don't import from the source directory.   That way, you will be referencing what will  actually be (and has been) installed, not what is in the source folder.  

Also, make sure you're working on a commited (if not pushed or merged) branch so that the version string from versioneer is correct.

> BTW, I didn't mention it before, the command I use to generate the docs is:
> 
> $ sphinx-build -a -b html . html/
> 
> and it is run from inside the docs folder in lmfit source tree.

That doesn't follow the convention of how the Makefile works, which puts all the built documentation into `_build` (which is already listed in `.gitignore`).

tritemio:
The advantage of the sys.path trick is that I don't even need to commit. I can try the edits/formattings until they work, then commit. It is important to have fast iteration of edit-and-build for docs.

You can achieve the same thing by installing the package in development mode and removing the sys.path trick. Note I'm not saying you should import from the source tree. Only to build the docs it's a passable trick, since the only reason the import lmfit in this case is getting the version string.

tritemio:
Regarding the second point I agree, better using:

```
$ sphinx-build -a -b html . _build/
```

I couldn't find the MakeFile so I typed the command manually.

newville:
@tritemio 
> The advantage of the sys.path trick is that I don't even need to commit. I can try the
>  edits/formattings until they work, then commit. It is important to have fast iteration of
>  edit-and-build for docs.

You don't have to commit the docs or the code to a git branch, but you do have to install the module.  This is proves the code is actually installed as you expect.   Put it this way: the tests do not play tricks with sys.path,  they run the installed code.

Of course, using a virtual environment is highly recommended.   As is building the docs using `make` with `doc/Makefile`.

PR #330, opened 2016-04-07
Minor fix for lmfit.ui.basefitter.BaseFitter in the case that model does not implement guess

gitj:
This fixes a simple error in lmfit.ui.basefitter.BaseFitter.guess.
Currently, if you try to use a model that does not implement a guess method, the code detects the NotImplmentedError and indicates guessing failed, but it also tries to update self.current_params = guess which raises an UnboundLocalError since guess never successfully got defined.

This can easily happen by simply creating a CompositeModel and not defining a guess method for it.

This pull request simply moves the assignment of self.current_params = guess such that it only happens if "guess" is defined. If guess fails, self.current_params shouldn't be updated anyway.

### DIFF
--- a/lmfit/ui/basefitter.py
+++ b/lmfit/ui/basefitter.py
@@ -149,7 +149,7 @@ class BaseFitter(object):
                 val = self.kwargs[key]
                 d = {key: val}
                 guess = self.model.guess(self._data, **d)
-                self.current_params = guess
+            self.current_params = guess
         except NotImplementedError:
             guessing_successful = False
         return guessing_successful
--- a/lmfit/ui/basefitter.py
+++ b/lmfit/ui/basefitter.py
@@ -149,9 +149,9 @@ class BaseFitter(object):
                 val = self.kwargs[key]
                 d = {key: val}
                 guess = self.model.guess(self._data, **d)
+                self.current_params = guess
         except NotImplementedError:
             guessing_successful = False
-        self.current_params = guess
         return guessing_successful
 
     def __assign_deps(self, params):


newville:
@gitj @danielballan should that be `self._current_params` or `self.current_params` here?   

Also: shouldn't that be set for all cases that can be guessed?

gitj:
@newville I think self.current_params is correct here, because self._current_params is initialized at creation to self.model.make_params() [here](https://github.com/lmfit/lmfit-py/blob/master/lmfit/ui/basefitter.py#L98) and then the intention of the self.current_params setter is to ensure that the same Parameters objects are kept with the fitter, with just the attributes being updated as needed. So if we directly set self._current_params, that would break this paradigm. I don't know that it would make any real difference in the end, but it seems more consistent to use self.current_params.

I don't know what you mean by "Also: shouldn't that be set for all cases that can be guessed?"

newville:
@gitj 

>  I think self.current_params is correct here, because self._current_params is initialized at
> creation to self.model.make_params() here and then the intention of the self.current_params 
> setter is to ensure that the same Parameters objects are kept with the fitter, with just the
> attributes being updated as needed. So if we directly set self._current_params, that would 
> break this paradigm. I don't know that it would make any real difference in the end, but it 
> seems more consistent to use self.current_params.

OK, relying on the property setter seems reasonable.  It would be nice to have tests!

> I don't know what you mean by "Also: shouldn't that be set for all cases that can be guessed?"

Well,  it seems like you set `self.current_params` only if `count_indep_vars == 1`.   If it is 0, you have a `guessing_successful` being `True`, but there is no change to the `current_params`.

gitj:
@newville Good catch regarding the count_indep_vars. I had accidentally inserted an additional level of indentation that I didn't intend. I've removed that extra indent.

I can write a quick test for the problem that this pull request solves, but it looks like there are no tests for the ui Fitter stuff that I can see. Should I just add an extra file?

newville:
OK, merging since no complaints

PR #331, opened 2016-04-15
MAINT: de shouldn't need bounds if not varying

andyfaff:
Currently the `differential_evolution` option in `scalar_minimize` requires finite upper and lower bounds for all parameters. This PR removes that requirement. Instead finite upper and lower bounds are only required for those parameters that vary.


newville:
@andyfaff OK, I think there can't be any objections to that.  I'll merge right away (else I'll forget!), but will also take complaints if some discussion is wanted.

Issue #334, opened 2016-05-18
Deprecating python 2.6?


ericdill:
Python 2.6 is no longer part of the automated test suite in lmfit.  It seems to still sort of be supported since `ordereddict` is being vendored which is only needed for python 2.6.  Are there plans to officially remove py26 support from lmfit at any point?

newville:
@ericdill I don't have a strong preference.  I think there aren't many folks who need to use Py26.   If "OrderedDict" is the only known thing that wouldn;t work on Py26, I don't see the harm in keeping it.  I suspect that more incompatibilities may creep in, if they haven't already. 

I think a deprecation warning in the next version saying that support will be dropped in the following version would be OK.   Do you agree?

ericdill:
> I think a deprecation warning in the next version saying that support will be dropped in the following version would be OK.

That seems quite reasonable.  Another good reason to drop python 2.6 is that you can only use dict comprehensions with py2.7+

newville:
closing

PR #335, opened 2016-05-28
FIX: Changed dict comprehensions to style compatible with py26

pseudocubic:
This PR addresses Issue #334 

Dictionary comprehensions are changed to a style compatible with Python 2.6, i.e. `dict((key, value) for key in list)`. This should continue to maintain basic py26 compatibility for now, even if it is not officially supported. No additional tests should be required.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ language: python
 sudo: false
 
 python:
+    - 2.6
     - 2.7
     - 3.3
     - 3.4
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -11,6 +11,9 @@ Non-Linear Least-Square Minimization and Curve-Fitting for Python
 
   Upgrading scripts from version 0.8.3 to 0.9.0?  See  :ref:`whatsnew_090_label`
 
+.. warning::
+
+  Support for Python 2.6.x will be officially dropped following version 0.9.4 (in 0.9.5)
 
 Lmfit provides a high-level interface to non-linear optimization and curve
 fitting problems for Python. Lmfit builds on and extends many of the
--- a/doc/installation.rst
+++ b/doc/installation.rst
@@ -13,11 +13,11 @@ Prerequisites
 The lmfit package requires Python, Numpy, and Scipy.  Scipy version 0.13 or
 higher is recommended, but extensive testing on compatibility with various
 versions of scipy has not been done.  Lmfit works with Python 2.7, 3.3 and
-3.4.  No testing has been done with Python 3.5, but as the package is pure
-Python, relying only on scipy and numpy, no significant troubles are
-expected.  The `nose`_ framework is required for running the test suite,
-and IPython and matplotib are recommended.  If Pandas is available, it will
-be used in portions of lmfit.
+3.4.  Lmfit works with Python 2.6 only for versions <= 0.9.4. No testing has
+been done with Python 3.5, but as the package is pure Python, relying only on
+scipy and numpy, no significant troubles are expected.  The `nose`_ framework
+is required for running the test suite, and IPython and matplotib are
+recommended. If Pandas is available, it will be used in portions of lmfit.
 
 
 Downloads
--- a/lmfit/__init__.py
+++ b/lmfit/__init__.py
@@ -33,6 +33,8 @@ from scipy.optimize.  It has a number of useful enhancements, including:
              Daniel B. Allen, Johns Hopkins University
              Antonino Ingargiola, University of California, Los Angeles
 """
+import warnings
+import sys
 
 from .minimizer import minimize, Minimizer, MinimizerException
 from .parameter import Parameter, Parameters
@@ -49,5 +51,10 @@ from .uncertainties import ufloat, correlated_values
 
 ## versioneer code
 from ._version import get_versions
+
 __version__ = get_versions()['version']
 del get_versions
+
+# PY26 Depreciation Warning
+if sys.version_info[:2] == (2, 6):
+    warnings.warn('Support for Python 2.6.x is depreciated in Lmfit 0.9.4 and will be dropped in 0.9.5', DeprecationWarning)
--- a/lmfit/asteval.py
+++ b/lmfit/asteval.py
@@ -86,28 +86,28 @@ class Interpreter:
         symtable['print'] = self._printer
 
         # add python symbols
-        py_symtable = {sym: __builtins__[sym] for sym in FROM_PY
-                              if sym in __builtins__}
+        py_symtable = dict((sym, __builtins__[sym]) for sym in FROM_PY
+                              if sym in __builtins__)
         symtable.update(py_symtable)
 
         # add local symbols
-        local_symtable = {sym: obj for (sym, obj) in LOCALFUNCS.items()}
+        local_symtable = dict((sym, obj) for (sym, obj) in LOCALFUNCS.items())
         symtable.update(local_symtable)
 
         # add math symbols
-        math_symtable = {sym: getattr(math, sym) for sym in FROM_MATH
-                              if hasattr(math, sym)}
+        math_symtable = dict((sym, getattr(math, sym)) for sym in FROM_MATH
+                              if hasattr(math, sym))
         symtable.update(math_symtable)
 
         # add numpy symbols
         if self.use_numpy:
-            numpy_symtable = {sym: getattr(numpy, sym) for sym in FROM_NUMPY
-                              if hasattr(numpy, sym)}
+            numpy_symtable = dict((sym, getattr(numpy, sym)) for sym in FROM_NUMPY
+                              if hasattr(numpy, sym))
             symtable.update(numpy_symtable)
 
-            npy_rename_symtable = {name: getattr(numpy, sym) for name, sym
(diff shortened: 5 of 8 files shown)


newville:
@pseudocubic @ericdill    I could be persuaded, but I'd rather not.  The (short) discussion at #334 led to the plan to include a deprecation warning in the next release (0.9.4) that is raised when used with Python 2.6, and to remove Python2.6-isms (like having to test for collections.OrderedDict) in the following release (0.9.5).    In fact, we sort of abandoned Python 2.6, but haven't explicitly stated so.  The next release will fix that statement, but not change the actual policy -- we don't test or expect lmfit to work with Python 2.6.

My view is that If people are really stuck with Python 2.6,  also being stuck with lmfit 0.9.3  (or 0.9.4) will be the least of their problems. 

Any objections, suggestions, other views?

pseudocubic:
I certainly understand not wanting to continue to support py26 in an official capacity, but I was hopeful that this change would be relatively benign with the side effect that py26 is once again supported. It changes no functionality, only style.

I couldn't ascertain when the dictionary comprehensions this PR "fixes" were introduced, but certainly they are included in lmfit 0.9.3, so attempts to import lmfit result in a Syntax Error. It would be nice from my perspective to nail down an exact final py26 compatible version, since the depreciation warning hasn't been included yet, but will in the next release, indicating that the following release will stop support.

My own selfish reason for wanting this final release is that there are some systems I use for data analysis which still use RHEL 6.x, which defaults to py26, so having an up-to-date release that I can use would be ideal.

ericdill:
> some systems that I use for data analysis ... default to py26

Can you use conda instead?  Then you get the benefit of a sandboxed environment that can run whatever version of python that you want

ericdill:
> I couldn't ascertain when the dictionary comprehensions this PR "fixes" were introduced

Looks like most of these dict comprehensions were introduced in #261 which went in to 0.9.2, so, in theory, 0.9.1 *might* be py26 compatible

newville:
I suppose we could accept this PR and be more definitive about support what versions support 2.6.  For example, we could make master and 0.9.4 known to work with Python 2.6,  0.9.5 explicitly deprecate 2.6, and 0.9.6 intentionally break with 2.6.   That is, I don't think support for 2.6 should be a high priority, but I don't think we lose a lot by maintaining it either.   

Do others have any suggestion, opinions?

andyfaff:
The PR is fairly innocuous with respect to changes. But it does indicate
that the lmfit project needs to come up with a clear policy for deprecation
and oldest supported version.
I think that the minimum supported version for numpy and scipy are Python
>= 2.7 and >= 3.3.

For a release I think the documentation needs to state under what
conditions it's expected to work, e.g. Python >=2.7, numpy >= ?, scipy >=
?. However, I see no reason why we can't explicitly say that 0.9.4 is going
to explicitly support a minimum of Py >= 2.7 or 3.3? It's not clear to me
why we need to have a deprecation cycle?

A.



On 30 May 2016 at 10:16, Matt Newville <notifications@github.com> wrote:

> I suppose we could accept this PR and be more definitive about support
> what versions support 2.6. For example, we could make master and 0.9.4
> known to work with Python 2.6, 0.9.5 explicitly deprecate 2.6, and 0.9.6
> intentionally break with 2.6. That is, I don't think support for 2.6 should
> be a high priority, but I don't think we lose a lot by maintaining it
> either.
>
> Do others have any suggestion, opinions?
>
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly, view it on GitHub
> <https://github.com/lmfit/lmfit-py/pull/335#issuecomment-222391192>, or mute
> the thread
> <https://github.com/notifications/unsubscribe/AAq51lDocWnebvjnqqZ2cdav-jhccb3Tks5qGixUgaJpZM4IpHjz>
> .
>



-- 
Dr. Andrew Nelson


newville:
@andyfaff wrote:

> For a release I think the documentation needs to state under what conditions it's expected to 
> work, e.g. Python >=2.7, numpy >= ?, scipy >= ?.

The docs and testing are explicit about supporting Python 2.7, 3.3, 3.4, and 3.5.   So it's not like we say it works with Python 2.6.    

The docs recommend scipy > 0.13, but don't give a minimal scipy and numpy version because we don't actually test against different scipy and numpy versions.   I think lmfit might now require scipy > 0.14, as it uses OptimizeResult.    I don't recall any complaints about which versions of scipy is needed,  but if someone (you?) wants to set up a more thorough test suite of such packages, that would be great. 

> However, I see no reason why we can't explicitly say that 0.9.4 is going to explicitly support a
> minimum of Py >= 2.7 or 3.3? It's not clear to me why we need to have a deprecation cycle?

That's what we did with 0.9.1.   And, I'm OK with leaving it as it is.

It's true that lmfit used to work with Python 2.6 and that "Py27"isms crept in without fanfare or a deprecation cycle.  There are still some "Py26"isms (carrying around our own copy of OrderedDict), but we could drop these. 

Supporting 2.6 is not a huge cost, but the benefit is also very small - at this point it is down to RHEL6 users who are unable to upgrade or use a virtual environment but want the latest version on lmfit.  I think that might be 1 person, and they already made the needed changes on their own branch ;).

So, should we explicitly support Python2.6 only to drop it ASAP with a "proper deprecation cycle", which likely means raising an obnoxious Warning?   I'm not sure a deprecation warning from lmfit is going to finally make someone realize that they need to move from Python 2.6!

pseudocubic:
FYI, I confirmed that 0.9.1 is the last py26 compatible version, in so far as it imports without error, and I can perform some basic fits without problem; I didn't perform extensive testing to check, but since this PR passes all tests after these small changes, I presume that it probably all works.

> So, should we explicitly support Python2.6 only to drop it ASAP with a "proper deprecation cycle", which likely means raising an obnoxious Warning? I'm not sure a deprecation warning from lmfit is going to finally make someone realize that they need to move from Python 2.6!

Personally, I just think it's a good practice to have an official depreciation cycle, but the obnoxious warning should only be seen by those with py26, and shouldn't have any effect on most of your users anyway. I really doubt that the majority of those still using py26 wants to, so the warning won't really serve to get them to switch, just serve as a notification that future versions will no longer work. 

At the very least, documentation showing the compatibility of different versions would be really nice, as @andyfaff suggested. That is, somewhere, probably in the Prerequisites section of the docs, it should be explicitly stated what the final py26 compatible version is.

newville:
@pseudocubic 

> Personally, I just think it's a good practice to have an official depreciation cycle, but the 
> obnoxious warning should only be seen by those with py26, and shouldn't have any effect 
> on most of your users anyway. I really doubt that the majority of those still using py26 wants 
> to, so the warning won't really serve to get them to switch, just serve as a notification that 
> future versions will no longer work.

I'm OK with a deprecation cycle.   I have to admit that being told that something should be done simply because it is "Good Practice" usually increases my skepticism.   I think the question should be: is there a benefit to a deprecation cycle?   And, well, is there?

We did go more than six months and two releases before anyone noticed or cared that Python2.6 no longer worked, so the user base is very, very small, and possibly already saturated.
 
If we do want to have a deprecation cycle, I think this PR that restores Py2.6 compatibility should also create such a deprecation cycle, including changes to the documentation and raising a Warning.

> At the very least, documentation showing the compatibility of different versions would be 
> really nice, as @andyfaff suggested. That is, somewhere, probably in the Prerequisites 
> section of the docs, it should be explicitly stated what the final py26 compatible version is.

I do not want to maintain historical information for dependencies. Do you?  The docs say what versions are tested and supported for the current version.

tacaswell:
For 2.6 support I always point people to http://www.curiousefficiency.org/posts/2015/04/stop-supporting-python26.html .  

For some actual usage numbers, see https://astrofrog.github.io/blog/2015/05/09/2015-survey-results/ which gives ~3% of users on 2.6.

matplotlib dropped 'official' support with 1.5.0 ( Oct 29, 2015) and 2.0 (beta out now) will not import on 2.6.

newville:
@tacaswell   Thanks -- I agree that we should stop enabling py26.   I have a hard time finding enthusiasm for this PR, and it definitely needs some work, including documenting the deprecation.

Is anyone motivated to work on this?

newville:
@pseudocubic thanks for the updates.   I know this is nit-picking but I think the wording should be be (support for Py2.) "is deprecated in 0.9.4 and will be dropped in 0.9.5".   No ambiguity.

pseudocubic:
@newville I updated in two places to make the depreciation path more clear as per your suggestion, I agree that no ambiguity is better

BTW, for PRs for lmfit is it preferable that I squash these commits into one?

newville:
@pseudocubic Thanks!  I don't think it's necessary to minimize commits.

At this point, I think this PR is OK to merge.   Are there any objections to making master and 0.9.4 compatible with Python 2.6, but with a clear warning, and then being clear about dropping support for Python 2.6 in 0.9.5?   If I don't hear objections or discussion, I'll plan on merging this on Monday, June 13th.

ericdill:
You might consider adding py2.6 back to Travis to make sure the test suite actually passes on py2.6

newville:
@ericdill good suggestion.

newville:
OK, with all tests passing and no objections, I'm merging this.  Thanks very much @pseudocubic.

PR #336, opened 2016-06-12
Add `__array__()` methods to Parameter and Parameters classes  

newville:
This add `__array__()` methods to the `Parameter` and `Parameters` classes.   For `Parameters()`, this allows one to easily get the values of a Parameters instance as an ndarray.

For Parameter, this allows implicitly accessing a Parameter in an expression using ndarrays.  Along with the existing `__float__()` method this means that one can reliably use a Parameter directly in an expression, so that one can replace
```python
   offset = params['offset'].value
   slope = params['slope'].value
   model = offset + slope * x
```
with the more readable and possibly more obvious 
```python
   model = params['offset'] + params['slope'] * x
```

The examples, tests, and docs have been updated to use the simpler approach.


andyfaff:
Will the __array__ method work OK with addition because there are also the
__add__ and __iadd__ methods?

newville:
@andyfaff    Yes, I think so:

```python 
>>> import lmfit
>>> import numpy as np
>>> x = np.linspace(0, 1, 11)
>>> p1 = lmfit.Parameters()
>>> p1.add('sigma', value=3)
>>> p2 = lmfit.Parameters()
>>> p2.add('alpha', value=100, min=0)

>>> p1 + p2
Parameters([('sigma', <Parameter 'x', 3, bounds=[-inf:inf]>), ('alpha', <Parameter 'alpha', 100, bounds=[0:inf]>)])

>>> p1['sigma'] + p2['alpha']
103

>>> np.array(p1+p2)
array([   3.,  100.])
>>> np.array(p2+p1)
array([ 100.,    3.])

>>> x * p1['alpha']
array([ 0. ,  0.3,  0.6,  0.9,  1.2,  1.5,  1.8,  2.1,  2.4,  2.7,  3. ])
```

There might be something we're not testing.  But all current tests pass, without changing the tests themselves except to remove the `.value` part of 
```python 
  attr = params['attr'].value   # old, still valid
  attr = params['attr']   # new version only
```
in the objective functions.

Tillsten:
While i really like adding the dunder methods to Parameter, i am not yet convinced about the usefulness of adding them to the Parameters class. Any useful application for that? Your example gives a good reason against it, imo:
```
>>> np.array(p1+p2)
array([   3.,  100.])
>>> np.array(p2+p1)
array([ 100.,    3.])
```
I know why it happens, but i think it will confuse users.

newville:
@Tillsten   I don't have a strong preference.   In the discussion for Issue #323 (which this PR addresses), @andyfaff  suggested that it would make it easier to transition code from scipy.optimize or other code that requires parameter values in an array.   That seems like a decent reason to me.   Currently you'd have to do
```python
pararray = np.array(params.valuesdict().values())
```
to get an array of parameter values. Replacing that with
```python
pararray = np.array(params)
```
seems a bit more obvious.

Still, I agree that accessing parameters by name is much better than index, so I don't think we want to encourage this for new code.

Tillsten:
Oh, makes sense. I am using the parameter to array conversion quite a lot myself. I think it is the combination with 'Parameters.__add__' which makes me feel a little uneasy, but i think you convinced me.

tacaswell:
The difference between `p1+p2` and `p2+p1` is not that worrying, addition is just non-commutative (like `+` on python lists or strings).

newville:
any objections to merging this?

Issue #333, opened 2016-05-06
Evaluating a model at a scalar value raises error


gitj:
Trying to execute lmfit.model.Model.eval(x=100.,....) runs into a problem at this line:
https://github.com/lmfit/lmfit-py/blob/master/lmfit/model.py#L412

`    # Handle special case of constant result and one
     # independent variable (of any dimension).
     if np.ndim(result) == 0 and len(self.independent_vars) == 1:
          result = np.tile(result, kwargs[self.independent_vars[0]].shape)`

Since kwargs[self.independent_vars[0]] is a scalar, it has no .shape attribute. There needs to be a check for this case.

newville:
OK, fixed with #342

PR #337, opened 2016-06-14
ENH: raise error if userfcn returns non-finite number

andyfaff:
As discussed on ML, raise a `ValueError` if `userfcn` returns a non-finite number. This is because all of the scipy optimizers do not work in such a situation. However, the failure is insidious: the fit success is still `True` and the optimized parameters are just the initial guesses. Thus, the fit appears to work, when it is in fact not doing anything at all.
It was discussed on ML whether to raise an Error or a warning. My preference is for Error because I consider it to be an Error state. With a warning you might suspect that the situation isn't as bad as it really is.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -25,3 +25,37 @@ The LMFIT-py code is distribution under the following license:
   LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
   FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
   DEALINGS IN THIS SOFTWARE.
+
+-----------------------------------------
+Some code sections have been taken from the scipy library whose licence is below.
+
+Copyright (c) 2001, 2002 Enthought, Inc.
+All rights reserved.
+
+Copyright (c) 2003-2016 SciPy Developers.
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are met:
+
+  a. Redistributions of source code must retain the above copyright notice,
+     this list of conditions and the following disclaimer.
+  b. Redistributions in binary form must reproduce the above copyright
+     notice, this list of conditions and the following disclaimer in the
+     documentation and/or other materials provided with the distribution.
+  c. Neither the name of Enthought nor the names of the SciPy Developers
+     may be used to endorse or promote products derived from this software
+     without specific prior written permission.
+
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDERS OR CONTRIBUTORS
+BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY,
+OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF
+THE POSSIBILITY OF SUCH DAMAGE.
--- a/doc/fitting.rst
+++ b/doc/fitting.rst
@@ -484,12 +484,12 @@ For full control of the fitting process, you'll want to create a
    :type  iter_cb:  callable or ``None``
    :param scale_covar:  flag for automatically scaling covariance matrix and uncertainties to reduced chi-square (``leastsq`` only)
    :type  scale_covar:  bool (default ``True``).
-   :param mask_non_finite:  if ``userfcn`` returns non-finite values then a
-                            ``ValueError`` is raised. However, if
-                            ``mask_non_finite`` is set to ``True`` then those
-                            non-finite values are ignored, allowing the fit to
-                            proceed with the finite values.
-   :type  mask_non_finite: bool (default ``False``)
+   :param nan_policy: Specifies action if `userfcn` (or a Jacobian) returns nan
+            values. One of:
+                'raise' - a `ValueError` is raised
+                'propagate' - the values returned from `userfcn` are un-altered
+                'omit' - the non-finite values are filtered.
+   :type  nan_policy: str (default 'raise')
    :param kws:      dictionary to pass as keywords to the underlying :mod:`scipy.optimize` method.
    :type  kws:      dict
 
--- a/doc/fitting.rst
+++ b/doc/fitting.rst
@@ -466,7 +466,7 @@ Using the :class:`Minimizer` class
 For full control of the fitting process, you'll want to create a
 :class:`Minimizer` object.
 
-.. class:: Minimizer(function, params, fcn_args=None, fcn_kws=None, iter_cb=None, scale_covar=True, **kws)
+.. class:: Minimizer(function, params, fcn_args=None, fcn_kws=None, iter_cb=None, scale_covar=True, mask_non_finite=False, **kws)
 
    creates a Minimizer, for more detailed access to fitting methods and attributes.
 
@@ -484,6 +484,12 @@ For full control of the fitting process, you'll want to create a
    :type  iter_cb:  callable or ``None``
    :param scale_covar:  flag for automatically scaling covariance matrix and uncertainties to reduced chi-square (``leastsq`` only)
    :type  scale_covar:  bool (default ``True``).
+   :param mask_non_finite:  if ``userfcn`` returns non-finite values then a
+                            ``ValueError`` is raised. However, if
+                            ``mask_non_finite`` is set to ``True`` then those
+                            non-finite values are ignored, allowing the fit to
+                            proceed with the finite values.
+   :type  mask_non_finite: bool (default ``False``)
    :param kws:      dictionary to pass as keywords to the underlying :mod:`scipy.optimize` method.
    :type  kws:      dict
 
--- a/lmfit/minimizer.py
+++ b/lmfit/minimizer.py
@@ -1326,9 +1326,6 @@ def _nan_policy(a, nan_policy='raise', handle_inf=True):
     """
 
     policies = ['propagate', 'raise', 'omit']
-    if nan_policy not in policies:
-        raise ValueError("nan_policy must be one of {%s}" %
-                         ', '.join("'%s'" % s for s in policies))
 
     if handle_inf:
         handler_func =  lambda a: ~np.isfinite(a)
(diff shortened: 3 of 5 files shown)


andyfaff:
@newville, so I've got this working. However, there is one test that fails, [test_lists_become_arrays](https://github.com/lmfit/lmfit-py/blob/master/tests/test_model.py#L221).

The constructor of the `Model` class accepts a [`missing`](https://github.com/lmfit/lmfit-py/blob/master/lmfit/model.py/#L63) keyword which instructs the fitting what to do if the data has null or missing observations. The options are `none`, `drop` and `raise`.
The `drop` option is the equivalent in this PR of `mask_non_finite=True`.
The `raise` option is the equivalent in this PR of `mask_non_finite=False`.
The `none` option leaves the NaN/Inf in the data.
There is purposefully no equivalent in this PR for the `none` option, if there are NaN or Inf received from `userfcn` the scipy minimizers don't work.

The reason the test fails is that it has a `None` in the dependent variables and so the userfcn returns a NaN.

One way of proceeding is to make the `missing` keyword only accept `drop` or `raise`. Whilst this causes back compatibility issues I'd say that the fact that the minimizers don't work properly if `userfcn` returns a NaN to be a bigger issue.

newville:
@andyfaff  This look pretty good to me. My recollection is that `missing=[None, 'raise', 'drop']` is borrowed from pandas.  But I agree that, for this purpose `None` is not sufficient.   I'd suggest changing the test with `missing=None` and generating NaNs so that it is expected to raise ValueError.   Like you say, that's what *should* happen.
   
Consistency in argument names is generally preferred, which might suggest changing `mask_non_finite` to `missing`, but I much prefer a keyword arguments that are True/False,  So I prefer your `mask_non_infinite=[True, False]`, though I might suggest calling that `mask_nans` or `suppress_nans`.  Too picky?

tritemio:
IMHO the argument name should not contain `mask`. What about `ignore_nans` (True, False).

newville:
@tritemio  This may be too picky, but I find "ignore" ambiguous.  Perhaps `remove_nans` is better?

tritemio:
"Ignore nan" is widely used by numpy. But `remove_nans` would be a second best choice.

newville:
@tritemio Oh, I wasn't aware of that!  Is `ignore_nans` common in numpy / scipy?   If so (or, if there is another common way to express this in numpy/scipy), let's go with that.

andyfaff:
`ignore_*`, `filter_*`, `drop_*` are the other ideas. I've reported this kind of behaviour on the [scipy issues page](https://github.com/scipy/scipy/issues/6264), and on the scipy dev mailing list. It might be an idea to see what the scipy community says about it first, we could possibly address this at source.

Note - in `Minimizer.__residuals` this PR is currently raising `ValueError` on all non-finite numbers, i.e. NaN **and** +/- inf. This is because `scipy.optimize.leastsq` and `scipy.optimize.curve_fit` don't work with +/- Inf either. I should mention that the new function `scipy.optimize.least_squares` does raise a `ValueError` if the residuals contain non-finite numbers. It's unclear to me at the moment what to do with the scalar minimizers with regards to +/- Inf (NaN are always bad). At the moment the default cost metric is chi2 (there were some tentative steps at implementing others). If there is a single +/- Inf in a residuals vector then that Inf is going to be returned to the scalar minimizer. There may be some cost metric where it makes sense to return an +/- Inf, such as a negative log-likelihood, where Inf would represent an impossible solution. But I don't know how the gradient based solvers work in such a situation, how does it they know which direction to go in to achieve a lower cost metric?

newville:
@andyfaff On `nan` and `inf`:  This may be very old-school (years of Fortran), but my view is that `inf` is a special category of `nan`.  Having any non-finite floating point number is almost certainly going to cause problems.  So,  `ValueError` seems perfectly suited for this.

I think (but could be wrong) that nans and infs in the Jacobian are just as bad as in the residual.  

IMHO, but definitely open for suggestions:  

For  `*_nan`  vs `r*_non_finite`, I just find `*_non_finite` a bit too verbose, and will never remember if it is `*_non_finite` or `*_nonfinite` or `*_infinite`, and will probably type `*_non-finite` many times.   So, `*_nan` seems better, and would easily include +/-Inf, I think.

For `ignore_nans`, `mask_nans`, `remove_nans`, `drop_nans`, etc:   I find `ignore_*` ambiguous.  It says to "not pay attention to nan", but what we want to do is pay attention to them.  So,  `ignore_nans=True` would mean  "don't do anything", and `ignore_nans=False` would mean "do something", but is that something "raise ValueError" or "mask out"?    So, I prefer the more definitive `drop_nans` or `remove_nans` (default False),  which would raise ValueError if a nan (or inf) was encountered.

If scipy hasn't chosen a convention, perhaps we can persuade them by providing a good example.

argriffing:
fwiw, some corners of scipy have grown 'nan_policy' options like {'propagate', 'raise', 'omit'}. This is kind of ugly, and the root of the problem is that the numeric/scientific python stack below pandas has no unified notation to distinguish between NA (a missing number, like if someone didn't fill out an entry in a form or if the person watching the instrument was on vacation that day) vs. NaN (someone tried computing 0/0 or did something else nonsensical, or tried computing a function of NaN).

newville:
@argriffing  Thanks.  I think it would be really helpful to have something like `remove_nans` as a boolean option for many scipy and numpy functions.  Numpy's separate `nanmax` etc is kind of unfortunate, as is using `nan` to signify 'missing' (that's what `None` is for). 

I find scipy.stats `nan_policy` and panda's `missing` keywords with string values a bit ugly, but I could be persuaded to go with `nan_policy` if that is becoming the scipy convention (is it?).

But, I do agree with @andyfaff that we should be more proactive in handling nans and infs.

argriffing:
Here's the relevant scipy thread https://github.com/scipy/scipy/issues/571. I became involved a bit, because I've seen a lot of bug reports with all of the various things people have expected to happen when they call a scipy or numpy function with an ndarray containing a nan. Not to say that all those things should be supported. I don't know how well the `nan_policy` thing is working. Personally I don't use nans, and at least one PR https://github.com/scipy/scipy/pull/5808 seems to have been adversely affected by going out of its way to support it.

newville:
@argriffing Thanks for the link.  I agree with the main developers in that thread that pandas use of nan as 'missing' is unfortunate and should not be encouraged over the IEEE meaning.   It also suggests that a `nan_policy` keyword taking strings (and so extensible) is the preferred scipy solution.    

Any objections to using `nan_policy` with options 'propagate', 'omit', and 'raise', with a default of `raise`?

Tillsten:
+1 on following scipy

newville:
@andyfaff  Any thoughts on how to proceed with this? Is `nan_policy` acceptable?   It might be good to add some tests for a few cases (nan in residual arrays, in jacobian, None for parameter value) with different values for `nan_policy`.

tritemio:
+1 for following the scipy's `nan_policy`.

andyfaff:
I think `nan_policy` is a good name to use and that 'propagate', 'omit',
and 'raise' are good options. However, what did we want to do with the
objective functions returning np.inf?

On 24 June 2016 at 23:08, Matt Newville <notifications@github.com> wrote:

> @andyfaff <https://github.com/andyfaff> Any thoughts on how to proceed
> with this? Is nan_policy acceptable? It might be good to add some tests
> for a few cases (nan in residual arrays, in jacobian, None for parameter
> value) with different values for nan_policy.
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub
> <https://github.com/lmfit/lmfit-py/pull/337#issuecomment-228340711>, or mute
> the thread
> <https://github.com/notifications/unsubscribe/AAq51lp3GKpFGCxAXIY_8Ki3EnHvXctOks5qO9blgaJpZM4I09U2>
> .
>



-- 
Dr. Andrew Nelson


andyfaff:
Just checking - 'propagate' means don't do anything, 'omit' means filter
the non-finite values?

On 26 June 2016 at 12:09, Andrew Nelson <andyfaff@gmail.com> wrote:

> I think `nan_policy` is a good name to use and that 'propagate', 'omit',
> and 'raise' are good options. However, what did we want to do with the
> objective functions returning np.inf?
>
> On 24 June 2016 at 23:08, Matt Newville <notifications@github.com> wrote:
>
>> @andyfaff <https://github.com/andyfaff> Any thoughts on how to proceed
>> with this? Is nan_policy acceptable? It might be good to add some tests
>> for a few cases (nan in residual arrays, in jacobian, None for parameter
>> value) with different values for nan_policy.
>>
>> —
>> You are receiving this because you were mentioned.
>> Reply to this email directly, view it on GitHub
>> <https://github.com/lmfit/lmfit-py/pull/337#issuecomment-228340711>, or mute
>> the thread
>> <https://github.com/notifications/unsubscribe/AAq51lp3GKpFGCxAXIY_8Ki3EnHvXctOks5qO9blgaJpZM4I09U2>
>> .
>>
>
>
>
> --
> _____________________________________
> Dr. Andrew Nelson
>
>
> _____________________________________
>



-- 
Dr. Andrew Nelson


newville:
@andyfaff   Yes. I would say:

  *  'propagate' == do nothing, expect downstream solver to handle this (that is, the current behavior). I think is almost certain to cause problems, and probably be poorly reported, and should not be the default.
  * 'omit' == mask out any nan values from the residual before passing it back to the solver.  There is always a risk that the length of the residual changes between calls, which might cause all sorts of havoc.  Masking out nans will sometimes (perhaps even "often") work, but the user should be aware that this is a "field-applied bandage" and not a cure for having nans.
  * 'raise' == raise a ValueError if a nan is detected.  I think this should be the default.

As for 'inf' vs 'nan', I would say to treat all 'inf' and 'nan' values as 'nan's.  If we get to a case where the subtle differences between how these are handled by the solvers matters to someone we can deal with it then. 

Does that seem right to you?

newville:
@andyfaff Looks great -- I like the separate `_nan_policy` function, and its optional `handle_inf`.  Any objections to merging this?   

Somewhat related:  I think we have enough recent changes to tag 0.9.4 soon, even though 0.9.3 is only a few months old.   I'd like to address #333, but that should be easy.  Any opinions on what/when to tag 0.9.4 or if there are other things to include/fix?

andyfaff:
I think it's ready to merge, bar a licence question. The _nan_policy
function was taken from scipy code and then modified.

In the last commit I've pointed that out, and also appended the scipy
licence to lmfit's licence to reflect that portions of the lmfit code (e.g.
_differential_evolution) are based on scipy's.

I'm not sure if that's necessary, but I wanted to be safe.

When 0.9.4 is tagged what does that mean for minimum scipy, minimum Python
versions going forward?

A.

On 27 June 2016 at 04:38, Matt Newville <notifications@github.com> wrote:

> @andyfaff <https://github.com/andyfaff> Looks great -- I like the
> separate _nan_policy function, and its optional handle_inf. Any
> objections to merging this?
>
> Somewhat related: I think we have enough recent changes to tag 0.9.4 soon,
> even though 0.9.3 is only a few months old. I'd like to address #333
> <https://github.com/lmfit/lmfit-py/issues/333>, but that should be easy.
> Any opinions on what/when to tag 0.9.4 or if there are other things to
> include/fix?
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub
> <https://github.com/lmfit/lmfit-py/pull/337#issuecomment-228615526>, or mute
> the thread
> <https://github.com/notifications/unsubscribe/AAq51vd4Hxbfm6271OP2MIjAAuIMALXBks5qPsc8gaJpZM4I09U2>
> .
>



-- 
Dr. Andrew Nelson


newville:
@andyfaff  Thanks -- I think that there is a lot of code snippets lifted and/or then altered from scipy.optimize, so adding this to the license is probably overdue! I guess I/we were assuming that acknowledgements in the docs were OK, otherwise the license gets completely out of control, but I think it's better to include it.

For 0.9.4, I think we should  say:

  * Python 2.6, 2.7, 3.3, 3.4, and 3.5 are supported.  Python 2.6 is deprecated, and will not be supported in 0.9.5 or beyond.   There is not yet a plan for how long 3.3 will be supported.

  * Scipy:  I think we haven't been very careful about this.   I think we should go with 0.14 as *required* , so that OptimizeResult can be assumed.  I think we don't want to require 0.17 yet, but should say that `least_squares` is only available with 0.17.  

  * Numpy:  I don't know that we actually have a dependency issue.  `requirements.txt` says 1.5, but I don't know that older versions do not work.  I don't think we've tested this, but I also don't think we've ever run into a problem.

We could also mention minimum versions of IPython (4) and pandas(?) and maplotlib (?) if those options are used.   Then again, I'm also willing to say that we're not going to be testing out older versions forever.  

Does that seem right to you (and everyone else)?

newville:
@andyfaff  OK, merged.   Thanks!  I'll try to prepare 0.9.4 over the next week.

PR #338, opened 2016-06-16
Add support for least_squares, deleted deprecated methods.

Tillsten:
Sorry for using the wrong repo for putting up the new branch. 

This PR adds support for new least_squares optimizer in scipy 0.17 with
inbuild support for bounds and robust likelihoods. For my problems the solver
generally is much faster than leastsq.

Not implemented yet is error estimation from the covariance matrices. This
can be done in another PR. Test failures are from 3.3 and 2.6, due to old scipy
versions.

### DIFF
--- a/lmfit/minimizer.py
+++ b/lmfit/minimizer.py
@@ -52,6 +52,14 @@ try:
 except ImportError:
     pass
 
+# check for scipy.opitimize.least_squares
+HAS_NEW_LEAST_SQUARES_MIN = False
+try:
+    from scipy.optimize import least_squares
+    HAS_NEW_LEAST_SQUARES_MIN = True
+except ImportError:
+    pass
+
 from .parameter import Parameter, Parameters
 
 # use locally modified version of uncertainties package
@@ -270,7 +278,7 @@ class Minimizer(object):
 
         return dict([(name, p.value) for name, p in self.result.params.items()])
 
-    def __residual(self, fvars):
+    def __residual(self, fvars, apply_scaling=True):
         """
         Residual function used for least-squares fit.
         With the new, candidate values of fvars (the fitting variables), this
@@ -282,11 +290,16 @@ class Minimizer(object):
         if self._abort:
             return None
         params = self.result.params
-        for name, val in zip(self.result.var_names, fvars):
-            params[name].value = params[name].from_internal(val)
+
+        if apply_scaling:
+            for name, val in zip(self.result.var_names, fvars):
+                params[name].value = params[name].from_internal(val)
+                params.update_constraints()
+        else:
+            for name, val in zip(self.result.var_names, fvars):
+                params[name].value = val
         self.result.nfev += 1
 
-        params.update_constraints()
         out = self.userfcn(params, *self.userargs, **self.userkws)
         if callable(self.iter_cb):
             abort = self.iter_cb(params, self.result.nfev, out,
@@ -348,7 +361,8 @@ class Minimizer(object):
         """
         # determine which parameters are actually variables
         # and which are defined expressions.
-        result = self.result = MinimizerResult()
+        self.result = MinimizerResult()
+        result = self.result
         if params is not None:
             self.params = params
         if isinstance(self.params, Parameters):
@@ -396,33 +410,7 @@ class Minimizer(object):
         """
         pass
 
-    @deprecate(message='    Deprecated in lmfit 0.8.2, use scalar_minimize '
-                       'and method=\'L-BFGS-B\' instead')
-    def lbfgsb(self, **kws):
-        """
-        Use l-bfgs-b minimization
 
-        Parameters
-        ----------
-        kws : dict
-            Minimizer options to pass to the
-            scipy.optimize.lbfgsb.fmin_l_bfgs_b function.
-
-        """
-        raise NotImplementedError("use scalar_minimize(method='L-BFGS-B')")
-
-    @deprecate(message='    Deprecated in lmfit 0.8.2, use scalar_minimize '
-                       'and method=\'Nelder-Mead\' instead')
-    def fmin(self, **kws):
-        """
-        Use Nelder-Mead (simplex) minimization
-
-        Parameters
-        ----------
-        kws : dict
-            Minimizer options to pass to the scipy.optimize.fmin minimizer.
-        """
-        raise NotImplementedError("use scalar_minimize(method='Nelder-Mead')")
 
     def scalar_minimize(self, method='Nelder-Mead', params=None, **kws):
         """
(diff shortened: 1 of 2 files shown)


newville:
@Tillsten thanks for doing this!    Perhaps the tests just need to see if  least_squares is available?

I wonder if `apply_scaling` (default `True`) should be called something like `apply_bounds` or something else that means "have lmfit rescale the problem to be bounded".  

There are other solvers that also take bounds too.  Perhaps this change should allow user-selection of whether lmfit should apply the bounds to an unconstrained solver, or whether the bounds should be handed over to the external, bounded solver?

newville:
@Tillsten   I think you want to add a check for lmfit.minimizer.HAS_NEW_LEAST_SQUARES_MIN to test_least_squares(), and raise nose.SkipTest() if it's not available.

Also `HAS_NEW_LEAST_SQUARES_MIN` is sort of long, and it wont' be "new" for long.  Can this be
something like `HAS_SCIPY017` or `HAS_LEAST_SQUARES`?  It's sort of unfortunate to have both least_squares() and leastsq(), but we're not going to change that.

lmfit/minimizer.py
```diff
@@ -887,6 +875,61 @@ def emcee(self, params=None, steps=1000, nwalkers=100, burn=0, thin=1,
...
         return result
 
+    def least_squares(self, params=None, **kws):
+        """
+        Use the least_squares (new in scipy 0.17) function to perform a fit.
+        This assumes that ModelParameters have been stored, and a function to
+        minimize has been properly set up. This method doesn't support
+        parameter.constraints.
+
+        This wraps scipy.optimize.least_squares, which has in build support
```
andyfaff: inbuilt

lmfit/minimizer.py
```diff
@@ -887,6 +875,61 @@ def emcee(self, params=None, steps=1000, nwalkers=100, burn=0, thin=1,
 
         return result
 
+    def least_squares(self, params=None, **kws):
+        """
+        Use the least_squares (new in scipy 0.17) function to perform a fit.
+        This assumes that ModelParameters have been stored, and a function to
+        minimize has been properly set up. This method doesn't support
+        parameter.constraints.
```
andyfaff: parameter constraints

lmfit/minimizer.py
```diff
@@ -887,6 +875,61 @@ def emcee(self, params=None, steps=1000, nwalkers=100, burn=0, thin=1,
 
         return result
 
+    def least_squares(self, params=None, **kws):
+        """
+        Use the least_squares (new in scipy 0.17) function to perform a fit.
+        This assumes that ModelParameters have been stored, and a function to
```
andyfaff: shouldn't `ModelParameters` just be `Parameters`?

lmfit/minimizer.py
```diff
@@ -887,6 +875,61 @@ def emcee(self, params=None, steps=1000, nwalkers=100, burn=0, thin=1,
...
+        params : Parameters, optional
+           Parameters to use as starting points.
+        kws : dict, optional
+            Minimizer options to pass to scipy.optimize.least_squares.
+
+        """
+
+        if not HAS_NEW_LEAST_SQUARES_MIN:
+            raise NotImplementedError("Scipy with a Version higher than 0.17 "
+                                      "is need for this method.")
```
andyfaff: needed

andyfaff:
The extra method should be added to the documentation.

andyfaff:
Are there circumstances where it's better to use `leastsq` instead of `least_squares`? Perhaps we should be testing for the presence of `least_squares`, if it's present then automatically use that. If not, then use `leastsq`?
I'd like the error estimation/covariance matrices to be done with this PR if possible. They really belong together.

lmfit/minimizer.py
```diff
@@ -282,11 +290,16 @@ def __residual(self, fvars):
...
-        for name, val in zip(self.result.var_names, fvars):
-            params[name].value = params[name].from_internal(val)
+
+        if apply_scaling:
+            for name, val in zip(self.result.var_names, fvars):
+                params[name].value = params[name].from_internal(val)
+                params.update_constraints()
+        else:
+            for name, val in zip(self.result.var_names, fvars):
+                params[name].value = val
```
andyfaff: You still need to call `params.update_constraints`, even if `apply_scaling is False`. This method is used to make sure that all `Parameter.expr` are evaluated. It's worth adding a test to see if `least_squares` works with expressions.

lmfit/minimizer.py
```diff
@@ -887,6 +875,61 @@ def emcee(self, params=None, steps=1000, nwalkers=100, burn=0, thin=1,
...
+                                      "is need for this method.")
+
+        result = self.prepare_fit(params)
+
+        replace_none = lambda x, sign: sign*np.inf if x is None else x
+        upper_bounds = [replace_none(i.max, 1) for i in self.params.values()]
+        lower_bounds = [replace_none(i.min, -1) for i in self.params.values()]
+        start_vals = [i.value for i in self.params.values()]
+
+        ret = least_squares(self.__residual,
```
andyfaff: I'd be tempted to use `functools.partial`:

    resid = partial(self.__residual, apply_scaling=False)
    ret = least_squares(resid, ...)

lmfit/minimizer.py
```diff
@@ -270,7 +278,7 @@ def values(self):
 
         return dict([(name, p.value) for name, p in self.result.params.items()])
 
-    def __residual(self, fvars):
+    def __residual(self, fvars, apply_scaling=True):
         """
         Residual function used for least-squares fit.
         With the new, candidate values of fvars (the fitting variables), this
```
andyfaff: The docstring needs to be updated (it needed overhauling anyway):

    Parameters
    ----------------
    fvars : np.ndarray
        <what are fvars>
    apply_scaling : bool, optional
         <what is apply_scaling>

lmfit/minimizer.py
```diff
@@ -282,11 +290,16 @@ def __residual(self, fvars):
         if self._abort:
             return None
         params = self.result.params
-        for name, val in zip(self.result.var_names, fvars):
-            params[name].value = params[name].from_internal(val)
+
+        if apply_scaling:
+            for name, val in zip(self.result.var_names, fvars):
+                params[name].value = params[name].from_internal(val)
+                params.update_constraints()
```
andyfaff: Take `params.update_constraints` out of the `for` loop. It's an expensive call and it only needs to be called once in `__residual`.

lmfit/minimizer.py
```diff
@@ -282,11 +290,16 @@ def __residual(self, fvars):
...
+        else:
+            for name, val in zip(self.result.var_names, fvars):
+                params[name].value = val
         self.result.nfev += 1
 
-        params.update_constraints()
```
andyfaff: Given my comments above this line needs to be added back in - `update_constraints` needs to be called even if `apply_scaling is False`.

tests/test_least_squares.py
```diff
@@ -0,0 +1,54 @@
...
+
+    fit_params = Parameters()
+    fit_params.add('amp', value=13.0, max=20, min=0.0)
+    fit_params.add('period', value=2, max=10)
+    fit_params.add('shift', value=0.0, max=pi/2., min=-pi/2.)
+    fit_params.add('decay', value=0.02, max=0.10, min=0.00)
+
+    min = Minimizer(residual, fit_params, (x, data))
+    out = min.least_squares()
+
```
andyfaff: PEP8: one blank line too many

tests/test_least_squares.py
```diff
@@ -0,0 +1,54 @@
+from lmfit import Parameters, minimize, fit_report, Minimizer
+from lmfit_testutils import assert_paramval, assert_paramattr
+
+from numpy import linspace, zeros, sin, exp, random, pi, sign
+
```
andyfaff: Can we have a test for:

1. that fits with `Parameter.expr` still work?
2. to check that the output from `least_sq` is the same as `leastsq`, as far as we know it should be?
3. that error estimation is correct?

lmfit/minimizer.py
Tillsten: That moved into the loop just by accident.

lmfit/minimizer.py
Tillsten: I wasn't sure if update_constraints needs the bound transformation to work, so thats why i left it out, but i am happy to put it back in.

lmfit/minimizer.py
Tillsten: Yeah, this is just copied from the docstring of leastsq above.

tests/test_least_squares.py
Tillsten: Will do a test for Nr. 1. Nr. 3 is not yet done, so nothing to test yet. 
Can you specify what you mean by output? Because testing if the parameters
converge to same result is done indirectly: we compare with the true result.

lmfit/minimizer.py
Tillsten: My first temptation was to use a lambda function, but i didn't succumb :). I changed it to use kwargs now, which should make it the line clear enough.

Tillsten:
Ok, most comments are addressed. Still missing the expression test and error estimation.

> There are other solvers that also take bounds too.  Perhaps this change should allow user-selection of whether lmfit should apply the bounds to an unconstrained solver, or whether the bounds should be handed over to the external, bounded solver?

I agree that the current design is a little bit smelly. Maybe add two different penaIty functions? One for the scaled version, one non-scaled? Also i fully agree with that this should be generalized, but luckily for me this is outside the scope of this PR. :)

newville:
@Tillsten @andyfaff   This looks good to me.  Any objections to merging this?

newville:
@Tillsten @andyfaff  OK, merged!

Issue #339, opened 2016-06-16
Allow user to pass x-axis and y-axis plot labels in ModelResult.plot


ampolloreno:
I would like to be able to pass in optional parameters xlabel and ylabel to ModelResult.plot and them appear on the plot, as expressed below:

from lmfit import model
import numpy as np

def gaussian(x, amp, cen, wid):
    return amp * np.exp(-(x-cen)**2 /wid)

gmod = model.Model(gaussian)
params = gmod.make_params(cen=5, amp=200, wid=1)
x = np.linspace(0, 10, 201)
y = gmod.eval(x=x, amp=10, cen=6.2, wid=0.75)
result = gmod.fit(y, x=x, amp=5, cen=5, wid=1)
result.plot(xlabel="Frequency(Hz)", ylabel = "Amplitude")
plt.show()

newville:
I think this can be closed.  Please let me know if not.

PR #340, opened 2016-06-16
Added optional parameters xlabel and ylabel for plots

ampolloreno:
This PR addresses Issue #339, I added in optional xlabel and ylabel arguments for labeling plots.

### DIFF
--- a/lmfit/model.py
+++ b/lmfit/model.py
@@ -886,10 +886,10 @@ class ModelResult(Minimizer):
             ax.plot(x_array, self.data, datafmt, label='data', **data_kws)
 
         ax.set_title(self.model.name)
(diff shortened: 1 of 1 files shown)


newville:
@ampolloreno   Thanks, I have to admit I'm never excited to add features to the plotting methods, but this seems like a reasonable change to me.   Any objections to merging this?

Tillsten:
The check for `None` should be done via `is`: `if a is None`.

Just a comment: in general new keyword arguments should added to end for backwards
compatibility. I am inclined to ignore this here, because nobody should use this function via calling it with all arguments supplied in order.

newville:
Thanks -- any objections to merging this?

PR #342, opened 2016-06-27
Model eval scalar

newville:
This allows evaluating a Model and ModelResult at a scalar value, addressing #333

### DIFF
--- a/doc/model.rst
+++ b/doc/model.rst
@@ -667,11 +667,12 @@ including `chisqr`, `redchi`, `aic`, and `bic`.
 These methods are all inherited from :class:`Minimize` or from
 :class:`Model`.
 
-.. method:: ModelResult.eval(**kwargs)
+.. method:: ModelResult.eval(params=None, **kwargs)
 
-   evaluate the model using the best-fit parameters and supplied
-   independent variables.  The ``**kwargs`` arguments can be used to update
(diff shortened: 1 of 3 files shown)


newville:
any objections to this?

andyfaff:
I'm not really familiar with Model.

Tillsten:
I don't fully get why the result gets tiled to the length of the indepentent varible, so if you have time you could clarify the comment i would be grateful. But it seems to fix the issue mentioned in 333. Small nitpicks: docstrings don't follow numpydoc (only small changes needed) and this could have test.

newville:
@Tillsten  As far as I can tell, the use of `np.tile()` was solely to make a test pass for a Model that took an array and returns a constant value.   The test `test_model.test_result_custom_x()` was testing that if the length of the independent variable changed from the initial supplied independent model, that the value returned by `ModelResult.eval()` would match the length of the new independent variable.

I think that's a bogus test: for a model that returns a constant, the result shouldn't match the independent variable.  

This check with `np.tile()` was the root cause of #333.   So I think the better solution is to remove all these checks and re-tiling from `Model.eval()`, leaving just the evaluation itself, and change that bogus test.    If a model is supposed to return a constant, that's OK.  

I changed the PR to reflect this.

newville:
OK, merging this.

PR #344, opened 2016-07-08
DOC: fix module names

andyfaff:
I've fixed some of the sphinx module names. This means that they will be linkable from other documentation sites.

### DIFF
--- a/doc/fitting.rst
+++ b/doc/fitting.rst
@@ -23,6 +23,8 @@ function that calculates the array to be minimized), a :class:`Parameters`
 object, and several optional arguments.  See :ref:`fit-func-label` for
 details on writing the objective.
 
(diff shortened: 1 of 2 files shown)


newville:
@andyfaff great -- I didn't know about `currentmodule`.  Does this help intersphinx?

andyfaff:
Yes, it helps intersphinx.

andyfaff:
For future reference you can use the following code segment to extract the inventory dictionary from the objects.inv file created by sphinx. You can use the entries in this to determine how to refer to them from other sphinx documentation projects.

    from sphinx.ext.intersphinx import fetch_inventory
    import warnings
    uri = ''
    inv = fetch_inventory(warnings, uri, uri + 'objects.inv')

PR #345, opened 2016-07-08
Fix log-likelihood computation

tritemio:
Fix the factor -2*log-likelihood used to compute AIC and BIC statistics.


newville:
@tritemio Thanks!!

PR #346, opened 2016-07-10
Docs: use python3 print

tritemio:
Add parenthesis to print statements in the docs and fix one typo.


newville:
@tritemio Thanks!

Issue #349, opened 2016-07-10
Error with composite models in ipython


tritemio:
Executing the folllowing script in an ipython kernel (either from the terminal or the notebook) gives and error:

```
import lmfit
print('lmfit: %s' % lmfit.__version__)

peak1 = lmfit.models.GaussianModel(prefix='p1_')
peak2 = lmfit.models.GaussianModel(prefix='p2_')
model = peak1 + peak2

model.make_params()
```

Error:
```
lmfit: 0.9.4rc1-7-gae7884b
ZeroDivisionError
 <_ast.Module object at 0x1198cb9e8>
 ^^^
float division by zero
---------------------------------------------------------------------------
ZeroDivisionError Traceback (most recent call last)
/Users/anto/miniconda3/lib/python3.5/site-packages/IPython/core/formatters.py in __call__(self, obj)
 668 type_pprinters=self.type_printers,
 669 deferred_pprinters=self.deferred_printers)
--> 670 printer.pretty(obj)
 671 printer.flush()
 672 return stream.getvalue()
/Users/anto/miniconda3/lib/python3.5/site-packages/IPython/lib/pretty.py in pretty(self, obj)
 366 if cls in self.type_pprinters:
 367 # printer registered in self.type_pprinters
--> 368 return self.type_pprinters[cls](obj, self, cycle)
 369 else:
 370 # deferred printer
/Users/anto/miniconda3/lib/python3.5/site-packages/IPython/lib/pretty.py in _ordereddict_pprint(obj, p, cycle)
 817 p.text('...')
 818 elif len(obj):
--> 819 p.pretty(list(obj.items()))
 820 
 821 def _deque_pprint(obj, p, cycle):
/Users/anto/miniconda3/lib/python3.5/site-packages/IPython/lib/pretty.py in pretty(self, obj)
 366 if cls in self.type_pprinters:
 367 # printer registered in self.type_pprinters
--> 368 return self.type_pprinters[cls](obj, self, cycle)
 369 else:
 370 # deferred printer
/Users/anto/miniconda3/lib/python3.5/site-packages/IPython/lib/pretty.py in inner(obj, p, cycle)
 550 p.text(',')
 551 p.breakable()
--> 552 p.pretty(x)
 553 if len(obj) == 1 and type(obj) is tuple:
 554 # Special case for 1-item tuples.
/Users/anto/miniconda3/lib/python3.5/site-packages/IPython/lib/pretty.py in pretty(self, obj)
[...]

PR #351, opened 2016-07-11
Param defvals

newville:
this should address #349 
It does a better job of setting default parameter values.

It also adds a guard for divide-by-zero in the height calc for the peak-like models.


newville:
any objections to this?

PR #352, opened 2016-07-15
STYPEP8: inconsequential PEP8 and style changes (built in packages co…

andyfaff:
Various pep8 and sty changes to parameters.py, no changed behaviour. In addition I moved around the imports at the top. scipy style is to have batteries included modules (e.g. copy) imported first.


newville:
@andyfaff  thanks!

PR #353, opened 2016-07-15
Prep094rc2

newville:
This has many doc tweaks for 0.9.4rc2 .  I think this is very close to being ready for 0.9.4.   Once 0.9.4 is released, I think we (I) should make a concerted effort to uniformly use napoleon-style docstrings and use autodoc throughout the code and docs.

Also adds a check with `isnan` or `isfinite` before calling _nan_policy(), so that it is only called when needed.


lmfit/minimizer.py
```diff
@@ -1258,7 +1260,8 @@ def _lnpost(theta, userfcn, params, var_names, bounds, userargs=(),
 
     # now calculate the log-likelihood
     out = userfcn(params, *userargs, **userkwargs)
-    out = _nan_policy(out, nan_policy=nan_policy, handle_inf=False)
```
andyfaff: This comment applies here, and on L363. `_nan_policy` was designed in such a way to be really efficient. If out is a large array (it's not unusual to have many datapoints) then `np.isnan(out)` creates a large temporary `bool` array (with associated overhead). `_nan_policy` circumvents this by a reduction operation first, `np.sum`, and checking if that single value is finite.
I had thought of implementing the checks in this commit, but I think the code looks neater as it exists currently, and I don't think there is any real benefit (no extra overhead) to making these two changes.

newville:
@andyfaff  Ah sorry and thanks for the clarification.  I was playing around with the asv benchmarking, and may have misinterpreted the results.   

I believe (please correct me if I'm wrong) that if I commit without these checks of `isnan()/isnfinite()` and run the benchmarks on this branch that we'll have the measurements.  

Perhaps the benchmarks should include both "smallish arrays" (say, < 5,000 element which I think this is "normal use case" for many of us and so important to not optimize against) and large arrays (say > 1,000,000 elements which might be "uncommon but not unheard of").      Suggestions welcome!

FWIW, I'll be on travel much of the week but am hoping to get 0.9.4 out by the end of the month.

newville:
@andyfaff  What I see with asv benchmarking is that using the checks with `isnan()/isfinite()` is a noticeable improvement for the "small array with no nans" case (your benchmark of 601 array elements).  For other tests with slightly larger arrays or arrays that actually contain nans, it's hard to see any difference.  But I think the "small array, no nans" is an important case to not not make worse.

Some of these results are at http://millenia.cars.aps.anl.gov/software/python/lmfit/benchmarks/

Any thoughts?

andyfaff:
I'm a little busy today, but I'll get to looking at the benchmarks soon.

newville:
No problem. I'm traveling the first half of this week. I'd like to push 0.9.4 sometime next week.

andyfaff:
I noticed from some of the rst document files that you've included some `..currentmodule` directives. Can you use the following code to check that the generated documentation has consistent intersphinx links:

    from sphinx.ext.intersphinx import fetch_inventory
    import warnings
    uri = ''
    inv = fetch_inventory(warnings, uri, uri + 'objects.inv')
    for key in inv.keys():
        for key2 in inv[key]:
            print(key2)

Currently the intersphinx links aren't consistent, and they need to be for third party sphinx generated websites to refer to them.

newville:
OK, added fully qualified module names and using `module` except for the print functions.    This seems like enough for 0.9.4, which I'd like to push out, and we can leave more doc fixes for 0.9.5.

Issue #354, opened 2016-07-15
use napoleon-style docstrings uniformly, use autodoc tools in documentation


newville:
We should move to using autodoc tools and napoleon-style docstrings uniformly.

andyfaff:
What are napoleon-style docstrings?

tritemio:
I also like the napoleon docstring style and use it in my projects. However for lmfit, I think it is more appropriate to stick with the numpy style. Most of the docstring are already in this style anyway.

We also have to decide on a package-level syntax for the `currentmodule` directive. Normally I use the directive as:

```
..currentmodule:: lmfit.minimizer
```

and create a link to a function using:

```
:func:`minimize`
```

If we change the directive to 

```
..currentmodule:: minimizer
```

the link is not created anymore and I don't know if there is an alternative syntax that works in this case.

@andyfaff mentioned that the former syntax result in intersphinx links as:

```
lmfit::lmfit.minimize
```

while with the latter, the links would become a shorter version:

```
lmfit::minimize
```

newville:
@andyfaff https://sphinxcontrib-napoleon.readthedocs.io/en/latest/   These are much better to read as plain text and much easier to write.   

Personally, I don't see a lot of value in following numpy's convention. The docstrings should be as clear and complete as possible, should be readable as plain text, and work well with sphinx's autodoc.    I think the napoleon style wins for this.   Fortunately, changing numpy doc style to napoleon is mostly removing no-longer-needed markup.

The main issue I have with `currrentmodule` is that for me (using sphinx 1.4.1) it does not fully work.  Using (in paramters.rst):

``` 
.. currentmodule:: parameter
```

fails due to relative imports:

```
/Users/Newville/Codes/lmfit-py/doc/parameters.rst:177: WARNING: autodoc: failed to import method u'Parameters.pretty_print' from module u'parameter'; the following exception was raised:
Traceback (most recent call last):
  File "/Users/Newville/anaconda/lib/python2.7/site-packages/Sphinx-1.4.1-py2.7.egg/sphinx/ext/autodoc.py", line 507, in import_object
    __import__(self.modname)
  File "/Users/Newville/Codes/lmfit-py/lmfit/parameter.py", line 13, in <module>
    from . import uncertainties
ValueError: Attempted relative import in non-package
```

Perhaps this is sphinx 1.4.1 ?  OTOH, using the full name everywhere:
 
```
.. currentmodule:: lmfit.parameter
[...]

Issue #332, opened 2016-05-05
maxfev and maxiter should be supported for all methods, and uniformly


newville:
Currently, one can set `maxfev` for the leastsq method with

``` python
result = minimize(fcn2min, params, args=(x, data), maxfev=20)
```

but when using a scalar minimizer, one has to do

``` python
mini = Minimizer(fcn2min, params, fcn_args=(x, data))
result = mini.scalar_minimize(method='Nelder-Mead', options={'maxiter': 150})
```

which is inconsistent and should be made more consistent. 

My preference would be to support either `maxfev` or `maxiter` as equivalent, and support this for all methods in `minimize()`.

I believe there may be similar issues with `tol` and `xtol` (which most scalar minimizers support).

tritemio:
+1

newville:
@uvainio Please do not hijack github Issues with questions about lmfit.   We have a mailing list for support and use github Issues for actual issues with the code and docs.   This is described many places in the docs and github pages.

It is possible to specify `maxfev` for Model.fit, as with
```
my_model.fit(ydata, params, xdata=x, fit_kws={'maxfev': 200})
```

uvainio:
Hi Matt,

Many thanks for the very informative answer, and sorry for misusing guthub. Again one step closer to learning how to use it correclty!

luriomer:
Are there any news about this issue? I highly support this, since I have wrappers around LMFIT and this can easily simplify things.

newville:
@luriomer  I guess this sort of got lost.  Just to be clear, the inconsistency is in `scipy`, and the question here is how much we try to "smooth out" the inconsistencies and poor API choices in scipy versus being a "thin wrapper".   <insert rant about scipy.optimize inconsistencies and poor API here>.

Anyway, I  guess I would suggest we standardize on "maxfev" as "maximum number of function evaluations", partly because that is easy to count and partly because "iteration" is not a well-defined (or consistently defined) concept.

PR welcome for that....

luriomer:
Yes, I understand. I have to do it on my end anyway so I hope I'll have also time to file a PR soon.

Best

luriomer:
Correct me if I'm wrong, but this issue is a bit bigger than `maxfev` vs. `maxiter`. As I understand, `Minimizer.minimize(method="leastsq",params=...)` and `Minimizer.leastsq(params=...)` do the same thing, which makes it hard to correctly wrap all of the desired parameters in a consistent way. 
[...]

PR #355, opened 2016-07-16
Make MinimizerResult API more similar to ModelResult

tritemio:
In the process of adding support for MinimizerResult in [pybroom](https://github.com/tritemio/pybroom) I found some little inconsistencies between MinimizerResult and ModelResult in lmfit.

This PR makes the interface of the two classes more similar by adding to MinimizerResults two attributes: the dict `init_values` and string `method`.

### DIFF
--- a/doc/fitting.rst
+++ b/doc/fitting.rst
@@ -458,25 +458,7 @@ For full control of the fitting process, you'll want to create a
 
 The Minimizer object has a few public methods:
 
-.. method:: minimize(method='leastsq', params=None, **kws)
-
-   perform fit using either :meth:`leastsq` or :meth:`scalar_minimize`.
-
-   :param method: name of fitting method.  Must be one of the names in
-		  :ref:`Table of Supported Fitting Methods <fit-methods-table>`
-   :type  method:  str.
-   :param params:  a :class:`Parameters` dictionary for starting values
-   :type  params:  :class:`Parameters` or `None`
-
-   :return: :class:`MinimizerResult` object, containing updated
-	    parameters, fitting statistics, and information.
-
-.. versionchanged:: 0.9.0
-   return value changed to :class:`MinimizerResult`
-
-   Additional keywords are passed on to the correspond :meth:`leastsq`
-   or :meth:`scalar_minimize` method.
-
+.. automethod:: Minimizer.minimize
 
 .. method:: leastsq(params=None, scale_covar=True, **kws)
 
--- a/doc/fitting.rst
+++ b/doc/fitting.rst
@@ -15,7 +15,9 @@ when using complicated constraints or comparing results from related fits.
 
 
 The :func:`minimize` function
-===============================
+=============================
+
+.. currentmodule:: lmfit.minimizer
 
 The :func:`minimize` function is a wrapper around :class:`Minimizer` for
 running an optimization problem.  It takes an objective function (the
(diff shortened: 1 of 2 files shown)


newville:
@tritemio Thanks -- I agree with much of this:

* support 'least_squares'
* add `method` name and `init_values` to `MinimizerResult`

but: 
 * dict comprehension is not supported in Py 2.6 and so cannot be used until lmfit 0.9.5.
 * add `least_squares` to docstring for `minimize()` function.
 * I'd rather not change the call signature to `prepare_fit()` (especially to change `params` from optional keyword to required positional argument).  If `method` is added to the call signature, it should be a keyword argument.  But, I'd rather *not* set the method name in `prepare_fit`, as it sort of looks like one would expect that to dictate which method was going to be used, and that is not the case.   So, better (for now, at least) to set after the `MinimizerResult` is created, and possibly even after the fit is completed.

doc/fitting.rst
```diff
@@ -15,59 +15,17 @@ when using complicated constraints or comparing results from related fits.
 
 
 The :func:`minimize` function
-===============================
+=============================
+
+.. currentmodule:: lmfit.minimizer
```
andyfaff: please leave as:

    ..currentmodule:: minimizer

This makes the intersphinx references uniform across parameters.rst and model.rst. For example, the intersphinx reference for `Parameters` is currently `lmfit::Parameters` and `minimize` is `lmfit::minimize`.
If this change is kept then the intersphinx reference becomes `lmfit::lmfit.minimize`.

lmfit/minimizer.py
```diff
@@ -1369,40 +1377,65 @@ def _nan_policy(a, nan_policy='raise', handle_inf=True):
...
-    The minimize function takes a objective function to be minimized, a
-    dictionary (lmfit.parameter.Parameters) containing the model parameters,
-    and several optional arguments.
+    This function performs a fit of a set of parameters by minimizing
+    an objective (or "cost") function using one one of the several
+    available methods. The minimize function takes a objective function
+    to be minimized, a dictionary (:class:`lmfit.parameter.Parameters`)
+    containing the model parameters, and several optional arguments.
+
+    Notes
```
andyfaff: The notes section should follow the `Parameters` and `Returns` sections, it shouldn't be at the top.

lmfit/minimizer.py
```diff
@@ -1369,40 +1377,65 @@ def _nan_policy(a, nan_policy='raise', handle_inf=True):
...
     Parameters
     ----------
     fcn : callable
-        objective function that returns the residual (difference between
-        model and data) to be minimized in a least squares sense.  The
-        function must have the signature:
-        `fcn(params, *args, **kws)`
-    params : lmfit.parameter.Parameters object.
+        objective function to be minimized. When method is `leastsq` or
+        `least_squares`, the objective function need to return an array
```
andyfaff: needs

lmfit/minimizer.py
```diff
@@ -1369,40 +1377,65 @@ def _nan_policy(a, nan_policy='raise', handle_inf=True):
...
+        of residuals (difference between model and data) to be minimized
+        in a least squares sense.  With other scalar methods the objective
```
andyfaff: `Minimizer.penalty` allows a scalar value *or* the array of residuals to be returned. It doesn't need to be a scalar.

    `With the scalar methods the objective function can either return the residuals array or a single scalar value`

lmfit/minimizer.py
```diff
@@ -1415,20 +1448,28 @@ def minimize(fcn, params, method='leastsq', args=None, kws=None,
     fit_kws : dict, optional
         Options to pass to the minimizer being used.
 
-    Notes
```
andyfaff: Why were the notes placed at the top of the docstring? When using interactive help people want to see Parameters first, then Returns, then Notes.

lmfit/minimizer.py
```diff
@@ -1415,20 +1448,28 @@ def minimize(fcn, params, method='leastsq', args=None, kws=None,
...
+    :ref:`fit-results-label` for further details.
+
+    This function is simply a wrapper around :class:`Minimizer`
+    and is equivalent to::
+
+        fitter = Minimizer(fcn, params, fcn_args=args, fcn_kws=kws,
+        	           iter_cb=iter_cb, scale_covar=scale_covar, **fit_kws)
+        fitter.minimize(method=method)
 
-    A common use for `args` and `kwds` would be to pass in other
```
andyfaff: Why are these lines removed?

doc/fitting.rst
andyfaff: If we do make the change, then all the module references in the other documentation files should be changed to match this.

lmfit/minimizer.py
tritemio: ok

lmfit/minimizer.py
tritemio: thanks

lmfit/minimizer.py
tritemio: done

lmfit/minimizer.py
tritemio: ok

lmfit/minimizer.py
tritemio: They are in the notes section

doc/fitting.rst
tritemio: With 

```
..currentmodule:: lmfit.minimizer
```

sphinx creates a link to a function using the intuitive syntax:

```
:func:`minimize`
```

When I change the directive to 

```
..currentmodule:: minimizer
```

the link is not created anymore. Is there a different syntax in this case?

In general, I would favor cleaner/shorter internal links over shorter intersphinx links.

tritemio:
I fixed the 3 issues raised by @newville 

Moreover, I added all the changes suggested by @andyfaff except for the currentmodule directive which breaks internal links (see my inline comment).

Since this PR also modifies `Minimizer.minimize` and `MinimizerResult`, I also expanded  the use autodoc to these two items (a class and a method) in the process of documenting new attributes and methods.

Other than that, I think we should address the decision on the currentmodule syntax and the completion of autodoc use (the entire Minimizer class should be autodoc-ed) to another PR and preferably continue the discussion in #354.

So, if there are no more outstanding issues related to the topic of this PR I would suggest merging it and moving further general docs discussion to #354 to be addressed in a new PR.

doc/fitting.rst
andyfaff: > In general, I would favor cleaner/shorter internal links over shorter
intersphinx links

OK.  One of the problems with lmfit documentation in general is that the
external intersphinx links aren't consistent. E.g. one might have to use
`lmfit::Parameters` but `lmfit.Minimizer.minimize`. It's impossible to
predict which one. One has to inspect the objects.inv file to see which one
to use.

tritemio: agree, see my comment in https://github.com/lmfit/lmfit-py/issues/354

tritemio:
@newville, I'm surprised and disappointed you marked the 0.9.4 release without including this PR. Any reasons? I need this PR to simplify the job of pybroom and I wanted to support 0.9.4+, but I can't now.

newville:
@tritemio Oh, very sorry. My mistake -- somehow in the discussion of sphinx markup I completely missed that you had fixed this.    So the only reason is my haste and carelessness.

That said, would merging this and releasing 0.9.5 now be sufficient or did you need this and pybroom to work with Python2.6?   I have no objection to doing that.   There are other cleanups of code and docs (#332 and #354)  I'd like to see, but I won't have time to work on these for a while.

newville:
@tritemio  merged this branch (by hand) to master.

tritemio:
@newville, no problem, mistakes happens.

Yes a 0.9.5 release would fix any issue for me, I don't plan to support py2 in pybroom anyway.

Let me check what you merged, I may have some more docs fixes.

newville:
@tritemio  Thanks for understanding, and apologies again.  I'll close this one, and we can continue from here.   Shall we aim for 0.9.5 release with your fixes on Tuesday?

PR #362, opened 2016-07-26
Prep095

newville:
this fixes a number of small needs for version 0.9.5 :  updated version requirements, warning messages, and docs.    I believe this is ready to be tagged as 0.9.5.


newville:
I think this is all uncontroversial (except perhaps the need for such a short time between last release).

PR #363, opened 2016-07-27
Autodoc Minimizer class

tritemio:
I enabled autodoc for the Minimizer class.

I basically removed the manual documentation copying in the docstring the things that were missing.

I also fixed some formattings, and math expression (\n, \r, \f need to be escaped in latex equations).

Builds correctly locally.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -31,6 +31,8 @@ extensions = [
     'sphinx.ext.mathjax',
     ]
 
+autoclass_content = 'both'
+
 try:
     import IPython.sphinxext.ipython_directive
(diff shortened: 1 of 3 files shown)


tritemio:
Also, I added `autoclass_content = 'both'` in conf.py which makes sphinx concatenate class-level and __init__ doctrings in a single section (otherwise the __init__ docstring is not shown). You can see the result in the Minimizer class documentation.

Ref: http://www.sphinx-doc.org/en/stable/ext/autodoc.html

newville:
Great!

PR #364, opened 2016-07-29
Fix ignored kws in Minimizer.minimize

tritemio:
Keyword arguments (to be passed to scipy.optimize functions) can be given during initializations of Minimizer object. However, Minimizer.minimize does not relays these arguments.

This also causes that the dict `fit_kws` passed to `Model.fit` was ignored.


newville:
@tritemio Thanks for catching those!

Issue #365, opened 2016-08-10
Minimizer.scalar_minimize is squaring scalar objective functions


caldwellshane:
For previous discussion see:
https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/lmfit-py/Q0TmAtKtYG8/0YaPhxupAAAJ

newville:
Should be fixed with #366

PR #366, opened 2016-08-10
Issue365

caldwellshane:
This PR addresses Issue #365. Includes one new unit test to check that `Minimize.scalar_minimize()` finds the correct minimum of an objective function that has a negative optimal value.


newville:
@caldwellshane  Thanks!  This looks good to me.  Any objections to merging this?

PR #367, opened 2016-08-14
Model uncertainties

newville:
This adds a method `eval_conf_band()` to ModelResult that attempts to calculate the n-sigma uncertainty in the model given the uncertainties and covariance matrix of the Parameters.  In that sense, it gives "confidence bands" on the best-fit model.   This is related to the question on the mailing list at https://groups.google.com/forum/#!topic/lmfit-py/r8y7VbTa3kU

The calculation is done using the method outlined at
    https://www.astro.rug.nl/software/kapteyn/kmpfittutorial.html#confidence-and-prediction-intervals

which references the original work of J. Wolberg,Data Analysis Using the Method of Least Squares, 2006, Springer.   The derivatives (d*model*/d*par*) are done explicitly here, and the results combined.  I believe this is done correctly, but it would be helpful to come up with a case where the answer is well understood (perhaps even analytic).

A docstring is included, as is a section in the doc for the Model and ModelResult classes.  An example is provided, with results shown in the doc. 

A unit-test has not yet been added.   Any thoughts on how to test this well?

In writing this, I noticed that the conf_interval() code uses confusing (to me_ notation for "sigma" that is different than what is used in eval_conf_band().  This should probably be rationalized so the APIs are more alike than different.

Comments / Suggestions welcome, and much appreciated.


newville:
OK, updated to add tests, checking the results with trivial linear fits.  Also, changed name from the awkward `ModelResult.eval_conf_band()` to `ModelResult.eval_uncertainty()`,  though suggestions for better name (and/or better tests!) most welcome.

danielballan:
Very nice! Two simple sanity tests would be the following:

1. For some data set, verify that the N+1 sigma band > N sigma band.
2. Verify that the N sigma uncertainty decreases monotonically as added noise decreases.

Not foolproof, but more rigorous that a simple smoke test and tolerant to minor fluctuations in the exact result.

newville:
@danielballan   Good ideas for tests -- I'll add those.    And, yeah, I think this is very useful!

The current tests are fairly trivial, but aim to demonstrate that the estimated uncertainty is the right value to at least 2 digits:  a non-linear fit to a line + known_noise, with only one true variable.  For each of these cases, I know what the uncertainty level in the model should be.    

Making plots of these uncertainties, it's often surprising how small the 1-sigma variations are.   So the test with line+noise at least gives me confidence that there isn't a serious math error (and given the trick of re-scaling the covariance by reduced chi-square, that was a real concern)!

andyfaff:
Coming a bit late to the party here. I think this is exactly what emcee can
be used for.
A possible test would be to check against an emcee posterior sampling. One
will get all the quantiles you need.

On 19 Aug 2016 5:40 am, "Matt Newville" <notifications@github.com> wrote:

@danielballan <https://github.com/danielballan> Good ideas for tests --
I'll add those. And, yeah, I think this is very useful!

The current tests are fairly trivial, but aim to demonstrate that the
estimated uncertainty is the right value to at least 2 digits: a non-linear
fit to a line + known_noise, with only one true variable. For each of these
cases, I know what the uncertainty level in the model should be.

Making plots of these uncertainties, it's often surprising how small the
1-sigma variations are. So the test with line+noise at least gives me
confidence that there isn't a serious math error (and given the trick of
re-scaling the covariance by reduced chi-square, that was a real concern)!

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/lmfit/lmfit-py/pull/367#issuecomment-240833501>, or mute
the thread

newville:
@andyfaff  Oh, that's an interesting idea.  I'm afraid I don't really know how to pull such information out of the emcee results.  Can you provide an example?

andyfaff:
Sorry, I don't have a ready made example to hand. THe flow would go
something like this:

1) `lmfit.Minimizer` is constructed.
2) run `Minimizer.emcee` to collect samples. Remember to burn and thin
properly.
3) Once you've run this the `MinimizerResult` has the `flatchain` or
`chain` attributes. These attributes contain the samples of the posterior
probability distribution.
4) You can work out statistics from the MCMC samples in `chain`, this is
how `MinimizerResult.params` is calculated. For a given parameter you have
thousands of samples. You just have to work out the quantiles, median,
mean, etc, to give the parameter value and its uncertainty.
5) But we want to work out confidence bands on the fit. So in this case you
have to take the samples in `chain` and calculate the generative `Model`
for each of those samples. You then end up with thousands of fits
corresponding to the thousands of samples.
6) For each datapoint in the fit you can work out required the required
quantiles using `np.percentile`. I think these are the equivalent to the
confidence band.
(7) Plot the confidence band using something along the lines of
http://ogrisel.github.io/scikit-learn.org/sklearn-tutorial/auto_examples/gaussian_process/plot_gp_regression.html

The majority of time is spent in the sampling in step (2).

For a model that is quick to calculate (pretty much all the `Models` are
fast) you could be able to do the required sampling in a fast enough time
to include in the test suite. I'm guessing on the order of <10 s. Indeed,
emcee is currently tested to check that parameter uncertainties are similar
to those from Levenberg Marquardt.

newville:
Two sanity tests added.

@andyfaff  That sounds like a good study to do, and I'll put it on my todo list, but it's a bit more than I'm able to do for unit tests at the moment. 

FWIW, I found `fill_between` to be a great way to show the uncertainty in the model.  Such a plot is in the doc section of the PR.

newville:
OK, I think this is ready to merge.

danielballan:
:+1:

PR #370, opened 2016-08-20
Plots: make a better default plot with ModelResult.plot()

tritemio:
As the subject says, small tweaks to create better default plots with `ModelResult.plot()`.

Changes:

- use `loc='best'` for legend positioning in `plot_fit` and `plot_residuals`
- remove title for bottom axis which is the same of the top axis and can overlap with other labels
- remove x-labels for top axis (we are sharing the x axis so they are redundant)
- return both figure and GridSpec. This is a convenience for easy tweaking the plot. As a side benefit, with this change calling `ModelResult.plot()` without capturing the return object will not display two duplicated figures anymore.
- Use a square figure by default using the default x-size of the figure also as y-size. Since we stack two plots vertically a square figure results in more legible plots in most cases. By using the default x-size we don't set an arbitrary size and respect user preferences.
- Fix typo in docstring.

### DIFF
--- a/lmfit/model.py
+++ b/lmfit/model.py
@@ -914,8 +914,7 @@ class ModelResult(Minimizer):
         if ylabel is None:
             ax.set_ylabel('y')
         else: ax.set_ylabel(ylabel)
-        ax.legend()
-
+        ax.legend(loc='best')
         return ax
 
     @_ensureMatplotlib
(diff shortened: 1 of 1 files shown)


newville:
@tritemio Thanks -- that looks good to me.   Any objections?

PR #374, opened 2016-10-27
Custom cost, replaces #226

ggggggggg:
I took the code from #226, I updated it to be based on master, and changed the API somewhat. Now you can pass `reduce2scalar` as an argument to either `minimize`, `Minimizer`, or `Model`. I added one test. I don't really understand your test framework, I'm pretty sure there is a way to drop this in to effectively re-use all your tests with very little code, but I don't know how to do it.

I updated the doc strings of everything I touched.

### DIFF
--- a/examples/example_stable.py
+++ b/examples/example_stable.py
@@ -1,0 +1,50 @@
+# -*- coding: utf-8 -*-
+"""
+Example using the t-likelihood for stable fitting.
+"""
+import numpy as np
+import scipy.special as sp
+import scipy.stats as stats
+import lmfit
+try:
+    import matplotlib.pyplot as plt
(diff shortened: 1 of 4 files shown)


ggggggggg:
Hmm. I realized this isn't general enough to address my use case, I need to make this work when calculating the residuals because I need access to model values as well.

newville:
@ggggggggg  (is that a real name?  We much prefer working with actual people.)  

I believe this should be doable with no changes to `Model` at all and be handled completely by specifying arguments in `**fit_kws`.  One would have to specify a scalar minimizer too.

Also, the numerous changes to empty lines is distracting.  Please undo that.

newville:
replaced by #387

PR #376, opened 2016-11-22
a few small changes to models.py and model.py

newville:
this PR contains a few small, unrelated changes to Model and Models over the past couple weeks.  

A.  Add a damped-harmonic-oscillator model based on a model from PAN (In IDL, from NIST) used by some collaborators.

B. Increase the default estimate of the amplitude of a peak with `Model.guess`

C. Improve the `ensureMatplotlib` decorator to reveal the docstring and name of the decorated method.

D.  Allow Model to wrap certain function-like objects with `argnames` and `kwargs` arguments.  This is mostly for my X-ray spectroscopy framework xraylarch, which does generate such things (cause, you know, lmfit/python by themselves aren't slow enough!).

These should each have no actual impact on any existing tests or usage.

### DIFF
--- a/lmfit/lineshapes.py
+++ b/lmfit/lineshapes.py
@@ -15,7 +15,7 @@ spi  = sqrt(pi)
 s2   = sqrt(2.0)
 
 functions = ('gaussian', 'lorentzian', 'voigt', 'pvoigt', 'moffat', 'pearson7',
-             'breit_wigner', 'damped_oscillator', 'logistic', 'lognormal',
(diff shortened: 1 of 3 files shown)


Tillsten:
*   The change of the guess could break someone code, oth the guessing is so primitive that nobody should depend on it. 
* The decorator improvements are easier done by applying functools.wraps

newville:
@Tillsten Thanks, I updated to use functools.wraps. 

 I think we never guaranteed the stability (or accuracy) of the automated guesses.

newville:
@tillsten @andyfaff  any objections to merging this?

PR #377, opened 2016-11-24
Normalized uniform prior

abeelen:
This PR add a proper normalization of the default uniform prior for the PTSampler, as well as adding the PTSampler betas option.

This allows to run the emcee multinodal.py example using lmfit.Parameters and get proper evidence for the model.

In general this allow to get the evidence of the model using the PTSamper.sampler.thermodynamic_integration_log_evidence() method

### DIFF
--- a/lmfit/minimizer.py
+++ b/lmfit/minimizer.py
@@ -1320,10 +1320,14 @@ def _lnprior(theta, bounds):
     """
     if (np.any(theta > bounds[:, 1])
         or np.any(theta < bounds[:, 0])):
-        return -np.inf
+        logp = -np.inf
     else:
-        return -np.sum(np.log(bounds.dot([ -1, 1 ])))
+        logp = -np.sum(np.log(bounds.dot([ -1, 1 ])))
 
+        if not np.isfinite(logp):
+            logp = 0
+
+    return logp
 
 def _lnpost(theta, userfcn, params, var_names, bounds, userargs=(),
             userkws=None, float_behavior='posterior', is_weighted=True,
--- a/lmfit/minimizer.py
+++ b/lmfit/minimizer.py
@@ -636,7 +636,8 @@ class Minimizer(object):
 
     def emcee(self, params=None, steps=1000, nwalkers=100, burn=0, thin=1,
               ntemps=1, pos=None, reuse_sampler=False, workers=1,
-              float_behavior='posterior', is_weighted=True, seed=None):
+              float_behavior='posterior', is_weighted=True, seed=None,
+              betas=None):
         """
         Bayesian sampling of the posterior distribution using the `emcee`.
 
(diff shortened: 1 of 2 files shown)


newville:
@abeelen @andyfaff  would it make sense to add a `ptsampler_kwargs` argument that was a dict of arguments passed to `emcee.PTSampler`?    Or perhaps explicitly include all of its arguments as part of `emcee()`?

Does `emcee.EnsembleSampler` need similar treatment?]

Anyway, Thanks @abeelen !

andyfaff:
Thank you @abeleen.  The prior is unnormalised on purpose. If we do the normalisation for the user then that removes the possibility of users employing their own prior, e.g. gaussian. We make clear in the docstring that the prior is uniform, and unnormalized, but perhaps we can amend the documentation to show how to do prior normalisation within the objective function.
Do you have situations where the default betas don't work well?

abeelen:
Thanks for the replies !

For the normalization : on the master branch I only have in the docstring of ```minimizer.emcee``` : ```The method assumes that the prior is Uniform.``` which I wrongly assumed means that it is normalized. At the moment, it is a boundary check, without normalization.

It is true that ```float_behavior``` can be set to ```posterior``` and not likelihood, meaning that the prior are/should be taken into account, but this could be made clearer from the documentation which only refer to ```log-likelihood```. So I would agree that a better documentation as how to get proper normalized uniform prior and/or change the prior would be enough, although pushing the prior into the objective function would require extracting the ```var_arr``` and ```bounds``` at each call, and the check of the boundaries would/could be made twice, by ```_lnprior``` and the ```usrfcn``` posterior...

Another solution would be to expose the prior function to the ```emcee``` call, with the same signature as ```userfcn``` plus ```var_arr``` and ```bounds```. But I can already see the mess for fixed parameters... Adding a ```lnprior```attribute to ```lmfit.Parameters``` could also be a solution.

As for the ```betas``` ladder, it is mainly a question of convergence speed. See, for e.g.,  ```emcee/examples/multinodal.py``` which I am trying to reproduce with ```lmfit``` . Exposing all specific ```ptsampler_kwargs``` would be a good idea, I list only 2 options : 
* ```betas=None```
* ```Tmax=None```

leaving only ```a``` (an optional proposal scale factor) the only hidden option both for ```ptsampler``` and ```ensemblesampler```

I do not know if I can change this PR to include this, or if I should make a new PR.

abeelen:
After some initial checks, It is not possible to include the priors in the likelihood function as the PTSampler use : (Ligne 267 & 331  of ```emcee/ptsampler.py```) :

```
lnprob0 = logls * self.betas.reshape((self.ntemps, 1)) + logps

[...]

qslnprob = qslogls * self.betas.reshape((self.ntemps, 1)) + qslogps
```

Thus the priors are not scaled as the likelihoods. Thus the prior function should be properly set/normalized.

andyfaff:
Please see https://github.com/dfm/emcee/issues/205

andyfaff:
In general I think the `emcee` method is already quite complicated and one concern would be adding method parameters that would increase code complexity. At some point it would get to a stage where it's probably easier to use the `emcee` package directly.

> The method assumes that the prior is Uniform. which I wrongly assumed means that it is normalized.

This can be made more explicit in the documentation, probably in a separate PR.

andyfaff:
@abeelen, prior normalisation isn't going to occur within the emcee method, it's a design decision. If the user wishes to do this then they should include it in an objective function that returns the log-posterior. But we will welcome a PR that improves the documentation in this regard. As such I'm closing this PR.

Issue #378, opened 2016-11-25
result.message in minimizer.py


colgan:
Hi,
by using result.message I always got the message: 'Tolerance seems to be too small.', so I decided to have a look at it. The code snipped in your file which generates the variable result.message is:

        result.ier = ier
        result.lmdif_message = errmsg
        result.message = 'Fit succeeded.'
        result.success = ier in [1, 2, 3, 4]
        if result.aborted:
            result.message = 'Fit aborted by user callback.'
            result.success = False
        elif ier == 0:
            result.message = 'Invalid Input Parameters.'
        elif ier == 5:
            result.message = self._err_maxfev % lskws['maxfev']
        else:
            result.message = 'Tolerance seems to be too small.'

In my opinion this is never leading to result.message being: 'Fit succeeded.'
I think the snipped should look something like this:

        result.ier = ier
        result.lmdif_message = errmsg
        result.success = ier in [1, 2, 3, 4]
        if result.aborted:
            result.message = 'Fit aborted by user callback.'
            result.success = False
        elif ier == 0:
            result.message = 'Invalid Input Parameters.'
        elif ier == 5:
            result.message = self.err_maxfev % lskws['maxfev']
        elif result.success:
            result.message = 'Fit succeeded.'
        else:
            result.message = 'Tolerance seems to be too small.'

Kind regards
Colgan

newville:
@colgan  Thanks -- you're right that the messages are not accurately reflecting the value of `ier` from MINPACK.  In fact "success" is not correct either.  It should be:

0.  improper input parameters (more variables than data points, tolerance < 0.0, or no data)
1.  fit successful
2.  fit successful
3.  fit successful
4.  one or more variable did not affect the fit
5.  too many iterations
6.  tolerance too small.
7.  tolerance too small.

That is `result.success` should be `ier in (1, 2, 3)`.  

If you're willing to make a PR for this, that would be great.

colgan:
Done.
Kind regards
Colgan

PR #379, opened 2016-11-28
Different fit messages depending on fit status code.

colgan:
##     API changes or a new feature is added.

This PR addresses Issue #378 

# Change description
Instead of comparing ier only to 0, 5 or neither of them, which never resulted in result.message being "Fit succeeded." the changes lead to a comparison to five different cases, namely:
- ier is equal to 1, 2 or 3, which results in result.message = 'Fit succeeded.'
- ier is equal to 0, which results in result.message = 'Invalid Input Parameters. I.e. more variables than data points given, tolerance < 0.0, or no data provided.'
- ier is equal to 4, which results in result.message = 'One or more variable did not affect the fit.'
- ier is equal to 5, which results in result.message = self._err_maxfev % lskws['maxfev'] <-- as before
- ier is not any of the above, which results in result.message = 'Tolerance seems to be too small.'

Furthermore I changed the string that is added to result.message when no error-bars could be estimated in such a way that no dot at the beginning of the statement is printed but instead one is printed at the end of the statement.

# Tests:
I tested the modified version by fitting some gaussians to some given datapoints but only provoked the cases where 'Fit succeeded.' and where 'One of more variable did not affect the fit', so ier == 4, appeard. But since those two cases worked properly I don't see why any other cases should not work, except for cases in which the definition of ier is wrong. But this definition was not affected by my changes.

### DIFF
--- a/lmfit/minimizer.py
+++ b/lmfit/minimizer.py
@@ -1133,18 +1133,21 @@ class Minimizer(object):
         result.residual = resid = infodict['fvec']
         result.ier = ier
         result.lmdif_message = errmsg
-        result.message = 'Fit succeeded.'
         result.success = ier in [1, 2, 3, 4]
         if result.aborted:
             result.message = 'Fit aborted by user callback.'
             result.success = False
+        elif ier in {1,2,3}:
+            result.message = 'Fit succeeded.'
         elif ier == 0:
-            result.message = 'Invalid Input Parameters.'
+            result.message = 'Invalid Input Parameters. I.e. more variables than data points given, tolerance < 0.0, or no data provided.'
+        elif ier == 4:
+            result.message = 'One or more variable did not affect the fit.'
         elif ier == 5:
             result.message = self._err_maxfev % lskws['maxfev']
         else:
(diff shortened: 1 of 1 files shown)


newville:
@colgan Thanks!  This looks good to me.  Anyone have any objections to merging this?

andyfaff:
I think a `dict` would be neater instead of all th elif clauses.

    msg = {0: 'success', 5: 'too many iterat ions %d' % nfev}

newville:
@colgan, @andyfaff   A dict is a fine way to go here, though I don't have a strong preference for dict over if/elif/else loop.  I'm inclined to let @coglan choose.

colgan:
@andyfaff The problem with the dict is, that the 'success' message corresponds to three different cases, `ier in {1,2,3}`, which would lead to unnecessary duplication. And the if else structure should in fact be slightly faster as long as the first case (success) happens most of the time (which is of course not really relevant in this case, since the difference is probably negligibly small). So it doesn't really matter, right?

Tillsten:
Thanks for you contribution, @colgan .

Issue #384, opened 2016-12-12
values set in param_hints override values passed into Model.make_params()


newville:
See
  https://mail.google.com/mail/u/0/#inbox/158efe1d5e28b004

Simple case exposing bug:

```python
from lmfit.models import PseudoVoigtModel
pmod = PseudoVoigtModel()
params = pmod.make_params(sigma=2, fraction=0.75)
for p in params.values():
   print(p.name, p.value)
```

gives

```python
('amplitude', 1.0)
('sigma', 2)
('center', 0.0)
('fraction', 0.5)
('fwhm', 4.0)
```

`fraction` should be 0.75, not 0.5.

newville:
resolved with #386

PR #386, opened 2016-12-28
allow values for param hints to be overwritten by explicit setting in Model.make_params()

newville:
This PR addresses Issue #384, by having parameter values passed into Model.make_params() overwrite values that may otherwise be set by a parameter hint.

A simple test is included.


Tillsten:
Looks good.

newville:
OK, merging this  one.

PR #387, opened 2016-12-31
Add reducefunc option - custom cost function for scalar minimizers

newville:
This adds a `reducefunc` option to `Minimizer` and `minimize` to specify a non-chi-square cost function for the scalar minimizers.   It replaces #226 and #374.   The discussions for #226 are still relevant, but the code is now out of date.

In this PR, `reducefunc` can be:

* `None` -- sum-of-squares is used
* `negent` -- (rho*log(rho)).sum() is used, where rho = stats.norm.pdf(residual)
* `neglogcauchy` -- neg loglikelihood, using Cauchy distribution, -stats.cauchy.logpdf(residual)
* any callable function that turns an array into a scalar.

There is a test and example, borrowed from #226 in which it is shown that outliers can be suppressed by using log-likelihood with a Cauchy distribution.


andyfaff:
Use `reducefcn` instead?

newville:
@andyfaff uYeah,  no preference her. uuu

Tillsten:
Third option: `reduce_fcn` would be my favourite.

newville:
@tillsten  Good point.  We mostly have multi-word arguments use underscore.  I will change to `reduce_fcn`. Hopefully that's the final variation.

Anyone have any thoughts on whether the `negent` and `neglogcauchy` are useful, sufficient, as alternatives to chi-square?  I thought it would be useful to have some built-in options, but I'm not completely sure what the most common or most statistically meaningful variations would be.  Using  Cauchy/Lorentzian nicely suppresses outliers, as in @tillsten original example (though that used the Student's t distribution, the effect is about the same).  
I'm not sure there is any actual statistical justification for the negative entropy approach.

newville:
@andyfaff @Tillsten OK, I think this has settled, so I'll merge this to master

PR #388, opened 2017-01-16
update to versioneer 0.18

newville:
this updates versioneer to 0.18, the latest version.  This seems to avoid some problems with the latest setuptools on oldish python 2.7.


reneeotten:
As I noticed myself when updating to a newer versioneer version on another project, it adds the following code 
```
from ._version import get_versions
__version__ = get_versions()['version']
del get_versions
```
to your `__init__.py` even if it's already there. So you have now duplicated that code... besides that it should be all good.

newville:
@reneeotten Thanks -- good catch!  I updated the PR.

newville:
@reneeotten  @Tillsten @andyfaff  Any objections to merging this?

andyfaff:
No

reneeotten:
no objections at all

Tillsten:
I don't even have an idea what versioneer does ... :)

newville:
@Tillsten versioneer uses the git commit hash to construct a unique version string like '0.9.5+58.ga7f2ea', which is very useful when developing/debugging.

PR #396, opened 2017-02-21
Fix returned logp when using the PTsampler

abeelen:
Fix the `result.lnprob` attribute when using the `~emcee.PTSampler`.

The confusion comes from emcee when `sampler.lnprobability` has a shape (ntemps, nwalkers, nsteps) for the `~emcee.PTSampler`, while the `~emcee.EnsembleSampler` returns a (nwalkers, nsteps).

### DIFF
--- a/lmfit/minimizer.py
+++ b/lmfit/minimizer.py
@@ -251,8 +251,12 @@ class MinimizerResult(object):
         """
         if hasattr(self, 'chain'):
             if HAS_PANDAS:
-                return pd.DataFrame(self.chain.reshape((-1, self.nvarys)),
-                                    columns=self.var_names)
+                if len(self.chain.shape) == 4:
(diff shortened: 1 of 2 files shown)


andyfaff:
Can you add a test to check that the correct shape arrays are returned for both `ntemps=1` and `ntemps>1` for non default burn and thin? Please check the chain and logp dimensions.

abeelen:
While adding the tests, I took some liberty and also tested for the length of chain/flatchain for both cases `ntemps = 1` & `ntemps > 1`, turns out I also add to change the `flatchain` property

newville:
This looks good to me.  Should it be included with #380?  They both seem like small but clear improvements.

newville:
Is this ready to merge?

andyfaff:
Looks good to me. Thanks @abeelen

newville:
@abeelen @andyfaff Thanks.

Issue #400, opened 2017-02-28
using Parameter.set() does not update symboltable for dependent constraints


newville:
From https://groups.google.com/forum/#!topic/lmfit-py/GsT3ENzvna4

```python
    import lmfit

    p = lmfit.Parameters()
    p.add('independent', value=1., min=0., max=2.)
    p.add('dependent', expr='independent')
    print p['dependent']  # <-- has correct value from expr
    p['independent'].set(value=1.2345)
    print p['dependent']  # <-- value not updated from expr
    p['independent'].value=1.12345
    print p['dependent']  # <-- value updated from expr

```

newville:
fixed with #401

PR #401, opened 2017-02-28
Param set symtable

newville:
This addresses #400, and changes `Parameter.set()` to set the `value` property, not just the internal value in `_val`.  The difference is that the `Parameters` symbol table is also updated with the property, so that dependent parameters using constraint expressions will be updated as soon as they are evaluated.  

A test is included.


newville:
I think there is no controversy here for this bugfix, so I'll merge.

PR #417, opened 2017-05-20
Uncertainties py3

newville:
this allows `lmfit.ufloat()`, which uses the local version of the uncertainties package to work with Python 3.   We're not using this internally, but it's helpful for downstream packages to be able to have this work....


newville:
OK, this now  makes a few more changes (apologies for scope creep).  FWIW, all these changes were added due to downstream code that interacts with `Parameters` and `Parameter` before a real fit, and this was causing some troubles that I cannot fully reproduce using `lmfit` alone.   This now includes some changes to parameters (attn @andyfaff):

1. `Parameters.__deepcopy__()` now does /not/ deepcopy the values in "unique_symbols", it just adds a simple assignment.    All tests pass, and I'm not entirely sure the motivation for __deepcopy__() is the same as it was 2 years ago -- we went through several tweaks here.

2.  `Parameter._getval()` now checks for `value is None` and directly sets the value in the expression symbol table (not using the `value` property).   This also passes all tests, and the result gives a slight improvement in speed with the asv benchmarks.

On the one hand, I'd like to push out 0.9.7 relatively soon, as there are a number of nice fixes to bugs relative to 0.9.6, and my downstream code will be simpler if it can rely on these changes (including this PR).

But, I'm perfectly willing to believe there could be a regression in someone else's downstream code from this PR.  Also: I think we want to try to at least partially address #416 before 0.9.7.  

Anyone have any objections to /this/ PR, or with the concept of a quick fix to #416 and releasing 0.9.7 in a week or so?   Are there other outstanding issues for 0.9.7?

andyfaff:
My memory is hazy on the copying of symbols into the symtable change. AFAICR the changes were part of a move to speed up (un)pickling of Parameters. I may have just used deepcopy because it was used elsewhere.
One thing I do remember vividly from that point in time was that the test suite on Parameter/Parameters wasn't detailed enough. THere were changes to rectify that, but there may still be some gaps.

andyfaff:
Just out of interest, what were the changes lmfit required to the upstream uncertainties code that mean that we have a forked version?

newville:
@andyfaff   Yes, I don't fully remember why deepcopy of Parameters needs to make sure be so deepin copying the asteval symbol table.  I don't know why it would be needed, but I suppose we'll find out. ;)

Re: `uncertainties` package.   Originally, lmfit needed a fix to the uncertainties package to propagate uncertainties to functions that took keyword arguments -- vital for uncertainties of constraint expressions.  It wasn't a major change to the code, but non-trivial.  I submitted a PR to that package but it took forever (like a year) for it to be merged.   During that time it seemed that almost no work was being done on the code.  I pinged the main author a couple times, but I feared the project was being abandoned.  So I just included my forked copy in lmfit for the initial release of lmfit.  I'm not sure it's been updated since.

Subsequently, I do believe that feature has been merged, and other features have been changed as well.  I believe there may be some API changes, I'm sure for the best.   I have no idea what would be needed to update our use of uncertainties to the latest version.

These days PyPI, conda, github, etc make it much easier to depend on external packages.   So, it might be fine to drop our own fork of uncertainties and rely on the official package (it does appear to be developed and maintained).   But I'd rather that wait for 0.9.8 ;).

andyfaff:
Perhaps we could fork uncertainties to the lmfit organisation? That would
give us our own repo.
 Then we could just rely on conda - forge or pip and remove the code we
have in the lmfit package.
If push came to shove then we always have our fork to fall back on.
I have been in contact with the uncertainties author in the past, he is
still doing work on the package. Not sure at what level though.

newville:
@andyfaff That could work.  I don't have a strong opinion on this.   I'm pretty sure it's pure python, so pip, conda, etc is going to work reliably.  We could also consider moving asteval to be a dependency and another package under lmfit.  But, again, I don't see that as urgent.

andyfaff:
uncertainties is available on conda-forge (just checked).

PR #420, opened 2017-06-21
 feat(params): utility methods for setting parm-attrs en-mass / use pandas instead of Parameters

ankostis:
Add utility setters/getters on Parameters class to modify (optionally subsets of) contained parameter instances at once.

The setters are particular handy because they allow the following usage pattern [edit: HAS CHANGED]:
```
>>> params = Parameters(**{p: Parameter() for p in 'abcd'})

## Set a value to all:
>>> params.set_values(1)
>>> params.valuesdict()
OrderedDict([('a', 1), ('b', 1), ('c', 1), ('d', 1)])

## Specify a subset of parameters:
>>> params.set_values(2, 'b', 'c')
>>> params.valuesdict()
OrderedDict([('a', 1), ('b', 2), ('c', 2), ('d', 1)])

## Specify key-values on top of the subset:
>>> params.set_values(3, *'bc', c=4, d=4)
>>> params.valuesdict()
OrderedDict([('a', 1), ('b', 3), ('c', 4), ('d', 4)])

## Assign to a single parameter:
>>> params.set_values(a=5)
>>> params.valuesdict()
OrderedDict([('a', 5), ('b', 3), ('c', 4), ('d', 4)])

## Back to defaults for all:
>>> params.set_values()
>>> params.valuesdict()
OrderedDict([('a', None), ('b', None), ('c', None), ('d', None)])

## Unknown parameters, raise:
>>> params.set_values(k=1)
KeyError: "'l' not in ['a', 'b', 'c', 'd']"
```

Actually the API has been simplified to accept `**kwds` only, like this:
```
params.set_values(a=2.72, c=3.14)
```

If you want to do imitate the above tricks, use this pattern:
```
newvalues = dict.fromkeys(['a', 'b', 'c'], 3.14)
newvalues.update(d=2.718)
params.set_values(**newvalues)
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,5 +43,16 @@ install:
     - python setup.py install
 
 script:
-    - cd tests
-    - nosetests
+    -   nosetests tests
+    - |
+        if [[ "$TRAVIS_PYTHON_VERSION" == "3.6" ]]; then
+            nosetests lmfit/ README.rst \
+                -e .*ui.* \
+                -e lmfit.model \
+                -e lmfit.confidence \
+                -e .*uncertainties.* \
+                --with-doctest \
+                --doctest-options=+NORMALIZE_WHITESPACE,+ELLIPSIS,+REPORT_NDIFF,+IGNORE_EXCEPTION_DETAIL \
+                --doctest-extension=rst
+
+        fi
--- a/lmfit/parameter.py
+++ b/lmfit/parameter.py
@@ -354,13 +354,13 @@ class Parameters(OrderedDict):
 
         Examples
         --------
-        >>>  params = Parameters()
-        # add with tuples: (NAME VALUE VARY MIN  MAX  EXPR  BRUTE_STEP)
+        >>> params = Parameters()
+        >>> # add with tuples: (NAME VALUE VARY MIN  MAX  EXPR  BRUTE_STEP)
         >>> params.add_many(('amp',   10, True, None, None, None, None),
         ...                 ('cen',   4, True,  0.0, None, None, None),
         ...                 ('wid',   1, False, None, None, None, None),
         ...                 ('frac', 0.5))
-        # add a sequence of Parameters
+        >>> # add a sequence of Parameters
         >>> f = Parameter('par_f', 100)
         >>> g = Parameter('par_g',  2.)
         >>> params.add_many(f, g)
--- a/lmfit/parameter.py
+++ b/lmfit/parameter.py
@@ -373,6 +373,100 @@ class Parameters(OrderedDict):
                 param = Parameter(*para)
                 self.__setitem__(param.name, param)
 
+    def _set_pattr(self, attr_name, attr_val, pnames, pvmap):
+        """The private workhorse of all ``set_XXXs()`` methods. """
+        if pnames:
+            pv = dict.fromkeys(pnames, attr_val)
+            pv.update(pvmap)
+        elif pvmap:
+            pv = pvmap
+        else:
+            pv = dict.fromkeys(self, attr_val)
+
+        try:
+            for pname, val in pv.items():
+                setattr(self[pname], attr_name, val)
+        except KeyError as ex:
+            raise KeyError("%s not in %s" % (ex, list(self.keys())))
+
+    def set_values(self, attr_val=None, *pnames, **pvmap):
+        """Utility method to modify subsets of parameters at once.
+
+        Parameters
+        ----------
+        attr_val : set this value as "value" attribute on affected parameters;
+             if undefined, set them to default-value: None.
+             The affected parameters are `pnames` if not empty, or all, except
+             those contained in `pvmap`.  Non-existent names will scream as Inde
+        pnames : subset of parameter names to affect.
+        pvmap : A mapping of parameter-names --> attr_values to set as "value"
+            attribute.
+
+        :raise:
+            KeyError if `pnames` or `pvmap` specify non-existent parameters.
+
+        If both `pnames` and `pvmap` are empty, set all parameters to
+        `attr_val`.
+        If both `pvmap` and `pvmap` specify a parameter, `pvmap` wins.
+
+        Examples
+        --------
+        >>> params = Parameters(**{p: Parameter() for p in 'abcd'})
+        >>> list(params.keys())
+        ['a', 'b', 'c', 'd']
+
+        ## Set a value to all:
+        >>> params.set_values(1)
+        >>> params.valuesdict()
+        OrderedDict([('a', 1), ('b', 1), ('c', 1), ('d', 1)])
+
+        ## Specify a subset of parameters:
+        >>> params.set_values(2, 'b', 'c')
+        >>> params.valuesdict()
+        OrderedDict([('a', 1), ('b', 2), ('c', 2), ('d', 1)])
+
+        ## Specify key-values on top of the subset:
+        >>> params.set_values(3, *'bc', c=4, d=4)
+        >>> params.valuesdict()
+        OrderedDict([('a', 1), ('b', 3), ('c', 4), ('d', 4)])
+
+        ## Assign to a single parameter:
+        >>> params.set_values(a=5)
+        >>> params.valuesdict()
+        OrderedDict([('a', 5), ('b', 3), ('c', 4), ('d', 4)])
+
+        ## Back to defaults for all:
+        >>> params.set_values()
+        >>> params.valuesdict()
+        OrderedDict([('a', None), ('b', None), ('c', None), ('d', None)])
+
+        ## Unknown parameters, raise:
+        >>> params.set_values(k=1)
+        Traceback (most recent call last):
+        KeyError: "'l' not in ['a', 'b', 'c', 'd']"
+        """
+        self._set_pattr('value', attr_val, pnames, pvmap)
+
+    def set_varys(self, attr_val=True, *pnames, **pvmap):
+        """See :method:`set_values()`, default-value: True."""
+        self._set_pattr('vary', attr_val, pnames, pvmap)
+
+    def set_mins(self, attr_val=-inf, *pnames, **pvmap):
+        """See :method:`set_values()`, default-value: -inf."""
+        self._set_pattr('min', attr_val, pnames, pvmap)
+
+    def set_maxs(self, attr_val=inf, *pnames, **pvmap):
+        """See :method:`set_values()`, default-value: inf."""
+        self._set_pattr('max', attr_val, pnames, pvmap)
+
+    def set_exprs(self, attr_val=None, *pnames, **pvmap):
+        """See :method:`set_values()`, default-value: None."""
+        self._set_pattr('expr', attr_val, pnames, pvmap)
+
+    def set_brute_steps(self, attr_val=None, *pnames, **pvmap):
+        """See :method:`set_values()`, default-value: None."""
+        self._set_pattr('brute_step', attr_val, pnames, pvmap)
+
     def valuesdict(self):
         """Return an ordered dictionary of parameter values.
 
@@ -720,15 +814,17 @@ class Parameter(object):
             self.from_internal = lambda val: val
             _val = self._val
         elif self.max == inf:
-            self.from_internal = lambda val: self.min - 1.0 + sqrt(val*val + 1)
+            self.from_internal = lambda val: self.min - \
+                1.0 + sqrt(val * val + 1)
             _val = sqrt((self._val - self.min + 1.0)**2 - 1)
         elif self.min == -inf:
-            self.from_internal = lambda val: self.max + 1 - sqrt(val*val + 1)
+            self.from_internal = lambda val: self.max + 1 - sqrt(val * val + 1)
             _val = sqrt((self.max - self._val + 1.0)**2 - 1)
         else:
             self.from_internal = lambda val: self.min + (sin(val) + 1) * \
-                                 (self.max - self.min) / 2.0
-            _val = arcsin(2*(self._val - self.min)/(self.max - self.min) - 1)
+                (self.max - self.min) / 2.0
+            _val = arcsin(2 * (self._val - self.min) /
+                          (self.max - self.min) - 1)
         return _val
 
     def scale_gradient(self, val):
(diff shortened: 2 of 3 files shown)


newville:
@ankostis I'd be OK with `params.set_values(a=5)`, but I'm not sure I see the other forms here as all that useful.   I'm happy to hear other opinions on this too.  It seems like a large code change for basically adding a convenience method.  

In general, it's a good idea to identify and discuss the problem before proposing a solution. Is there a problem with `params['a'].value = 5`?   Again, that's not to say that I'm opposed to having `params.set_values(a=5)`, but I can't recall when I wanted to set all (or even some) the parameters to the same value.  Is this a common need?

ankostis:
We frequently do set `vary` and `value` on subsets, and min max sometimes en mass.
The rest are not used that often, but i added them to be symmetrical.

newville:
@ankostis 

> We frequently do set vary and value on subsets, and min max sometimes en mass.
The rest are not used that often, but i added them to be symmetrical.

I think it would be highly unusual to want to set parameters to the same value, and see no reason to make that super-extra easy.   I would assume that `params.set_values(7)` would invite mistakes -- no one ever wants to do that, yet it is the default form for this function.  It invites trouble.  Even worse is a form like:
 
> params.set_values(3, *'bc', c=4, d=4)

What if you have a parameter named `bc`?  

So:  the `params.set_values(**kws)` would be OK, the rest are not.  I'm not enthusiastic about similar methods for setting min/max/brute_step/etc for Parameters.   This can be done per parameter, and I think we want to encourage that way of thinking about the problem:  A Parameter has these attributes, and Parameters is an ordered dictionary of Parameter objects.

ankostis:
> > `params.set_values(3, *'bc', c=4, d=4)`
> 
> What if you have a parameter named `bc`?

I'm not sure I understand the problem?
You may wrtie `params.set_values(3, *'bc', c=4, d=4, 'bc'=6)` which is identical to this: `params.set_values(3, 'b', 'c', c=4, d=4, 'bc'=6)`

> So: the params.set_values(**kws) would be OK, the rest are not. I'

Practically what we do is work with "mask" of parameters stored in dictionaries.
We use it among others when running Design of Experiments with "partial" optimizations:

```
subset1 = list('ad')
vary1 = p.get_varys(subset1)
...
p.set_varys(False, vary1)

p.set_varys(True, vary1)

## Restore subset1 to original
p.set_varys(**vary1)


> ... and see no reason to make that super-extra easy.

I'm not sure I agreed with this line of thought, but you are the DDFL for this.

**Comment by newville** [comment]
On Thu, Jun 22, 2017 at 10:51 AM, Kostis Anagnostopoulos <
notifications@github.com> wrote:

> params.set_values(3, *'bc', c=4, d=4)
>
> What if you have a parameter named bc?
>
> I'm not sure I understand the problem?
> You may wrtie params.set_values(3, *'bc', c=4, d=4, 'bc'=6) which is
> identical to this: params.set_values(3, 'b', 'c', c=4, d=4, 'bc'=6)
>

Your proposed change supports (and in its initial PR comment even
advertises) the use of something like `*'bc'`. This works (if that is the
right word) if and only if parameter names are single characters, relying
on the fact that `*'bc'` becomes ('b', 'c') to be used as a varargs.

Single character parameter names are not encouraged and should be extremely
rare.  This PR adds many new methods that have special support for
one-character parameter names.  I think we don't want to encourage that.

So: the params.set_values(**kws) would be OK, the rest are not. I'
>
> Practically what we do is work with "mask" of parameters stored in
> dictionaries.
> We use it among others when running Design of Experiments with "partial"
> optimizations:
>
> subset1 = list('ad')
>
>
Again, this is the essence of the problem as I see it.  I don't think we
should encourage this.  `params.set_values(**kws)` is better.


> vary1 = p.get_varys(subset1)
> ...
> p.set_varys(False, vary1)
>
> p.set_varys(True, vary1)
>
> What is `vary1` there?  I would guess that to be a dictionary of
{param_name: p['param_name'].vary}  or is it a list of bools?  Either way,
what would it mean to pass *that* as a second argument?

Using current code, this would be written like:

   params_i_care_about = ('a', 'b', 'other_param_name')
   for pname in params_i_care_about:
       params[pname].vary = True
       params[pname].value = 8.321
   ...
   for pname in params_i_care_about:
       params[pname].min= 0

There is much to be said for relying on Python instead of writing
setter/getter functions, especially ones with non-standard and slightly
ugly signature of   `Thing.set_attriibute(value, *names)` to mean "set the
attribute of the objects of all items with key in `names` to this value".
That's a bit hard to unwrap, actually.  OTOH,
`Thing.set_attriibute(name1=value1,
name2=value2, ...)` is acceptable.

>
> > ... and see no reason to make that super-extra easy.
>
> I'm not sure I agreed with this line of thought,
>
>
The context was making it easy to set multiple parameter values to the same
initial value.   It is possible already (see above).  I don't see a reason
to add extra methods to make this unusual need ever so slightly easier.

>  but you are the DDFL for this.
>
> I think you misunderstand.   I am neither dictator nor invested "for
life".  This is a library my main software project needs, but even that
larger software project is not my main work task.  It does seem that many
people also find this library useful, and there are a good number of people
who have made substantial contributions. I have learned quite a bit from
this project, and gotten to work with many great people outside my normal
spheres because of it. I find that community aspect both enjoyable and
encouraging. But, this is not my main work nor the only python library I
support, and I don't really feel possessive or it.  I view myself as
/steward/ of this code, documentation, and project, until such time as
someone else wants to take over, or I lose interest.

Again, we're all very happy to see interest and contributions, but that
does not mean that we agree on every idea for change.  We try to strike a
balance of adding features that add value and can be supported and
maintained.

**Comment by ankostis** [comment]
> this PR adds many new methods that have special support for one-character parameter names.  

Not at all - there is no code in this PR whatsoever dealing with single-character parameters, and that wasn't even part of my plan (we don't use single-name params internally unless backed up by the theoretical formulas).

It was for the shake of simplicity that I used single-name parameters in the doctests, and I wouldn't have done if I knew of the misunderstandings.
Of course you can write `set_vary(False, 'foo', 'bar']`.

But if we are to keep just the `kwd` arg, then the above becomes: 

    set_vary(**dict.fromkeys(['foo', 'bar'], False)) 

Not as terse, but I believe it removes the "ugliness" of having to put the "assigned-value" before the "destination-parameters".   I can tell that this ugliness makes it into beautiful code, once you get the grip of how these methods work, and that's what I tried to imprint in the doctests.

**Comment by ankostis** [comment]
Pushed 3cd219ff1 that drops common-value & list-of-params args from the en-masse setters API .
Now only this is acceptable: 

    params.set_values(foo=2.72, bar=3.14)

**Comment by newville** [comment]
@ankostis  I am still lukewarm on this PR.  

First, I am reluctant to include doctest for the Travis tests,  especially using nose.  This was mentioned earlier for #419, but you migrated those changes in here.  This should really be a separate changeset. It's unfortunate that you chose to not bring this up for discussion on the mailing list even when explicitly asked.  Speaking only for myself, I do not like and do not want to use doctest.

Second, I am reluctant to add a pile of `set_*` and `get_*` methods to Parameters, which is really an OrderedDict with some convenience functions.  I  don't see a use case for the `get_*` methods, as they are all reduced to 

```
[getattr(par, attribute) for par in params.values()]
```
with guaranteed order.

Even `set_values()` (the only method I like) is not that hard o emulate with 
```
def set_values(self, **kws):
    for name, val in kws.items():
        self.params[name].value = val
```

I'm not sure this really need it's own method.

But also:  I am willing to defer to opinion of others on these points.

ankostis:
> First, I am reluctant to include doctest for the Travis tests, especially using nose. 
> ...
> Speaking only for myself, I do not like and do not want to use doctest.

Addressed the nose+doctest point earlier.  
If you plan to merge this PR, I can add regular test-cases for getters/setters, but I see no reason to retract doctests - they are useful and improve significantly the quality of this project.

> This was mentioned earlier for #419, but you migrated those changes in here. This should really be a separate changeset.

I'm using a single commit in common from #419, specifically b8ee0a9bf  where doctests for main-code are checked, because the doctests in the `/doc` are an order of magnitude more problematic, and not really finished.  Can't think of other way to ensure changes get merged to other computers in the future ([`rere` functionality works locally only](https://git-scm.com/blog/2010/03/08/rerere.html)).

> don't see a use case for the get_* methods, as they are all reduced to ...

You missed the *subset*, but I agree, it is a one-liner.  I did them not to violate POLA and because we use them in too many places that it worth the effort (we monkeypatch `Parameters`).


> Even set_values() (the only method I like) is not that hard o emulate with ...

It's still 2 lines and your version does not support all-or-nothing - it modifies partially params if a bad name is given.  But here it's lurking something bigger...

> Second, I am reluctant to add a pile of set_* and get_* methods to Parameters, which is really an OrderedDict with some convenience functions. 

You nailed it!
The `OrderedDict` is a poor container for parameters.  
Everybody is using pandas these days, so that would feel much natural:
image

So borth a `Parameter` and the result of the getters would be `pd.Series` that are a much improved `OrderDict.  Which brings us to...

> This should really be a separate changeset. It's unfortunate that you chose to not bring this up for discussion on the mailing list even when explicitly asked. 

Explicitly asked?  
Really you think [that was clear](#issuecomment-309963181)?
But that is no problem anymore - now i have something worthwhile to discuss in the mailing list, and pasted the images here (yes, mailing lists are not that good for technical discussion, no code formatting, no diagrams, etc).
If only I could find the time to do it...:-)

newville:
Hi,

On Mon, Jun 26, 2017 at 11:27 AM, Kostis Anagnostopoulos <
notifications@github.com> wrote:

> First, I am reluctant to include doctest for the Travis tests, especially
> using nose.
> ...
> Speaking only for myself, I do not like and do not want to use doctest.
>
> Addressed the nose+doctest point earlier.
>

Sorry for not understanding, but how was this point addressed?

If you plan to merge this PR, I can add regular test-cases for
> getters/setters, but I see no reason to retract doctests - they are useful
> and improve significantly the quality of this project.
>

Well, I think it's still undecided whether to merge this PR.  I think I
have never given a rousing and enthusiastic response. I could be persuaded,
but I don't see a lot of people clamoring for these changes.

I'm not sure I agree that doctests are a significant improvement if they
impede readability of the docstrings or add a dependency on nose for the
Travis testing.   Doctests are a poor way of testing code.  What they do is
test the docstrings.   But that does not necessarily imply that they need
to part of the automated CI.  They *could* be, but so *could* building the
docs, or building conda packages..... We don't do these things because they
add dependencies and make assumptions that are not necessary.

This was mentioned earlier for #419
> <https://github.com/lmfit/lmfit-py/pull/419>, but you migrated those
> changes in here. This should really be a separate changeset.
>
> I'm using a single commit in common from #419
> <https://github.com/lmfit/lmfit-py/pull/419>, specifically b8ee0a9
> <https://github.com/lmfit/lmfit-py/commit/b8ee0a9bf3319362a99a8298fab2f8d66d92771d>
> where doctests for main-code are checked, because the doctests in the /doc
> are an order of magnitude more problematic, and not really finished. Can't
> think of other way to ensure changes get merged to other computers in the
> future (rere functionality works locally only
> <https://git-scm.com/blog/2010/03/08/rerere.html>).
>

Well, that's not exactly in keeping with our normal workflow.  We try to
not be too prescriptive or demanding, but we also reserve the right to not
merge every PR.    Adding doctests is completely separable from adding more
Parameters.set_*() methods.

don't see a use case for the get_* methods, as they are all reduced to ...
>
> You missed the *subset*, but I agree, it is a one-liner. I did them not
> to violate POLA and because we use them in too many places that it worth
> the effort (we monkeypatch Parameters).
>

Sorry, but I don't understand this.  What subset?   And what is POLA?

Even set_values() (the only method I like) is not that hard o emulate with
> ...
>
> It's still 2 lines and your version does not support all-or-nothing - it
> modifies partially params if a bad name is given. But here it's lurking
> something bigger...
>

A bad name doesn't modify params, it raises a KeyError.  The explicit,
non-builtin-method approach also makes clear that the user interacts with
Parameter objects and tries to keep Parameters as small as reasonably
achievable.  It (the Parameters class) is already pretty big considering
it's primary role is to keep an ordered set of Parameter objects.


> Second, I am reluctant to add a pile of set_* and get_* methods to
> Parameters, which is really an OrderedDict with some convenience functions.
>
> You nailed it!
> The OrderedDict is a poor container for parameters.
> Everybody is using pandas these days, so that would feel much natural:
>

Not everybody is using pandas.  OrderedDict is a poor container for
Parameters? Hm, maybe. OrderedSet might be better.


> [image: image]
> <https://user-images.githubusercontent.com/501585/27549395-1bba5180-5a9c-11e7-9139-2e7a5eabb79a.png>
>
> So borth a Parameter and the result of the getters would be pd.Series
> that are a much improved `OrderDict.
>

I think that changing Parameters to a pandas Series is not likely. First,
moving from a builtin / standard library / numpy type is not likely to
happen.  But also, I think that a pandas Series is not a suitable type for
Parameters.  It has properties that Parameters does not need, and is
missing features Parameters does need.


> Which brings us to...
>
> This should really be a separate changeset. It's unfortunate that you
> chose to not bring this up for discussion on the mailing list even when
> explicitly asked.
>
> Explicitly asked?
> Really you think that was clear
> <#m_7277749766191140557_m_-9138260610541109427_issuecomment-309963181>?
>

Yeah, I do. Also, if you read http://lmfit.github.io/lmfit-py/support.html
you will see that stating with a discussion on the mailing list is strongly
encouraged.

But that is no problem anymore - now i have something worthwhile to discuss
> in the mailing list, and pasted the images here (yes, mailing lists are not
> that good for technical discussion, no code formatting, no diagrams, etc).
> If only I could find the time to do it...:-)
>

Well, this is all volunteer work, so finding the time is always appreciated
and never expected.  But that goes for us too.   We do not feel obligated
to go out of our way to do things we do not want to do.  We're trying to
find a careful balance between supporting new features, fixing problems,
maintaining existing code, and being able to add maintainable features
going forward.

Hope that helps,

--Matt

ankostis:
>  I'm not sure I agree that doctests are a significant improvement if they
impede readability of the docstrings...

Take the time and read my fixes in #419; then show me just 2 places where they "impede readability"? 

> Sorry, but I don't understand this.  What subset?   

Read my code.

> And what is POLA?

Google it.

>  Also, if you read http://lmfit.github.io/lmfit-py/support.html

Of course I've read them; particularly where they say: *"submitting a GH PR would be greatly appreciated."*
I don't think so that now.

> Hope that helps,

Yes it did - to give it up :-) 
Ciao

PR #429, opened 2017-10-23
Add limit to Pearson7 expon to prevent nan error when approaching normal distribution

openafox:
Add limit to Pearson7 expon to prevent nan error when approaching normal distribution

On my system a max above 171 creates a nan error when fitting a normal distribution such as:
`from lmfit.models import Pearson7Model
from scipy.stats import norm
mu = 0
variance = 1
sigma = np.sqrt(variance)
x = np.linspace(mu - 6*sigma, mu + 6*sigma, 100)
y = norm.pdf(x, mu, 1)
mod = Pearson7Model()
pars = mod.guess(y, x=x)
out = mod.fit(y, pars, x=x)
`
So, limit is set to 100 to keep things nice and even.

### DIFF
--- a/lmfit/models.py
+++ b/lmfit/models.py
@@ -508,7 +508,7 @@ class Pearson7Model(Model):
         kwargs.update({'prefix': prefix, 'missing': missing,
                        'independent_vars': independent_vars})
         super(Pearson7Model, self).__init__(pearson7, **kwargs)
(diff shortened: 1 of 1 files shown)


newville:
@openafox  That seems completely reasonable to me.   Anyone object?

PR #431, opened 2017-10-24
Tuple assignment fixed bug causing crash in examples, and some minor changes.

AllanLRH:
The example files didn't work because it assumed that `fit.plot()` returns a Matplotlib `Figure`, bit it returns a tuple with the Matplotlib objects (´Figure´, ´Gridspec´).
This pullrequest simply assign the ´Gridspec´ to the variable ´gridspec´, and ´Figure´ to the variable ´fig´, so that the examples work again.

Also replace print-statements with print-functions, since they work in both Python 2 and 3.

Finally I removed the backend-setting, since the Qt4Agg is not available in all installations.

### DIFF
--- a/examples/benchmark_fit.py
+++ b/examples/benchmark_fit.py
@@ -31,12 +31,12 @@ def run_fit(nruns=100):
         params.add('shift', value= 0.0, min=-np.pi/2., max=np.pi/2)
         params.add('omega', value= 1.0, min=-10.0, max=10.0)
         out = minimize(fcn2min, params, args=(x, data))
-        # print out.params['amp']
+        # print(out.params['amp'])
         assert out.params['amp'].value < 5.5
         assert out.params['amp'].value > 4.5
         assert out.params['omega'].value < 2.25
         assert out.params['omega'].value > 1.75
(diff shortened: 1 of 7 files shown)


newville:
@AllanLRH Thanks.  I agree that all the examples should work with Python3.  (FWIW, we don't have a timeline for dropping Python2, but I would assume that we either follow scipy or aim for mid-2019, following what seems like majority of scientific python libraries).  

Maybe we should make automated tests with more of the examples?   It does seem much better to have examples that work, and not very encouraging to have broken examples.  


AllanLRH:
I like the idea of using the examples as unit tests, but it could be hard to do in practice, besides checking that it doesn't crashes when running under Python 3?

One option would be to export the plots as image-files under Python 2, and and compare the same image files produced in Python 3... something like `abs(gaussianBlur(python2image) - gaussianBlur(python3image)) < someThreshold` in order to account for small rendering differences from font faces and such.
But it does add an extra fairly complex layer which itself could lead to a failed unit test.

newville:
@AllanLRH thanks!  I think we merge these improvements and think about how to better test the examples.

PR #434, opened 2017-10-25
True calc of Voigt height

openafox:
The current estimation of Voigt height can be off by a significant amount. This will calculate the true value. Docs already state height is included.

### DIFF
--- a/lmfit/asteval.py
+++ b/lmfit/asteval.py
@@ -113,6 +113,7 @@ class Interpreter:
         # add scipy symbols
         if self.use_scipy:
             scipy_symtable = {"gamfcn": getattr(scipy.special, "gamma")}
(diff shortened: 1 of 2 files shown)


newville:
@openafox should these values for height calcs be tested?

openafox:
@newville Ok. I created a general test [function](https://github.com/openafox/lmfit-py/blob/24e5058014ac004e10dde44d6e0c639c2fecd2f0/tests/test_height_fwhm.py) and it looks good.

newville:
@openafox  Great, and fast.  That will be very helpful.

openafox:
@newville Should I PR the function?

newville:
Yes please do.

PR #436, opened 2017-10-26
Fix Pearson7 FWHM

openafox:
Running test with #435 revealed the fwhm parameter for Pearson7 was incorrect. Went through the derivation and found this which check good.

### DIFF
--- a/lmfit/models.py
+++ b/lmfit/models.py
@@ -520,8 +520,8 @@ class Pearson7Model(Model):
                        'independent_vars': independent_vars})
         super(Pearson7Model, self).__init__(pearson7, **kwargs)
         self.set_param_hint('expon', value=1.5, max=100)
-        self.set_param_hint('fwhm', expr=fwhm_expr(self))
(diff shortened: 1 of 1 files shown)


newville:
@openafox I applaud the effort.  Would you be willing to make a single PR with as many height/fwhm etc calculations as you think are needed?   And please include docstrings as appropriate and some unit tests.

openafox:
@newville I am working on the derivations for the rest of the peak like functions as I have time. Will make a PR in the next week.

newville:
@openafox  Any progress on this PR?  I fear I may have made too much work for you, and now there is a small conflict with "missing" vs "nan_policy".   Any chance of resolving this?  Thanks,

openafox:
@newville I have been busier than I anticipated. I am through derivations for about half of them. We could close this PR and I will open a new one in the next few weeks with them all.

newville:
@openafox OK, I completely understand!

PR #444, opened 2017-11-13
Fixed emcee model selection example

andyfaff:
This should fix the emcee example referred to in #442.

### DIFF
--- a/examples/lmfit_emcee_model_selection.ipynb
+++ b/examples/lmfit_emcee_model_selection.ipynb
@@ -16,10 +16,8 @@
   },
   {
    "cell_type": "code",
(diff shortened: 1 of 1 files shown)


reneeotten:
thanks @andyfaff - this indeed works again! In retrospect it makes sense that one should use ```residual``` instead of ```lnprob```  as the ```userfunc``` for ```Minimize```. But apparently that did work correctly ~2 years ago when the example was added, or...?

andyfaff:
Hmm. I'm not sure what the history is.

newville:
Thanks @andyfaff  -- @reneeotten  I think this finishes the huge cleanup of #442.

PR #448, opened 2017-11-30
FWHM & Height for all peak like functions

openafox:
Here FWHM and Height parameters are added for all peak like functions. Derivations of both can be found [here](http://openafox.com/science/peak-function-derivations.html). A few functions (Voigt, DHO, Exponential Gaussian, Skewed Gaussian, and Donaich Sunjic) I was not able to derive FWHM for, however, I was able to find estimates or find FWHM by observation. I was not able to find a FWHM for the Donaich Sunjic, so it only includes height <- any help on these would be great ([derivations source](https://github.com/openafox/openafox.github.io/blob/master/_posts/science/2017-11-29-peak-function-derivations.md)). 
A number of other changes were also made:
-Docs were updated to reflect additions.
-The Log-normal line shape is modified to accept int and float input not only lists. Its docs are also updated to reflect the actual equation. Its center parameter is limited to a min of 1e-19 to prevent error.
-The Doniach Sunjic docs are updated to reflect the actual equation.
-The DHO equation is modified (divide by pi rather than *) to reflect the cited ref and docs. The lineshape has also been modified to change x value less than 0 to prevent errors.
-The Exponential Gaussian gamma parameter is limited to a max of 20 to prevent overflow.
-The Students T sigma parameter is limited to a max of 100 to prevent overflow.
-erfc is added to asteval. 

Finally a test function is added to confirm all values are correct, using the test_peak.dat from examples.

One thing of note (likely unimportant) in Breit Wigner "divide by zero encountered in double_sc
alars ast.Div: lambda a, b: a / b," warning is raised


newville:
@openafox Awesome -- thanks so much!  I'm in favor of merging.  Anyone object?

PR #451, opened 2018-01-26
set minimum on tau in expmodel to prevent overflow

openafox:
Ran into this today. Thought I would send along. At tau min = -20 it errs with my data but -15 seems ok.

### DIFF
--- a/lmfit/models.py
+++ b/lmfit/models.py
@@ -904,7 +904,7 @@ class ExponentialModel(Model):
         kwargs.update({'prefix': prefix, 'nan_policy': nan_policy,
                        'independent_vars': independent_vars})
         super(ExponentialModel, self).__init__(exponential, **kwargs)
-        self.set_param_hint('decay', min=-15)
+        self.set_param_hint('decay', min=0)
 
     def guess(self, data, x=None, **kwargs):
         try:
--- a/lmfit/models.py
+++ b/lmfit/models.py
@@ -904,6 +904,7 @@ class ExponentialModel(Model):
         kwargs.update({'prefix': prefix, 'nan_policy': nan_policy,
                        'independent_vars': independent_vars})
         super(ExponentialModel, self).__init__(exponential, **kwargs)
+        self.set_param_hint('decay', min=-15)
 
     def guess(self, data, x=None, **kwargs):
         try:


newville:
@openafox wouldn't that depend on the range of `x`?  I would be more inclined to say that the minimum should be 0 (forcing **decay** by default) than some arbitrary finite number.

Values that become `inf` are normally handled by the `nan_policy` (omit, raise, propagate).  I believe that having an `inf` in the residual will basically break most algorithms -- chi-square will be `inf` and the algorithm won't know which way to go to improve that.   That is, I think it is similar to `nan` in the respect that the fit cannot proceed sensibly.  

But maybe that's not always the case -- any ideas?  Anyway, we don't have a separate `inf_policy` at the moment. We could consider that.

openafox:
@newville I think you are correct. I have updated to tau min = 0. 
However...
Creating a separate inf_policy that defaults to 'omit' might be better than imposing limits on any parameters that could cause overflow (except those that break model definitions). This would allow models to be pushed to limits (without overflowing chi-square) based on the input data rather than an arbitrary (even if well founded) limit. Thoughts? -- Raise as new issue?

newville:
@openafox   I think the question is: should `decay` have a minimum value of 0 by default or not?   I don't have a strong preference

Similarly, I'm not sure why the policy for `inf` should be different than that for `nan`.  I could be convinced, but when does it need to different ?

openafox:
@newville As I think about it more the answer is probably no. Having `tau` go negative does not break the model. 

I may be wrong but the difference I see between `inf` and `nan` is that `inf` is caused by a computational limitation where `nan` is caused by breaking a model (eg: div by 0). Both can be caused by x, y inputs as well but I would suspect that `nan` is more common. Ideally `nan` caused by breaking a model should be prevented by defined parameter limits but `inf`, as it doesn't technically break the model, should not be prevented by parameter limits and should only be omitted to let the fit proceed with the suspect parameter settling at the limit of computation. For both, however, capturing input `nan` and `inf` would be beneficial to the user.
So separating `inf` and `nan` may not be necessary but creating separate `input` and `computation` policies could be beneficial?

I move to close and open an issue. Closing now and will wait to open issue until I hear back.

newville:
@openafox   `decay` having default minimum value of 0 would be OK with me: not vital, but justifiable to prevent problems.  That is, it makes sense for decay, and exponential growth is always in much more danger of getting an`inf`.  So, if the idea is to model exponential growth, the user is likely to need bounds on `decay`. Whether that's with  `param_hints` or the bounds on the `decay` parameter doesn't seem like much of a difference to me.  

For `nan` v `inf`, I guess I was taking a more practical view of floating point numbers than a conceptual view of the cause of the problem.  For the "practical" view, getting either of these are bad news, essentially ending the ability to refine the parameters -- chi-square won't change in a meaningful way.

For the "conceptual" view, I agree that there is more of a case for a distinction, with `nan` often coming from a model function that does not guard against them (or sometimes from the data).  
I could be persuaded that `inf_policy` should be different from `nan_policy`.   At this point, I don't have an opinion....

PR #452, opened 2018-02-04
Evaluate uncertainty for arbitrary independent variable value

mrclary:
Hello, new contributor here.
I've extended the `eval_uncertainty` method in the `ModelResult` class to permit changes to the independent variables. The purpose is to be able to evaluate the uncertainty (confidence band) for the model at arbitrary values of the independent variable. Previously, only uncertainties at the original independent variable values would be evaluated.

Detailed changes:
* don't overwrite ModelResult.userkws
* ensure fjac ad df2 are correct size if independent variable is updated
* ensure original ModelResult.ndata is not changed for computing t.pp

### DIFF
--- a/lmfit/model.py
+++ b/lmfit/model.py
@@ -1305,9 +1305,9 @@ class ModelResult(Minimizer):
 
         nvarys = self.nvarys
         # ensure fjac and df2 are correct size if independent var updated by kwargs
-        ndata = userkws[self.model.independent_vars[0]].size
+        ndata = self.model.eval(self.params, **userkws).size
         covar = self.covar / self.redchi
-        fjac = np.zeros(ndata*nvarys).reshape((nvarys, ndata))
+        fjac = np.zeros((nvarys, ndata))
         df2 = np.zeros(ndata)
 
         # find derivative by hand!
--- a/lmfit/model.py
+++ b/lmfit/model.py
@@ -1306,7 +1306,7 @@ class ModelResult(Minimizer):
         nvarys = self.nvarys
         # ensure fjac and df2 are correct size if independent var updated by kwargs
         ndata = self.model.eval(self.params, **userkws).size
-        covar = self.covar / self.redchi
+        covar = self.covar
         fjac = np.zeros((nvarys, ndata))
         df2 = np.zeros(ndata)
 
@@ -1334,7 +1334,7 @@ class ModelResult(Minimizer):
             prob = sigma
         else:
             prob = erf(sigma/np.sqrt(2))
-        return np.sqrt(df2*self.redchi) * t.ppf((prob+1)/2.0, self.ndata-nvarys)
+        return np.sqrt(df2) * t.ppf((prob+1)/2.0, self.ndata-nvarys)
 
     def conf_interval(self, **kwargs):
         """Calculate the confidence intervals for the variable parameters.
--- a/lmfit/model.py
+++ b/lmfit/model.py
@@ -1298,28 +1298,30 @@ class ModelResult(Minimizer):
            give the same results, within precision errors.
 
         """
-        self.userkws.update(kwargs)
+        userkws = self.userkws.copy()
+        userkws.update(kwargs)
         if params is None:
             params = self.params
 
         nvarys = self.nvarys
-        ndata = self.ndata
+        # ensure fjac and df2 are correct size if independent var updated by kwargs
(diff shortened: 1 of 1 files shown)


newville:
@mrclary  Thanks!  Those look to all like real improvements to me.  

As I look at this, I have one question.  Maybe you've thought about this too?

You changed `pars=self.params` to `pars.self.params.copy()` and moved it out of the loop over variables.  Taking this out of the loop looks good, but I'm not sure the copy is necessary. as `eval()` won't change the parameter values or other attributes, and then the value is set back.   A copy would be needed if the parameters were going to be changed.  OTOH, a copy isn't expensive compared to the rest of this method (2*nvary `evals()` plus a necessary double-loop over nvarys).  So I think it's probably fine either way, but am curious if there might be a reason one way or the other that I'm not seeing.

Anyone else have any concerns, comments?   If not, I'm +1 for merging soon.

mrclary:
@newville Thank you for the response. To answer your question, my motivation for making a copy of the parameters was merely to ensure that the original ModelResults params don't get changed. However, I agree that in this instance it is not necessary, since the value is reset.

One concern I have is the explicit reference to the first independent variable on line 1308 for the assignment of `ndata`. This will certainly break for models with more than one independent variable. Any thoughts? Perhaps `ndata` should be changed from a scalar to a list of independent variable sizes?
```
ndata = [userkws[iv].size for iv in self.model.independent_vars]
fjac = np.zeros((nvarys, *ndata))
df2 = np.zeros(ndata)
``` 
However, I don't know if `scipy.stats.t.ppf` behaves properly under this circumstance.

newville:
>@newville Thank you for the response. To answer your question, my motivation for making a copy of the parameters was merely to ensure that the original ModelResults params don't get changed. However, I agree that in this instance it is not necessary, since the value is reset.

I think either way is OK. 

>    One concern I have is the explicit reference to the first independent variable on line 1308 for the assignment of ndata. This will certainly break for models with more than one independent variable. Any thoughts? 


Yeah, that's another good point.
 

>    Perhaps ndata should be changed from a scalar to a list of independent variable sizes?
>
>        ndata = [userkws[iv].size for iv in self.model.independent_vars]
>        fjac = np.zeros((nvarys, *ndata))
>        df2 = np.zeros(ndata)
>
>    However, I don't know if scipy.stats.t.ppf behaves properly under this circumstance.


`ndata` should be the length of the model (and residual).  Multiple independent variable don't necessarily imply multiple dimensions. Even when they do, the residual needs to be flattened to 1D.   So, the calculation of `ndata` should probably be something like:

```python
     ndata = self.model.eval(self.params, **userkws).size
```

mrclary:
@newville I incorporated your suggestion. In addition, I removed redundant divisor/factor of redchi, and changed the double loop with `numpy.dot`s. The latter gave a speed up of 3.7x for me on the calculation for df2 for ndata=100 and nvarys=4. However, this could be strongly dependent on the size of ndata and nvarys. Since the diagonal must be taken from an ndata-by-ndata array, it may lose it's advantage over the loops for large values of ndata.

Let me know what you think.

mrclary:
Okay, I went back to the loop. For even ndata=1e4, it had a speed _hit_ of 2000x for me.

newville:
@mrclary +1 on removing redundant `redchi`.  Speeding up that calc would be great.  But as a task that was "run once after a good fit", it didn't seem like a priority. Are you saying that using `np.dot`  make it much slower for large `ndata`?

mrclary:
@newville Yes, np.dot was much slower for large data.
ndata=100, nvarys=4 -> loop = 57us, np.dot=14us
ndata=1e4, nvarys=4 -> loop = 270us, np.dot=583ms (yes, milliseconds)

newville:
@mrclary Huh, OK.   That seems like something is fishy with `np.dot`.  Let's leave that out of the PR for now, but maybe that needs study or benchmarking. 

Any objections to merging this?

mrclary:
@newville I don't have objections to merging. I don't think anything is fishy with np.dot; I think the difference is a result of creating a large 1e4x1e4 data array, of which only the diagonal is extracted. The double loop process, while not c-optimized like the numpy functions, doesn't generate a new and larger array and only operates with 1e4 size array.

newville:
@mrclary OK, there were no other comments or objections, I'm going to merge this.  Thanks for the great contribution!

PR #456, opened 2018-03-12
Prefix renames

newville:
This PR allows Model prefixes to be changed after the Model has been created.  Resetting the prefix includes updating the parameter hints to be evaluated correctly.


newville:
I'd like to be able to use this feature in a downstream project, so am eager to merge.  I think that no one was relying on ***not*** being able to change model prefixes.   If this does cause unintended downstream problems, we can figure out a workaround.

Issue #489, opened 2018-08-16
ExpressionModel max_time problem


newville:
See https://groups.google.com/d/msg/lmfit-py/Ex3IfjNjiQE/tOJkOeJWCwAJ for bug report and proposed solution.

PR #490, opened 2018-08-18
specify max_time and evaluation time for ExpressionModel

newville:
This addresses #489, using the remedy proposed there....


newville:
will merge this....

PR #662, opened 2020-08-04
Add sympy example

Tillsten:
#### Description
This PR adds an simple example of using lmfit with sympy.

###### Type of Changes

- [ ] Bug fix
- [ ] New feature

###### Verification 
Have you

Not possible under windows currently.

- [x ] added an example?

### DIFF
--- a/examples/example_sympy.py
+++ b/examples/example_sympy.py
@@ -1,0 +1,77 @@
+"""
+Building a lmfit model with SymPy
+=================================
+
+SymPy is a Python library for symbolic mathematics. It can be very useful to
+build a model with sympy and then use that apply that model to the data with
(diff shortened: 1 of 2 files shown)


newville:
@Tillsten Hi Till, this looks Ok to me.  

Out of curiosity, how does sympy know that `x` is the independent parameter, and then coerce (with its `lambdify`, I guess) that into a function with call signature that lmfit can inspect?  


Would it be worth including "here is how you might do the same steps in plain Python+lmfit, without using sympy"?

newville:
@Tillsten @reneeotten if there are no objections, I'll merge this in the next couple of days.

newville:
@Tillsten @reneeotten  OK, merging this.  Thanks!!

PR #663, opened 2020-08-12
Register models

rayosborn:
#### Description
This PR is based on a post I made to the lmfit Google group (https://groups.google.com/forum/?#!topic/lmfit-py/U3XwKBSLrUQ). Its purpose is to allow third-party packages to determine programmatically all the models defined within lmfit in `models.py` and all the valid forms available to modify some of those models. It makes the following changes:

-  A new variable, `lmfit_models`, which contains a dictionary of all the models, can be directly imported from `models.py`.
- A new Model class attribute, `valid_forms`, contains a tuple of allowed forms. It is empty by default.
- The positional argument, `degree`, of the PolynomialModel constructor is converted to a keyword argument of the same name. This allows the polynomial degree to be set by a `form` keyword argument that would override the `degree` argument. However, the `degree` argument is first and can still be entered as a positional argument as before.

I have tested the functionality of these changes in NeXpy (https://nexpy.github.io/nexpy/), which has a GUI interface to the lmfit library. The new dictionary and class attributes have been used successfully to generate pull-down menus allowing the model and, if appropriate, its form to be chosen. 

Here is an example of the changes in action. Note that the change to the PolynomialModel does not change the existing call signature.

```
>>> from lmfit.models import lmfit_models, PolynomialModel
>>> lmfit_models['Step'].valid_forms
('linear', 'atan', 'arctan', 'erf', 'logistic')
>>> lmfit_models['Polynomial']
lmfit.models.PolynomialModel
>>> p=PolynomialModel(4, prefix='P1')
>>> p.poly_degree
4
>>> p.prefix
'P1'
>>> p=PolynomialModel(4, form='5')
>>> p.poly_degree
5
``` 

###### Type of Changes

- [ ] Bug fix
- [x] New feature

###### Tested on
Python: 3.8.0 (default, Nov  6 2019, 15:49:01) 
lmfit: 1.0.1+53.gb2f308e.dirty (forked from the latest development version in the Master branch)
scipy: 1.4.1, numpy: 1.18.1, asteval: 0.9.16, uncertainties: 3.1.4

###### Verification 
Have you

I haven't added the new class attributes to the docstrings, because none are currently defined, and I don't know what the lmfit convention is.

Running pytest generates 3 errors and 7 warnings, but as far as I can tell, none of them relate to the code changes. I assume this is an issue with the development version. 

I haven't changed the documentation. 

It is unlikely that my changes will be covered by existing tests. 

- [ ] added an example?

I'm happy to do this, but I would need some guidance on how to do these.

### DIFF
--- a/lmfit/model.py
+++ b/lmfit/model.py
@@ -191,10 +191,10 @@ class Model:
     _invalid_par = "Invalid parameter name ('%s') for function %s"
     _invalid_hint = "unknown parameter hint '%s' for param '%s'"
     _hint_names = ('value', 'vary', 'min', 'max', 'expr')
-    forms = []
+    valid_forms = ()
 
     def __init__(self, func, independent_vars=None, param_names=None,
-                 nan_policy='raise', prefix='', name=None, **kws):
+                 nan_policy='raise', prefix='', name=None, form=None, **kws):
         """
         The model function will normally take an independent variable
         (generally, the first argument) and a series of arguments that are
@@ -218,6 +218,8 @@ class Model:
         name : str, optional
             Name for the model. When None (default) the name is the same
             as the model function (`func`).
+        form : str, optional
+            Additional 
         **kws : dict, optional
             Additional keyword arguments to pass to model function.
 
--- a/lmfit/model.py
+++ b/lmfit/model.py
@@ -194,7 +194,7 @@ class Model:
     valid_forms = ()
 
     def __init__(self, func, independent_vars=None, param_names=None,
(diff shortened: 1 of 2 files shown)


newville:
@rayosborn Thanks very much!  I think this looks really good and I will plan on using it in my own codes.  It looks to me like there is little chance of unintended consequences (but @reneeotten always has a keen eye for this sort of thing)
If there are no objections from anyone, I'll plan on merging this in a few days.

newville:
@rayosborn OK, merging this - thanks very much!

PR #664, opened 2020-08-14
FIX: Parameter forced positiv

s-weigand:
#### Description

This fixes a bug introduced by `lmfit==1.0.1` in #631, which forces parameters which can lead to a `ZeroDivisionError` to be positive definite.

##### Example and explanation of the problem
I found this bug while fitting some exponential growth values.
Since `exponential` is implemented as exponential decay:
https://github.com/lmfit/lmfit-py/blob/480072b9f7834b31ff2ca66277a5ad31246843a4/lmfit/lineshapes.py#L468
the value of `decay` needs to be negative, for the equation to represent exponential growth.

Since the `ExponentialModel` wouldn't even fit, when I gave it the parameters used to generate synthetic data as start parameters, I did some digging and found the following lines.
https://github.com/lmfit/lmfit-py/blob/480072b9f7834b31ff2ca66277a5ad31246843a4/lmfit/lineshapes.py#L15
https://github.com/lmfit/lmfit-py/blob/480072b9f7834b31ff2ca66277a5ad31246843a4/lmfit/lineshapes.py#L467

This effectively forces `decay>tiny`, which makes it impossible to fit exponential growth.
The intended meaning to prevent  a `ZeroDivisionError` would have been `abs(decay)>tiny`.
This behaviour can be achieved in the following manner (see: the first two commits):
```python
float(copysign(max(tiny, abs(value)), value))
```
Which is why I implemented a helperfuction `not_zero` (maybe `none_zero` would have been a better name?) and replaced all occurrences of `max(tiny, <value>)`.

##### Sidenotes
While working on this PR I found the  following lines in a test:
https://github.com/lmfit/lmfit-py/blob/480072b9f7834b31ff2ca66277a5ad31246843a4/tests/test_covariance_matrix.py#L118-L124

IMHO (even so the test is implemented to test a different functionality) it is kinda bad practice to generate the data that are to be fitted, with the same function used in the fit, if that function has some extra side effects besides the minimal implementation.
In this example the fit of an exponential growth, would have worked perfectly, while no real exponential growth data were generated. But I also see the problem with code duplication, especially in tests that aren't intended to test the function itself (...difficult topic, the `there is no perfect answer` kind 😉 ).

When running `pre-commit run --all`, the checks `rstcheck` and `yesqa` failed, maybe add this as an additional test to the CI.

The docs won't build on Windows (`gitbash` with `make`), because `doc/doc_examples_to_gallery.py` does not copy the files.
My guess is that this can be solved by replacing `os.system('<command>')` with `subprocess.call('<command>',shell=True)`.
Or dropping the sys calls all together and do everything inside of python. Tell me if you want me to give it a shot 😄 

###### Type of Changes

- [x] Bug fix
- [ ] New feature

###### Tested on

Python: 3.7.8 | packaged by conda-forge | (default, Jul 31 2020, 01:53:57) [MSC v.1916 64 bit (AMD64)]

lmfit: 1.0.1+55.gbfc7da5, scipy: 1.5.2, numpy: 1.19.1, asteval: 0.9.18, uncertainties: 3.1.4

###### Verification 
Have you

- [ ] added an example?


s-weigand:
Adding `not_zero` to `thermal_distribution` as follows:

https://github.com/lmfit/lmfit-py/blob/480072b9f7834b31ff2ca66277a5ad31246843a4/lmfit/lineshapes.py#L383

```python
return real(1/(not_zero(amplitude)*exp((x - center)/not_zero(kt)) + offset + tiny*1j))
```

Will get rid of :
```
tests/test_lineshapes.py::test_no_ZeroDivisionError_and_finite_output[thermal_distribution]
  d:\git\lmfit-py\lmfit\lineshapes.py:401: RuntimeWarning: divide by zero encountered in true_divide
    return real(1/(amplitude*exp((x - center)/kt) + offset + tiny*1j))
```

But cause this instead:
```
tests/test_lineshapes.py::test_no_ZeroDivisionError_and_finite_output[thermal_distribution]
  d:\git\lmfit-py\lmfit\lineshapes.py:401: RuntimeWarning: overflow encountered in exp
    return real(1/(not_zero(amplitude)*exp((x - center)/not_zero(kt)) + offset + tiny*1j))
```
So I'm not sure what would be preferable.

s-weigand:
Maybe reverting https://github.com/lmfit/lmfit-py/pull/664/commits/a1aaa211c8048ab8f4221e5e0937cf65633c9d42 and instead writing tests, against the literal implementation of the lineshapes is more valuable.

newville:
@s-weigand   For the very narrow issue of forcing 'decay' to be not only non-zero but also positive for the `exponential` lineshape and  `ExponentialModel`, the proposed solution introducing `not_zero()` function seems fine with me.   Supporting exponential *growth* models will always be perilous, and the parameter named `decay` might seem a little odd for that. On the one hand, rolling your own exponential growth model just isn't that hard, but on the other, canonical definitions are helpful.  So, I would be OK with that narrow fix and calling the change for `exponential` at #631 an "introduced bug". 

But you have replaced `max(tiny, value)` with your new `not_zero(value)` in a great many places.  It seems that you may not have realized that the forcing of positivity is, in fact, an important part of using `max(tiny, value)` -- it is not an accidental side effect.  It is not saying "not zero", it is saying "positive".   So, for the value of `sigma` in Gaussian, Lorentzian, etc, this will have to change:  these values should be forced to be positive, not merely non-zero. 

Your side note about using `exponential()` in the test of `ExponentialModel` is wrong.   We're testing the fitting, not the evaluation of the model function. The data there can definitely  *should* use data generated by `exponential`.

For the pre-commit problem, you did not include enough details to know what is going on.

On doc generation, my view is that the developers need to be able to generate docs, but we are under no obligation to support the generation of these docs on every platform.   We do test the generation of the docs with Travis (uses some version of debian).   I believe @reneeotten  and I regularly test this by hand on MacOS too.  I do not recall the last time I tried to generate the lmfit docs on Windows.  If you want to fix things to generate docs on Windows, that's OK with me.   We have tests that have to pass ;).

s-weigand:
Thank you for your feedback.

> Supporting exponential growth models will always be perilous, and the parameter named `decay` might seem a little odd for that.

IMHO It isn't odd since for me a negative `growth` is a `decay` and vise versa. Same as negative acceleration is deceleration.

>  It seems that you may not have realized that the forcing of positivity is, in fact, an important part of using max(tiny, value) -- it is not an accidental side effect. It is not saying "not zero", it is saying "positive". So, for the value of sigma in Gaussian, Lorentzian, etc, this will have to change: these values should be forced to be positive, not merely non-zero.

Sorry I really missed this intent completely, acted too rashly and possibly screwed up some things.
While from a purely mathematical point of view (at least for Gaussian, Lorentzian) the only condition is that it is not zero (only way to differentiate between peak and dip in the normalized form).
As a user you would expect that the `amplitude` consistently indicates if is a peak or a dip.

But as I see it there are  other functions (besides `exponential` ) where using `not_zero` instead of `max`, would bring some merit:
- `thermal_distribution`: [since the docs clearly state stat negative values for `kt` are allowed](https://lmfit.github.io/lmfit-py/builtin_models.html#thermaldistributionmodel)
- `step`: allowing `sigma` to be negative would allow fitting of `amplitude - step`, since the sign of sigma only determins if it rises or falls ( but the `guess` method on the model would need to be changed)

Speaking of `step` I just saw that `logistic`, is never used outside of the tests, since `step` and `rectangle` have their own implementation (not sure if I missed something again :cold_sweat:):
https://github.com/lmfit/lmfit-py/blob/2d09bed696ce4e31723b7e591269763d581f98dc/lmfit/lineshapes.py#L386-L458

> Your side note about using `exponential()` in the test of `ExponentialModel` is wrong. We're testing the fitting, not the evaluation of the model function. The data there can definitely should use data generated by `exponential`.

You are compleatly right,  again I acted too rashly in the heat of the moment.
Guess I was frustrated after several hours of debugging my code and since this was the only place `exponential` was directly called I thougth "This could have been prevented if the function wasn't tested against itself." :disappointed:
That said, what do you think about the [implementation tests I proposed](https://github.com/lmfit/lmfit-py/pull/664#issuecomment-674248833)? This could also prevent changing of inteded features as I did with `not_zero` in some cases.

> For the pre-commit problem, you did not include enough details to know what is going on.

If I run `pre-commit run --all` (run on all tracked files) on the current master I get the following output:
```bash
$ pre-commit run --all
pyupgrade................................................................Passed
Check python ast.........................................................Passed
Check builtin type constructor use.......................................Passed
Check for merge conflicts................................................Passed
Debug Statements (Python)................................................Passed
Fix End of Files.........................................................Passed
Mixed line ending........................................................Passed
Trim Trailing Whitespace.................................................Passed
Fix requirements.txt.....................................................Passed
Fix python encoding pragma...............................................Passed
Mixed line ending........................................................Passed
flake8...................................................................Failed
- hook id: flake8
- exit code: 1

tests/test_covariance_matrix.py:229:1: E302 expected 2 blank lines, found 1
tests/test_covariance_matrix.py:237:10: E225 missing whitespace around operator

rstcheck.................................................................Failed
- hook id: rstcheck
- exit code: 1

doc/bounds.rst:28: (ERROR/3) Error in "math" directive:
doc/fitting.rst:326: (ERROR/3) Error in "math" directive:

rst ``code`` is two backticks............................................Passed
codespell................................................................Passed
Strip unnecessary `# noqa`s..............................................Failed
- hook id: yesqa
- exit code: 1
- files were modified by this hook

Rewriting lmfit/printfuncs.py
Rewriting tests/test_itercb.py
Rewriting tests/test_model.py
Rewriting tests/test_nose.py
Rewriting lmfit/minimizer.py
Rewriting tests/test_manypeaks_speed.py
```

Since those tests are for  quality assurance, an extra CI test for that would be good, in case someone forgets to install the hooks with `pre-commit install` and since there are no other QA tests on the CI.

> On doc generation, my view is that the developers need to be able to generate docs, but we are under no obligation to support the generation of these docs on every platform. We do test the generation of the docs with Travis (uses some version of debian). I believe @reneeotten and I regularly test this by hand on MacOS too. I do not recall the last time I tried to generate the lmfit docs on Windows. If you want to fix things to generate docs on Windows, that's OK with me. We have tests that have to pass ;).

Ofc it isn't an 'obligation' since in the end the docs only have to build properly at RTD.
But being able to build the doc cross platform for a cross platform package IMHO is a nice to have, since some people won't have a say in the choice of the OS (i.e. Work/University) but would like to contribute. 
So I will give it a try in a different PR (hopefully that time more thought through 😅)

newville:
On Sun, Aug 16, 2020 at 4:10 PM Sebastian Weigand <notifications@github.com>
wrote:

> Thank you for your feedback.
>
> Supporting exponential growth models will always be perilous, and the
> parameter named decay might seem a little odd for that.
>
> IMHO It isn't odd since for me a negative growth is a decay and vise
> versa. Same as negative acceleration is deceleration.
>

OK. negative 'decay' is sort of like growth.  Anyway, a minor point.

It seems that you may not have realized that the forcing of positivity is,
> in fact, an important part of using max(tiny, value) -- it is not an
> accidental side effect. It is not saying "not zero", it is saying
> "positive". So, for the value of sigma in Gaussian, Lorentzian, etc, this
> will have to change: these values should be forced to be positive, not
> merely non-zero.
>
> Sorry I really missed this intent completely, acted too rashly and
> possibly screwed up some things.
> While from a purely mathematical point of view (at least for Gaussian,
> Lorentzian) the only condition is that it is not zero (only way to
> differentiate between peak and dip in the normalized form).
>

Sigma is typically 'variance' and zo positive definite.

> As a user you would expect that the amplitude consistently indicates if
> is a peak or a dip.
>

Yes, the amplitude can be of any sign or magnitude.

But as I see it there are other functions (besides exponential ) where
> using not_zero instead of max, would bring some merit:
>
>    - thermal_distribution: since the docs clearly state stat negative
>    values for kt are allowed
>    <https://lmfit.github.io/lmfit-py/builtin_models.html#thermaldistributionmodel>
>
>
Yes, negative values of kt are allowed.  Using `not_zero()` might be good
here.


>    - step: allowing sigma to be negative would allow fitting of amplitude
>    - step, since the sign of sigma only determins if it rises or falls (
>    but the guess method on the model would need to be changed)
>
>
The amplitude can be negative.  I'm not sure a negative sigma is sensible.


>
> Speaking of step I just saw that logistic, is never used outside of the
> tests, since step and rectangle have their own implementation (not sure
> if I missed something again 😰):
>
> https://github.com/lmfit/lmfit-py/blob/2d09bed696ce4e31723b7e591269763d581f98dc/lmfit/lineshapes.py#L386-L458
>

Yes, `logistic` and `step(..., form='logistic')` are kind of redundant.
It's nice to be able to use `step()` and be able to change the form.

> Your side note about using exponential() in the test of ExponentialModel
> is wrong. We're testing the fitting, not the evaluation of the model
> function. The data there can definitely should use data generated by
> exponential.
>
> You are compleatly right, again I acted too rashly in the heat of the
> moment.
> Guess I was frustrated after several hours of debugging my code and since
> this was the only place exponential was directly called I thougth "This
> could have been prevented if the function wasn't tested against itself."
> 😞
> That said, what do you think about the implementation tests I proposed
> <https://github.com/lmfit/lmfit-py/pull/664#issuecomment-674248833>? This
> could also prevent changing of inteded features as I did with not_zero in
> some cases.
>

I'm not sure I understand.  What test?

> For the pre-commit problem, you did not include enough details to know
> what is going on.
>
> If I run pre-commit run --all (run on all tracked files) on the current
> master I get the following output:
>
> $ pre-commit run --all
>
> pyupgrade................................................................Passed
>
> Check python ast.........................................................Passed
>
> Check builtin type constructor use.......................................Passed
>
> Check for merge conflicts................................................Passed
>
> Debug Statements (Python)................................................Passed
>
> Fix End of Files.........................................................Passed
>
> Mixed line ending........................................................Passed
>
> Trim Trailing Whitespace.................................................Passed
>
> Fix requirements.txt.....................................................Passed
>
> Fix python encoding pragma...............................................Passed
>
> Mixed line ending........................................................Passed
>
> flake8...................................................................Failed
>
> - hook id: flake8
>
> - exit code: 1
>
>
>
> tests/test_covariance_matrix.py:229:1: E302 expected 2 blank lines, found 1
>
> tests/test_covariance_matrix.py:237:10: E225 missing whitespace around operator
>
>
>
> rstcheck.................................................................Failed
>
> - hook id: rstcheck
>
> - exit code: 1
>
>
>
> doc/bounds.rst:28: (ERROR/3) Error in "math" directive:
>
> doc/fitting.rst:326: (ERROR/3) Error in "math" directive:
>
>
>
> rst ``code`` is two backticks............................................Passed
>
> codespell................................................................Passed
>
> Strip unnecessary `# noqa`s..............................................Failed
>
> - hook id: yesqa
>
> - exit code: 1
>
> - files were modified by this hook
>
>
>
> Rewriting lmfit/printfuncs.py
>
> Rewriting tests/test_itercb.py
>
> Rewriting tests/test_model.py
>
> Rewriting tests/test_nose.py
>
> Rewriting lmfit/minimizer.py
>
> Rewriting tests/test_manypeaks_speed.py
>
>
You'll have to fix the PEP-8 and `noqa` complaints


> Since those tests are for quality assurance, an extra CI test for that
> would be good, in case someone forgets to install the hooks with pre-commit
> install and since there are no other QA tests on the CI.
>
Sorry, I don't understand you here. We use pre-commit, it has to run and it
has to pass for us to be able to see your commit.

On doc generation, my view is that the developers need to be able to
> generate docs, but we are under no obligation to support the generation of
> these docs on every platform. We do test the generation of the docs with
> Travis (uses some version of debian). I believe @reneeotten
> <https://github.com/reneeotten> and I regularly test this by hand on
> MacOS too. I do not recall the last time I tried to generate the lmfit docs
> on Windows. If you want to fix things to generate docs on Windows, that's
> OK with me. We have tests that have to pass ;).
>
> Ofc it isn't an 'obligation' since in the end the docs only have to build
> properly at RTD.
>
We don't regularly push to RTD.  FWIW, I don't know whether the docs would
build with RTD any more.  But it's the same idea - doc generation is
necessary once. It is a fine thing to be able to do more often, but it's OK
if means adding dependencies that using the code does not need otherwise.
We are under no obligation to support all people (or package maintainers)
to be able to build the docs on the system of their choice.  We are never
going to promise that.

> But being able to build the doc cross platform for a cross platform
> package IMHO is a nice to have, since some people won't have a say in the
> choice of the OS (i.e. Work/University) but would like to contribute.
>

We're all volunteers here and this is not a main or paid project for
anyone.  We do that which seems most important.  If that means not
supporting building the docs everywhere, that's ok.  If someone, perhaps
you, wants to see that as a problem worth fixing, that's fine too.

So I will give it a try in a different PR (hopefully that time more thought
> through 😅)
>

OK.

s-weigand:
I've somehow got the feeling that I offended you with the text of my PR (I'm not a native speaker).
If that was the case I sincerely apologize, I did at no point mean to offend/attack someone. 
The only thing I wanted to do is give constructive criticism, since there is pretty much always room for improvement.
That said, I hope you won't understand my comments as an attack, but a question of someone who wants to improve the project even more.

> We're all volunteers here and this is not a main or paid project for anyone. 

Which is why I'm (and I guess all other users are) greatefull for this project, all the time you and the other maintainers spend on it and I offered to try help improve this project even further.

### pre-commit

> Sorry, I don't understand you here. We use pre-commit, it has to run and it
has to pass for us to be able to see your commit.

I'm also fairly new to using pre-commit, but this is how I understand and experienced it.
1. `pre-commit install`
  Installs [git hooks](https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks), on the pre commit event
2. `git commit`
  This runs the hooks/tests, but only on the currently staged files (files you want to add to the commit)
3. If any hook fails (exit != 0)
  This prevents the commit from being made, until all tests succeed.

The only problem here is, that it completely relies on the contributor, to actually install (or update if there were changes) the hooks, which you as maintainer can't see in the PR.
Thus the project collects style violations from contributors that didn't install (update) the precommit hooks, while contributors who installed them have to fix them, if they are unlucky enoght to edit a file with violations.

If you run `pre-commit run --all` on the other hand it checks all tracked files (newly created files won't be checked if they aren't added with `git add <file_path>`).

Having a CI test on travis with  `pre-commit run --all` would enforce a 'clean up your own mess' policy and save time for the maintainers, since the failing test would be fixed by the contributors right away or is at least visable as failing test.

Please correct me if I understood something wrong (I'm still a noob with `pre-commit`).

### Forcing positive signs

> Sigma is typically 'variance' and zo positive definite.

Hmmm guess I got something wrong again, I thought the variance was \sigma^2.

But anyway, I already agreed that I will have to revert most usages of `not_zero`.

What I ment with:
> As a user you would expect that the amplitude consistently indicates if is a peak or a dip.

Is that for many functions, the sign of `sigma` (let's use sigma as pseudo variable, since it mostly is sigma) doesn't change the shape of the peak/dip, but in the nomalized form it changes the fact if it is a peak OR a dip.
Since many  functions have a term similar to `amplitude/sigma` and `amplitude/(-sigma) == (-amplitude)/sigma`, that could lead to a positve `amplitude` while being a dip, just because `sigma` was allowed to be positive.
So in the cases where the sign `sigma` doesn't only influence the symmetry, but also the sign of other parameters (i.e. `amplitude`) I definitly need to change it back to `max(tiny, value)`

```python
In[1]
import numpy as np
from lmfit.lineshapes import gaussian
x =  np.linspace(-10, 10)
y1 = gaussian(x, amplitude=1, sigma=-1)
y2 = gaussian(x, amplitude=-1, sigma=1)
all(y1 == y2)
Out[1]
True
```

### Implementation tests

What I ment with implementation tests ( https://github.com/lmfit/lmfit-py/pull/664#issuecomment-674248833 ) are tests against the basic implemetatioan of a fuction (just type out the fourmula).
With `@pytest.mark.parametrize` this could test all combinations of 'none dangerous' parameters while documenting in the test why some parameters have to be positive or more general why some restriction are implemented.

### logistic and step

> Yes, `logistic` and `step(..., form='logistic')` are kind of redundant.
It's nice to be able to use `step()` and be able to change the form.

I totaly agree with you, but `logistic` isn't used when `step(..., form='logistic')` is called.
Instead this happens:
https://github.com/lmfit/lmfit-py/blob/2d09bed696ce4e31723b7e591269763d581f98dc/lmfit/lineshapes.py#L404-L405

I would expect something like this:
```python
elif form == 'logistic':
    return logistic(x, amplitude, center, sigma)
```

### Documentation

> We don't regularly push to RTD.

Oh yeah sorry, you publish them at [github.io](https://lmfit.github.io/lmfit-py/).
Since I quite often see myself googeling `lmfit documentation` and this gives you RTD aswell as [github.io](https://lmfit.github.io/lmfit-py/), a link in the readme pointing to the up to date docs would be awesome. With out you saying it, I wouldn't have guessed that the docs on RTD might be out of date.

### How to go on
I guess, not to clutter the commit tree, best would be if I made changes according the agreement, we will find here and then force push them (sorry I didn't find anything about this in [CONTRIBUTING.md](https://github.com/lmfit/lmfit-py/blob/master/.github/CONTRIBUTING.md))

newville:
@s-weigand it would be helpful for a PR to address the non-zero problem.  I'm not sure we need an exhaustive discussion of our development process.  I'll try to respond in-line to some points. If you want to *contribute*, then please make a PR that we can act on. 

> The only thing I wanted to do is give constructive criticism, since there is pretty much always room for improvement. That said, I hope you won't understand my comments as an attack, but a question of someone who wants to improve the project even more.

For future reference, you'll probably make more progress by asking questions than making unsolicited PRs and complaining about other people's development practices and choices.  If you're going to cite "good practices" you should know what the practices actually are.

> pre-commit

pre-commit does linting and some checks of the text of the code before a commit is allowed.  Travis does testing of the running of the code, perhaps in multiple configurations.  These are totally different things.  We use both.   If you want to contribute, you will have to use pre-commit.  

>> Forcing positive signs
>> Sigma is typically 'variance' and zo positive definite.
> Hmmm guess I got something wrong again, I thought the variance was \sigma^2.

You're right, `sigma^2` is the variance, `sigma` the standard deviation.  Both are positive.

> Implementation tests
> What I ment with implementation tests ( #664 (comment) ) are tests against the basic implementation 
> of a fuction (just type out the formula).
> With @pytest.mark.parametrize this could test all combinations of 'none dangerous' parameters 
> while documenting in the test why some parameters have to be positive or more general why some 
> restriction are implemented.

Please leave the existing tests alone.  If you're adding new code, add a test for the new code.   Do not also change tests for existing code.

s-weigand:
I removed the inappropriate uses of `not_zero`, so it is only used in `exponential` and `thermal_distribution`.

> You're right, sigma^2 is the variance, sigma the standard deviation. Both are positive.

Since the standard deviation (std) is defined as `\sqrt{variance} = \sqrt{sigma^2} = abs(sigma)`, forcing `sigma>0` IMHO is a useful convention allowing `std = sigma` and ensuring that the sign is always on `amplitude`, but not a mathematical necessity.

angelo-peronio:
This bite me today: the data I am fitting can show either an exponential decay or an exponential growth depending on the experimental conditions.

Thank you @newville @s-weigand  for your work.

newville:
@s-weigand Thanks for updating this PR. I will merge this in a couple of days unless I hear objections, say from @reneeotten.

angelo-peronio:
My unsolicited two pence: the pre-#631 `ExponentialModel()` with no constraints on `decay` was perfect for my use case. If after merging this PR `decay` will be forced to be non-zero, I will have to roll my own model — not that hard, indeed.

My use case: in [scanning tunneling microscopy](https://en.wikipedia.org/wiki/Scanning_tunneling_microscope) the tunneling current depends exponentially on the distance between the sample and the tip of the microscope. With an exponential fit of current-vs-time data I can measure the tip-sample drift velocity, which can be positive, negative or zero.

s-weigand:
@newville Thanks and sorry again for all the uncalled for stuff.
Would you still be interested in a PR making the doc creation cross platform or should I open an issue first?

@angelo-peronio I don't think that you would need to roll your own model, but to make sure could you checkout this PR and try it out?

Here are some tests with `tiny` instead of `0.0`.

```python
In [1]: import numpy as np

In [2]: tiny = np.finfo(np.float64).eps

In [3]: float(tiny)
Out[3]: 2.220446049250313e-16

In [4]: def my_exp(x, amplitude=1, decay=1):
      :     return amplitude*np.exp(-x/decay)
      :

In [5]: my_exp(1, decay=tiny)
Out[5]: 0.0

In [6]: my_exp(1, decay=-tiny)
<ipython-input-4-7707c25ff1e9>:2: RuntimeWarning: overflow encountered in exp
  return np.exp(-x/decay)
Out[6]: inf

In [7]: my_exp(1e-13, decay=tiny)
Out[7]: 2.577230801592414e-196

In [8]: my_exp(1e-12, decay=tiny)
Out[8]: 0.0

In [9]: my_exp(1e-12, decay=-tiny)
<ipython-input-4-7707c25ff1e9>:2: RuntimeWarning: overflow encountered in exp
  return np.exp(-x/decay)
Out[9]: inf

In [10]: my_exp(1e-13, decay=-tiny)
Out[10]: 3.8801336666553966e+195

In [13]: my_exp(1, decay=-0)
ZeroDivisionError                         Traceback (most recent call last)
<ipython-input-13-b8af3c05c4ce> in <module>
----> 1 my_exp(1, decay=-0)

<ipython-input-4-7707c25ff1e9> in my_exp(x, amplitude, decay)
      1 def my_exp(x, amplitude=1, decay=1):
----> 2     return amplitude*np.exp(-x/decay)
      3

ZeroDivisionError: division by zero

In [14]: my_exp(1, decay=0)
ZeroDivisionError                         Traceback (most recent call last)
<ipython-input-14-d028d8df7b9d> in <module>
----> 1 my_exp(1, decay=0)

<ipython-input-4-7707c25ff1e9> in my_exp(x, amplitude, decay)
      1 def my_exp(x, amplitude=1, decay=1):
----> 2     return amplitude*np.exp(-x/decay)
      3

ZeroDivisionError: division by zero

In [15]: my_exp(1, decay=-0.0)
ZeroDivisionError                         Traceback (most recent call last)
<ipython-input-18-03ea746a1a32> in <module>
----> 1 my_exp(1, decay=-0.0)

<ipython-input-15-7cd1ab526cb2> in my_exp(x, amplitude, decay)
      1 def my_exp(x, amplitude=1, decay=1):
----> 2     return amplitude*np.exp(-x/decay)
      3

ZeroDivisionError: float division by zero

In [16]: my_exp(1, decay=0.0)
ZeroDivisionError                         Traceback (most recent call last)
<ipython-input-19-56ad3c223b15> in <module>
----> 1 my_exp(1, decay=0.0)

<ipython-input-15-7cd1ab526cb2> in my_exp(x, amplitude, decay)
      1 def my_exp(x, amplitude=1, decay=1):
----> 2     return amplitude*np.exp(-x/decay)
      3

ZeroDivisionError: float division by zero
```

If it doesn't work (and as a hotfix for now) you can just overwrite the function method of the model.
The only important thing with this approach is that the signatures of the functions match.
As a bonus, you can still use `model. guess (data, x=x)`, since it guesses pretty good for exponential growth as well as decay.

```python
model = ExponentialModel()
model.func = my_exp
```

PR #672, opened 2020-10-01
Feature convolved model

kpounot:
#### Description

Hi everyone,

Here is my first ever pull request.

We are working on a python package for the analysis of quasi-elastic neutron data that would be useful and easy to use for users.
There are two features that would strongly ease the model building process:
1 - analytic convolutions
2 - 2D (momentum transfer q and energy transfer) model fitting with parameters that can be either global, with an explicit q-dependence or independent for each q. 

This pull request address the first point. The analytic convolutions can be useful when we use the so-called fixed-window scans, where the data point in energy axis are very limited (less than 10), which do not allow numeric convolutions. Moreover, analytic convolutions can be faster and more provide a more stable fit than numeric convolutions.

To allow the user to build a model for resolution function convolved with a model for the sample dynamics in simple terms, that is, using Gaussian, Lorentzian, Dirac delta or other, I have written some convolutions function and overloaded the left shoft operator on models to create a ConvolvedModel class.
We can thus obtain something like this:
```
resolution = lmfit.models.PseudoVoigt(prefix='res_')
# ...set model parameters, fit the resolution and fix the fitted parameters.

delta =lmfit.models.DeltaModel(prefix='d_') 
delta.convolutions['pvoigt'] = lmfit.convolutions.conv_delta_pvoigt

l1 = lmfit.models.LorentzianModel(prefix='l1_') 
l1.convolutions['pvoigt'] = lmfit.convolutions.conv_lorentzian_pvoigt

l2 = lmfit.models.LorentzianModel(prefix='l2')
l2.convolutions['pvoigt'] = lmfit.convolutions.conv_lorentzian_pvoigt

# ...set parameter hints

model1 = (delta + l1) << resolution
model2 = (delta + l1 + l2) << resolution
model3 = ((delta + l1 + l2) << resolution) [+ other terms]
```

The resolution function or the model components can thus be easily modified  without having to write an analytic convolution function for each possibilities.

###### Type of Changes

- [ ] Bug fix
- [ x ] New feature

###### Tested on
 Python: 3.6.9 (default, Jul 17 2020, 12:50:27) 
[GCC 8.4.0]
lmfit: 1.0.1+71.g16ec466.dirty, scipy: 1.5.2, numpy: 1.19.2, asteval: 0.9.19, uncertainties: 3.1.4

###### Verification 
Have you

- [ x ] included docstrings that follow PEP 257?

- [ x ] verified that existing tests pass locally?

- [ x ] added or updated existing tests to cover the changes?
- [ x ] added an example?

Got an error with `make` for the documentation (shall I try to fix that directly in that branch?):
```
  File "./doc_examples_to_gallery.py", line 23, in <module>
    os.makedirs(examples_documentation_dir, exist_ok=True)
TypeError: makedirs() got an unexpected keyword argument 'exist_ok'
Makefile:35: recipe for target 'examples/index.rst' failed
make: *** [examples/index.rst] Error 1

```


newville:
@kpounot I don't really understand what this is meant to do or how it is supposed to work. 
Please give a concrete example.  Feeling pretty over-extended, I probably won't have a lot of time to study this in much detail, by which I mean: explain why you want this.

Just as a general comment: starting with an unsolicited feature PR is probably a poor strategy.  For sure, I'm reading this as "what.... and left-shift is overridden, for something...?".   In the future, please start a conversation with the concrete, minimal example, and explain why current options do not work well enough.  Use the mailing list or, if there is an obvious bug, raise an issue.

kpounot:
Hi @newville 

Thanks for your feedback and the useful comment.
I add a, hopefully better and clearer, explanation and an example in the following and I will update the example in the feature branch.

### Context

When measuring dynamics with neutron scattering, we usually use a model for the resolution of the instrument, and one for the dynamics of the sample.

S(q, \omega) = R(q, \omega) \otimes S_{sample}(q, \omega)

where S(q, &omega;) is the measured signal, R(q, &omega;) is the resolution function and S<sub>sample</sub>(q, &omega;) is the dynamic structure factor for the sample, q being the momentum transfer and &omega; the energy transfer.

Let's say the user wants to use analytic convolutions and a pseudo-Voigt profile for R(q, &omega;).
For S<sub>sample</sub>(q, &omega;), the user does not know yet what the best model is, but it consists in a Dirac delta that accounts for elastic scattering and  a sum of Lorentzians that account for quasi-elastic contributions.

One possibility is to write each components of S(q, &omega;) one by one (using VoigtModel and LorentzianModel), but this is not convenient and error-prone, especially if the user wants to make a small change in the resolution function or the sample model or when the resolution function and the sample model contain many terms.
Another possibility is to have the analytic convolutions already defined in the model so that the user can simply write:
`measuredModel = ConvolvedModel(sampleModel, resolutionModel) `.

To achieve that, we can identify the functions being convolved by their names, 'pvoigt' in R(q, &omega;) and 'delta' or  'lorentzian' in S<sub>sample</sub>(q, &omega;). This is done using the `convolutions` attribute, which is a dictionary that associate the other model function name, with the convolution function to be used.

The convolution function act as a standard operator and has the signature `convFunc(model_left, model_right, params, **kwargs)`.

The `ConvolvedModel.eval` method simply takes each pair of functions in **left** and **right** and call the appropriate analytic convolution function for each pair. The different convolutions are then combined together using the operators used in `CompositeModel`, if any.

### Example code for neutron scattering data analysis
```python
    import numpy as np

    from scipy.signal import fftconvolve

    from lmfit import Model
    from lmfit.models import PseudoVoigtModel, LorentzianModel, DeltaModel
    from lmfit.convolutions import conv_lorentzian_pvoigt, conv_delta_pvoigt



    # set the models and their convolutions
    resolution = PseudoVoigtModel(prefix='res_')

    delta = DeltaModel(prefix='d_')
    delta.set_param_hint('amplitude', value=0.5, min=0, max=1)
    delta.convolutions['pvoigt'] = conv_delta_pvoigt

    l1 = LorentzianModel(prefix='l1_')
    l1.set_param_hint('amplitude', min=0, max=1)
    l1.set_param_hint('center', vary=False)
    l1.set_param_hint('sigma', value=5) 
    l1.convolutions['pvoigt'] = conv_lorentzian_pvoigt

    l2 = LorentzianModel(prefix='l2_')
    l2.set_param_hint('amplitude', min=0, max=1)
    l2.set_param_hint('center', vary=False) 
    l2.set_param_hint('sigma', value=12) 
    l2.convolutions['pvoigt'] = conv_lorentzian_pvoigt

    l3 = LorentzianModel(prefix='l3_')
    l3.set_param_hint('amplitude', min=0, max=1)
    l3.set_param_hint('center', vary=False) 
    l3.set_param_hint('sigma', value=26) 
    l3.convolutions['pvoigt'] = conv_lorentzian_pvoigt

    # create some data for resolution and sample
    X = np.linspace(-30, 30, 1000)
    data_res = resolution.eval(x=X, amplitude=0.8, center=0.02, sigma=0.3, fraction=0.05)
    data_res += np.random.normal(scale=0.001, size=X.size)

    data_sample = l1.eval(x=X, amplitude=0.36, center=0.0, sigma=4.3)
    data_sample += l2.eval(x=X, amplitude=0.64, center=0.0, sigma=26.2)
    data_sample += np.random.normal(scale=0.001, size=X.size)

    data_measured = np.convolve(0.1 * data_res, 0.9 * data_sample, mode='same')
    data_measured /= data_measured.sum() * (X[1] - X[0])

    # fit the resolution and fix the parameters
    fit_res = resolution.fit(data=data_res, x=X)
    for key, val in fit_res.best_values.items():
        resolution.set_param_hint(key, value=val, vary=False)

    # build model with 1, 2 and 3 Lorentzians to analyze the data
    # we use the left shift operator for clarity and conciseness
    model1 = (delta + l1) << resolution 
    model1.set_param_hint('d_center', expr='res_center')
    model1.set_param_hint('l1_amplitude', expr='1 - d_amplitude')

    model2 = (delta + l1 + l2) << resolution 
    model2.set_param_hint('d_center', expr='res_center')

    model3 = (delta + l1 + l2 + l3) << resolution 
    model2.set_param_hint('d_center', expr='res_center')

    # fit the different models
    fit_model1 = model1.fit(data=data_measured, x=X, method='basinhopping', fit_kws={'niter':10})
    fit_model2 = model2.fit(data=data_measured, x=X, method='basinhopping', fit_kws={'niter':10})
    fit_model3 = model3.fit(data=data_measured, x=X, method='basinhopping', fit_kws={'niter':10})

    # plot the fits, the bayesian information criterium (BIC) shows we are already
    # overfitting our data with three Lorentzians
    fig, ax = plt.subplots(1, 3)
    fit_model1.plot_fit(ax[0])
    ax[0].lines[1].set_label('best-fit, BIC = %.2f' % fit_model1.bic)
    ax[0].legend()
    fit_model2.plot_fit(ax[1])
    ax[1].lines[1].set_label('best-fit, BIC = %.2f' % fit_model2.bic)
    ax[1].legend()
    fit_model3.plot_fit(ax[2])
    ax[2].lines[1].set_label('best-fit, BIC = %.2f' % fit_model3.bic)
    ax[2].legend()
```

newville:
The question for me is definitely:  why does what you want to do mean that
**lmfit** has to change -- why can this not be all done in user code?  Why
does `lmfit.Model` need a `convolutions`, and what is it meant to do?   Is
there a benefit here for other lmfit users?

On Fri, Oct 2, 2020 at 5:29 AM Kevin Pounot <notifications@github.com>
wrote:

> Hi @newville <https://github.com/newville>
>
> Thanks for your feedback and the useful comment.
> I add a, hopefully better and clearer, explanation and an example in the
> following and I will update the example in the feature branch.
> Context
>
> When measuring dynamics with neutron scattering, we usually use a model
> for the resolution of the instrument, and one for the dynamics of the
> sample.
>
> [image: S(q, \omega) = R(q, \omega) \otimes S_{sample}(q, \omega)]
> <https://camo.githubusercontent.com/5cbed45144840e2bf6295e420519970673f45140/687474703a2f2f7777772e736369776561766572732e6f72672f74657832696d672e7068703f65713d53253238712532432532302535436f6d65676125323925323025334425323052253238712532432532302535436f6d6567612532392532302535436f74696d6573253230535f25374273616d706c65253744253238712532432532302535436f6d6567612532392662633d57686974652666633d426c61636b26696d3d6a70672666733d31322666663d6172657626656469743d30>
>
> where S(q, ω) is the measured signal, R(q, ω) is the resolution function
> and Ssample(q, ω) is the dynamic structure factor for the sample, q being
> the momentum transfer and ω the energy transfer.
>
> Let's say the user wants to use analytic convolutions and a pseudo-Voigt
> profile for R(q, ω).
>

I am already pretty skeptical. You seem to believe that 'analytic
convolutions' are going to be intrinsically better than numerical
convolutions, and yet you use a pseudo-Voigt as the resolution function.
Pseudo-Voigt is not ever the correct convolution function.  It is an
approximation of a true Voigt function (itself meant to convolve Lorentzian
and Gaussian responses), apparently from the days when people modeling
powder diffraction data were too lazy to implement the Faddeeva function in
Fortran or C.   I find the use of pseudo-Voigt to *always* mean laziness
and a lack of rigor.  That simply does not fit with a concern about using
numerical convolutions.

For Ssample(q, ω), the user does not know yet what the best model is, but
> it consists in a Dirac delta that accounts for elastic scattering and a sum
> of Lorentzians that account for quasi-elastic contributions.
>
One possibility is to write each components of S(q, ω) one by one (using
> VoigtModel and LorentzianModel), but this is not convenient and
> error-prone, especially if the user wants to make a small change in the
> resolution function or the sample model or when the resolution function and
> the sample model contain many terms.
>
Well, you would have to do "something" one-by-one if you want S(q, omega)
to be a user-definable sum of primitive lineshapes.  Or you could write a
specialized Model that included doing whatever work you wanted.

Another possibility is to have the analytic convolutions already defined in
> the model so that the user can simply write:
> measuredModel = ConvolvedModel(sampleModel, resolutionModel) .
>
> To achieve that, we can identify the functions being convolved by their
> names, 'pvoigt' in R(q, ω) and 'delta' or 'lorentzian' in Ssample(q, ω).
> This is done using the convolutions attribute, which is a dictionary that
> associate the other model function name, with the convolution function to
> be used.
>

Why does this need to be a feature of the Model class?  Why don't you just
write
      measuredModel = ConvolvedModel(sampleModel, resolutionModel)

or
     measuredModel = ConvolvedModel(sampleModel, kernel='voigt')

or something?   Whether that does numerical or analytic convolution would
be an implementation detail.  Which, well, it kind of should be.



> The convolution function act as a standard operator and has the signature convFunc(model_left,
> model_right, params, **kwargs).
>

Isn't that the case already?

The ConvolvedModel.eval method simply takes each pair of functions in *left*
> and *right* and call the appropriate analytic convolution function for
> each pair. The different convolutions are then combined together using the
> operators used in CompositeModel,
>

Why can't you do that now, without any changes to `lmfit.Model`?

if any.
> Example code for neutron scattering data analysis
>
>     import numpy as np
>     from scipy.signal import fftconvolve
>
>     import matplotlib.pyplot as plt
>
>     from lmfit import Model
>     from lmfit.models import PseudoVoigtModel, LorentzianModel, DeltaModel
>     from lmfit.convolutions import conv_lorentzian_pvoigt, conv_delta_pvoigt
>
>     # set the models and their convolutions
>     resolution = PseudoVoigtModel(prefix='res_')
>
>
>     delta = DeltaModel(prefix='d_')
>     delta.set_param_hint('amplitude', value=0.5, min=0, max=1)
>
>
Hm, values should not be set in parameter hints.   Basically, user-level
code should not use parameter hints.

    delta.convolutions['pvoigt'] = conv_delta_pvoigt
>
>
Um, what does that do?  And, what would it do if someone set
       delta.convolutions['foo'] = np.sin

cause, like, why not?

    # build model with 1, 2 and 3 Lorentzians to analyze the data
>     # we use the left shift operator for clarity and conciseness
>
>     model1 = (delta + l1) << resolution
>
>
How does over-riding the meaning of "left-shift" help clarity?  Why does
"left-shift" imply "convolution"?   I'm **very** skeptical of this.  I also
do not understand why you would need to apply the convolution here if there
is also something held in `model.convolutions`.

I'm not really getting any of this.  I think I would need a lot of
convincing here.

kpounot:
@newville,

Ok, I see your point.
Indeed, all the changes could be done within the class `ConvolvedModel`, such that I can remove all the changes that are in `lmfit.Model` or elsewhere.

The convolution can already be performed using `model = ConvolvedModel(left, right)`. The `Model.convolutions` attribute was here to define the mapping "(model1, model2) -> operator1, (model1, model3) -> operator2", it does not perform the convolution yet, this is done upon call to `ConvolvedModel.eval`. This could be removed, then the mapping between each pair of functions and operator could be defined within the `ConvolvedModel` class itself and there would be no change to the original code.

Actually, thinking about it, the class could be even more general, something like `CustomOperatorModel`. 
If I understand correctly the current version of **lmfit**, if the operator to be applied depends on the functions in *left* and *right*, then you can still write `CompositeModel(model1, model2, operator1) + CompositeModel(model1, model3, operator2)`. But you would have conflicting parameter names, so the user would need to create two instances of *model1* and then set the parameters hints `expr` to tell that the parameters of model1 in the first term are the same as the parameters of model1 in the second term. 

In the `ConvolvedModel` class, the idea was to define the mapping "(model1, model2) -> operator1, (model1, model3) -> operator2", inside the models. So if you write `ConvolvedModel(CompositeModel(model1, model2, operator.__add__), model3)`, each model appears only once, and you don't have to create two instances of *model1*. 
On the call `eval()`, the *operator1* and *operator2* are applied on the pairs (model1, model2) and (model1, model3), respectively. The model1, model2 and model3 are not yet evaluated, such that the operators have access to the full `Model` instance.
Finally, the *operator.__add__* from `CompositeModel` is applied on the result like:  `operator1(model1, model2) + operator2(model1, model3)`.


If you feel this does not justify a new feature in **lmfit**, I can simply implement the `ConvolvedModel` in our API for neutron data analysis, which will mostly rely on **lmfit** anyway (with or without analytic convolutions).

newville:
i'm pretty sure we won't merge this...

PR #673, opened 2020-10-04
Make docs build cross platform

s-weigand:
#### Description
As promised some time ago, here is the PR to make the docs build cross platform.
I think the main problem with building the docs on Windows were the `RegEx`s for the paths in `sphinx_gallery_conf ` (at least with git bash).
Rewriting `doc/doc_examples_to_gallery.py` to use python to copy files instead of relieing on sys calls to `cp` makes it work on `cmd` and `powershell`.
A positive side effect is that the execution time is only half of what it used to be (5s vs. 10s before).

###### Type of Changes

- [x] Bug fix
- [ ] New feature

###### Tested on

Windows 10
```
Python: 3.8.5 | packaged by conda-forge | (default, Jul 31 2020, 01:53:45) [MSC v.1916 64 bit (AMD64)]
lmfit: 1.0.1+71.gc67d55e, scipy: 1.5.2, numpy: 1.19.1, asteval: 0.9.18, uncertainties: 3.1.4
```
With:
- `cmd`
- `powershell`
- `git-bash`

Building the docs on Windows with `cmd` or `powershell` can be done by calling `/doc/make.bat html`.
I only changed the build instructions for html, to mimic the Makefile ones, since it suffices as test if the docs still build properly.

Ubuntu 20.04
```
Python: 3.8.5 (default, Sep  4 2020, 07:30:14)[GCC 7.3.0]
lmfit: 1.0.1+71.gc67d55e, scipy: 1.5.2, numpy: 1.19.2, asteval: 0.9.19, uncertainties: 3.1.4
```
With:
- `bash`

###### Verification 
Have you

###### Heads-up about numpy problems on Windows 2004
I hope this helps to identify reported errors with `lmfit` on Windows, which aren't problems with `lmfit` but a windows update breaking parts of `numpy`.
Sadly, after [Windows 10 update to 2004 broke numpy polyfit and eig regression](https://github.com/numpy/numpy/issues/16744), I again can't build the docs on Windows and the tests suite fails.
Before that update it worked fine, but now it crashes when evaluating `examples\documentation\model_savemodelresult2.py` with `numpy.linalg.LinAlgError: SVD did not converge in Linear Least Squares`, which is exactly the problem described in the numpy issue.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -166,8 +166,8 @@ image_converter_args=["-density", "300"]
 sphinx_gallery_conf = {
     'examples_dirs': '../examples',
     'gallery_dirs': 'examples',
-    'filename_pattern': '/documentation|/example_',
-    'ignore_pattern': '/doc_',
+    'filename_pattern': r'(\\|/)documentation|(\\|/)example_',
+    'ignore_pattern': r'(\\|/)doc_',
     'ignore_repr_types': r'matplotlib',
 }
 
--- a/doc/doc_examples_to_gallery.py
+++ b/doc/doc_examples_to_gallery.py
@@ -11,55 +11,65 @@ Process the examples in the documentation for inclusion in the Gallery:
 - copy the data files
 
 """
+
 import os
-import time
+import shlex
+import subprocess
 
-basedir = os.getcwd()
+from pathlib import Path
+from shutil import copy2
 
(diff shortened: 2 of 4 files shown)


newville:
@s-weigand  Thanks very much for this, and for using `pathlib`  (and sorry for the delay in merging this!)

PR #678, opened 2020-10-27
Fix #677

### LINKED ISSUE #677: ModelResult.dumps() broken for lbfgsb method
#### Description
When I use the limited BFGS (lbfgsb) method to fit, I cannot use dumps().
It works fine when I do not specify a method, or if i specify the "bfgs" method, but the 'lbfgsb' method fails.

```Python
from numpy import exp, loadtxt, pi, sqrt

from lmfit import Model

data = loadtxt('model1d_gauss.dat')
x = data[:, 0]
y = data[:, 1]

def gaussian(x, amp, cen, wid):
    """1-d gaussian: gaussian(x, amp, cen, wid)"""
    return (amp / (sqrt(2*pi) * wid)) * exp(-(x-cen)**2 / (2*wid**2))

gmodel = Model(gaussian)
result = gmodel.fit(y, x=x, amp=5, cen=5, wid=1, method='lbfgsb')

print(result.fit_report())

s = result.dumps()
```

###### Error message:

```
<ipython-input-9-b8a48d1fdbd7> in <module>
     17 result = gmodel.fit(y, x=x, amp=5, cen=5, wid=1, method='lbfgsb')
     18 
---> 19 s = result.dumps()

/usr/local/Cellar/python@3.8/3.8.6/Frameworks/Python.framework/Versions/3.8/lib/python3.8/site-packages/lmfit/model.py in dumps(self, **kws)
   1629                 val = bool(val)
   1630             out[attr] = encode4js(val)
-> 1631         return json.dumps(out, **kws)
   1632 
   1633     def dump(self, fp, **kws):

/usr/local/Cellar/python@3.8/3.8.6/Frameworks/Python.framework/Versions/3.8/lib/python3.8/json/__init__.py in dumps(obj, skipkeys, ensure_ascii, check_circular, allow_nan, cls, indent, separators, default, sort_keys, **kw)
    229         cls is None and indent is None and separators is None and
    230         default is None and not sort_keys and not kw):
--> 231         return _default_encoder.encode(obj)
    232     if cls is None:
    233         cls = JSONEncoder

/usr/local/Cellar/python@3.8/3.8.6/Frameworks/Python.framework/Versions/3.8/lib/python3.8/json/encoder.py in encode(self, o)
    197         # exceptions aren't as detailed.  The list call should be roughly
    198         # equivalent to the PySequence_Fast that ''.join() would do.
--> 199         chunks = self.iterencode(o, _one_shot=True)
    200         if not isinstance(chunks, (list, tuple)):
    201             chunks = list(chunks)

/usr/local/Cellar/python@3.8/3.8.6/Frameworks/Python.framework/Versions/3.8/lib/python3.8/json/encoder.py in iterencode(self, o, _one_shot)
    255                 self.key_separator, self.item_separator, self.sort_keys,
    256                 self.skipkeys, _one_shot)
--> 257         return _iterencode(o, 0)
    258 
    259 def _make_iterencode(markers, _default, _encoder, _indent, _floatstr,

/usr/local/Cellar/python@3.8/3.8.6/Frameworks/Python.framework/Versions/3.8/lib/python3.8/json/encoder.py in default(self, o)
    177 
    178         """
--> 179         raise TypeError(f'Object of type {o.__class__.__name__} '
    180                         f'is not JSON serializable')
    181 

TypeError: Object of type bytes is not JSON serializable
```

###### Version information
```
Python: 3.8.6 (default, Oct  8 2020, 14:07:53) 
[Clang 11.0.0 (clang-1100.0.33.17)]

lmfit: 1.0.1, scipy: 1.5.2, numpy: 1.19.2, asteval: 0.9.19, uncertainties: 3.1.4
```

leonfoks:
#### Description
Fixes #677 where dumping a fit using the lbfgsb method would barf.  scipy.optimize was returning a bytes object instead of string. 
I added a conversion of the key "message" to str(kwargs.get('message', '')).  Explicit use of "utf-8" in str() would let the lbfgsb method work, but cause the other methods to fail.  Using 'utf-8' was not needed in the end.

###### Type of Changes

- [ ] Bug fix
- [ ] New feature
- [x] Workaround for scipy.optimize bug

###### Tested on
```
Python: 3.8.6 (default, Oct  8 2020, 14:07:53) 
[Clang 11.0.0 (clang-1100.0.33.17)]

lmfit: 1.0.1+84.g7b45bbb.dirty, scipy: 1.5.2, numpy: 1.19.2, asteval: 0.9.19, uncertainties: 3.1.4
```

###### Verification 
Have you

- [ ] included docstrings that follow PEP 257? N/A

- [ ] referenced existing Issue and/or provided relevant link to mailing list? N/A

I had two fails which I doubt are related to my PR.
FAILED tests/test_models.py::TestSineModel::test_less_than_1_period[1] - AssertionError: wrong fit for parameter frequency: expected 0.65, fitted 0.9133860189041971.
FAILED tests/test_models.py::TestSineModel::test_typical_regime[6] - AssertionError: wrong fit for parameter amplitude: expected 23.5, fitted 22.529682640337313.

- [ ] verified that the documentation builds locally? N/A

- [x] added or updated existing tests to cover the changes? test_saveload
- [ ] added an example?

### DIFF
--- a/lmfit/model.py
+++ b/lmfit/model.py
@@ -1625,6 +1625,9 @@ class ModelResult(Minimizer):
             if isinstance(val, np.bool_):
                 val = bool(val)
             out[attr] = encode4js(val)
+
+        out['message'] = str(out.get('message', ''))
+
         return json.dumps(out, **kws)
 
     def dump(self, fp, **kws):
--- a/lmfit/model.py
+++ b/lmfit/model.py
@@ -1624,9 +1624,12 @@ class ModelResult(Minimizer):
                 continue
             if isinstance(val, np.bool_):
                 val = bool(val)
+
             out[attr] = encode4js(val)
 
-        out['message'] = str(out.get('message', ''))
+        val = out.get('message', '')
+        if isinstance(val, bytes):
+            out['message'] = str(val, encoding='ASCII')
 
         return json.dumps(out, **kws)
 
--- a/tests/test_saveload.py
+++ b/tests/test_saveload.py
@@ -186,7 +186,7 @@ def test_saveload_modelresult_exception():
 
 
 @pytest.mark.parametrize("method", ['leastsq', 'nelder', 'powell', 'cobyla',
-                                    'bfgsb', 'differential_evolution', 'brute',
+                                    'bfgs', 'lbfgsb', 'differential_evolution', 'brute',
                                     'basinhopping', 'ampgo', 'shgo',
                                     'dual_annealing'])
 def test_saveload_modelresult_roundtrip(method):


newville:
@leonfoks yes, but

```
>>> str(message)
"b'a message'"
>>> str(message, encoding='ASCII')
'a message'
```

so adding the encoding there is probably preferred.  And, just to be clear, that message is coming from a C (or Fortran) shared library and is really in ASCII, or you could use 'utf-8'.

leonfoks:
If I add encoding to the line I added,

out['message'] = str(out.get('message', ''), encoding='ASCII')

I get "TypeError: decoding str is not supported" when using methods other than lbfgsb. So adding the encoding is not a problem for the lbfgsb method, but for other methods it now fails.

Just to confirm, yes, the string is being converted from a bytes object to str object, and the instance contains b'text' in it before and after conversion.

Without encoding, its working for all method so I can

a) leave as is or

b) add a check to the loop over attrs, but this assumes all bytes objects should be str
```
if isinstance(val, bytes):
    val = str(val, encoding='ASCII')
```

or this outside the loop so we only do this for the offending message
```
val = out['message']
if isinstance(val, bytes):
    out['message'] = str(val, encoding='ASCII')
```

newville:
@leonfoks hm, I'm sorry this is turning into more work.   So far, it seems like it is only 'message' from `lbfgsb`, and it 
It looks like the problem will be solved in scipy soon (but we should solve it too for existing versions).

Anyway, maybe solving the narrow problem -- checking only for `isinstance(out['message'], bytes)` -- is the best approach.  If we hit a similar problem, then we can think about a more global solution.

leonfoks:
No problem! I’ll get that submitted

newville:
@leonfoks great, thanks!

Issue #80, opened 2014-01-02
Minor fixes to make scalar minimization more usable


sdh4:
I've needed to make three minor fixes to lmfit, listed below, to get scalar minimization with the conjugate gradient (cg) algorithm working. All changes are to lmfit/minimizer.py:
#1. Rationale: The way scalar_minimize is declared and used the 'tol=None' parameter seems to override, and yet it cannot be set explicitly through minimize(). With this change and the first part of #2, below, a tol= parameter on the minimize() command line is properly passed through to the scipy minimize code. (perhaps hess= should be removed also)

Patch: 

```
@@ -343,7 +343,7 @@ or set  leastsq_kws['maxfev']  to increase this maximum."""
         self.unprepare_fit()
         return

-    def scalar_minimize(self, method='Nelder-Mead', hess=None, tol=None, **kws):
+    def scalar_minimize(self, method='Nelder-Mead', hess=None, **kws):
         """use one of the scaler minimization methods from scipy.
         Available methods include:
           Nelder-Mead
```
#2. Rationale: The first line of this change completes fixing the tolerance parameter. The remainder of the change allows passing a jacobian to minimizers (such as cg) that can make use of one. Before this fix, while jac can be supplied, it does not get user arguments, and its parameters are not scaled according to the algorithm used to make variables bounded.

Patch: 

```
@@ -376,10 +376,20 @@ or set  leastsq_kws['maxfev']  to increase this maximum."""
         if method not in ('L-BFGS-B', 'TNC', 'SLSQP'):
             opts['maxfev'] = maxfev

-        fmin_kws = dict(method=method, tol=tol, hess=hess, options=opts)
+        fmin_kws = dict(method=method, hess=hess, options=opts)
         fmin_kws.update(self.kws)
         fmin_kws.update(kws)

+        if 'Dfun' in fmin_kws and fmin_kws['Dfun'] is not None and not isinstance(fmin_kws['Dfun'],bool):
+            # Provided an explicit derivative (jacobian) function
+            self.jacfcn = fmin_kws['Dfun']
+
+            # scipy.minimize uses 'jac' to name the jacobian parameter, not Dfun
+            del fmin_kws['Dfun'] 
+            fmin_kws['jac'] = self.__jacobian
+
+
+
         ret = scipy_minimize(self.penalty, self.vars, **fmin_kws)
         xout = ret.x
         self.message = ret.message
```
[...]

PR #681, opened 2020-11-16
Fix use of callable Jacobian for scalar methods

mstimberg:
#### Description
The "scalar" methods (i.e. all except for `leastsq`) do not seem to correctly wrap the provided Jacobian to deal with parameter bounds. In addition, the code seems to attempt to translate the `Dfun` argument to `jac` for these methods, but the code to do so can never succeed (if `jac` is not in `fmin_kws`, popping it will raise an error):
https://github.com/lmfit/lmfit-py/blob/c3d07725a0ca1bcb3241182904f8a3338035edaf/lmfit/minimizer.py#L975-L977

From looking at #80, I think this came about by mixing up the two issues.

This PR should fix the issues, I added a very basic test that basically copies `test_bounded_jac` but uses the `CG` method instead. I guess an additional example or more details in the documentation would be helpful as well, but it'd take me a bit more time to provide this.

###### Type of Changes

- [X] Bug fix
- [ ] New feature

###### Tested on

Python: 3.8.5 (default, Sep  4 2020, 07:30:14) 
[GCC 7.3.0]
lmfit: 1.0.1+89.ga70ef7e, scipy: 1.5.4, numpy: 1.19.4, asteval: 0.9.21, uncertainties: 3.1.4

###### Verification 
Have you
- [X] verified that existing tests pass locally?

- [X] squashed/minimized your commits and written descriptive commit messages?

- [X] added or updated existing tests to cover the changes?

### DIFF
--- a/lmfit/minimizer.py
+++ b/lmfit/minimizer.py
@@ -519,6 +519,7 @@ class Minimizer:
         self.residual = None
         self.reduce_fcn = reduce_fcn
         self.params = params
+        self.col_deriv = False
         self.jacfcn = None
         self.nan_policy = nan_policy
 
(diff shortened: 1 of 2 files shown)


newville:
@mstimberg Thanks - I guess it is kind of confusing how it ever worked in one specified 'Dfun' and not 'fmin'!  And thanks too for adding the test.  

I think this looks good.  If anyone has any other comments or suggestions, feel free to comment here. Otherwise, I'll expect to merge this in a day or 2.

newville:
@mstimberg Sorry for the delay, let's merge this, and: Thanks very much!

Issue #617, opened 2020-02-04
Fits to Models fail for x and y input data with certain datatypes, but not others


heltluke:
#### Description
Fits using `set_param_hint()`, `make_params()`, and `fit()` appear to fail when the data (y-values) used as input have dtype `numpy.int32` and the x-values used as input have dtype `numpy.int32`. It doesn't appear to fail if both are `numpy.float32` or both `numpy.int32`.

###### A Minimal, Complete, and Verifiable example
```python
import numpy as np
from lmfit import Model

def my_line(x, a, b):
    return a + b * x

def fit_my_line(xs, ys, a_guess, b_guess):
    l_model = Model(my_line)
    l_model.set_param_hint("a", value=a_guess)
    l_model.set_param_hint("b", value=b_guess)
    l_params = l_model.make_params()
    return l_model.fit(ys, l_params, x=xs)

test_xs = np.arange(0,100, dtype=np.float32)
test_ys = np.asarray(test_xs * 2, dtype=np.int32)

print(fit_my_line(test_xs, test_ys, 3, 5).fit_report())
```

###### Error message:
Not exactly an error, but still unexpected: 
```
[[Model]]
    Model(my_line)
[[Fit Statistics]]
    # fitting method   = leastsq
    # function evals   = 3
    # data points      = 100
    # variables        = 2
    chi-square         = 3045150.00
    reduced chi-square = 31072.9592
    Akaike info crit   = 1036.38905
    Bayesian info crit = 1041.59939
##  Warning: uncertainties could not be estimated:
    a:  at initial value
    b:  at initial value
[[Variables]]
    a:  3.00000000 (init = 3)
    b:  5.00000000 (init = 5)
```

###### Version information
```
Python: 3.7.6 (default, Jan  8 2020, 19:59:22) 
[GCC 7.3.0]

lmfit: 1.0.0, scipy: 1.3.2, numpy: 1.18.1, asteval: 0.9.18, uncertainties: 3.1.2
```

###### Link(s)
None

newville:
@heltluke Yup.  If you had read the FAQ you would have known that the fits work with 64-bit floats:

      The fitting routines accept data arrays that are one-dimensional and double precision.

Please do read the FAQ, and the instructions for raising an issue.

lneuhaus:
[...]

PR #722, opened 2021-04-09
[WIP] assert input data for least squares is of correct type

peendebak:
#### Description

For least squares fitting some odd behaviour happens when the type of the independent data is not float64. Also see the discussion in https://github.com/lmfit/lmfit-py/issues/617. The documentation states that the input data should be float64, but that is easy to miss for new users. Also, adding explicit conversions to float64 around all calls to `lmfit` adds unnecessary additional code.

This PR add a warning if data with incorrect type is supplied to the `least_squares` method.

A minimal example that shows how a simple sine fit can go wrong, without warnings or errors from `lmfit`:
```
import numpy as np

from lmfit import Model
from typing import Union


def sine(x: Union[float, np.ndarray], amplitude: float, frequency: float,
         phase: float, offset: float) -> Union[float, np.ndarray]:
    """ Model for sine function with offset

        y = offset + amplitude * np.sin(2 * np.pi * frequency * x + phase)

    Args:
        x : Independent data points
        amplitude, frequency, phase, offset: Arguments for the sine model
    Returns:
        Calculated data points

    """
    y = amplitude * np.sin(2 * np.pi * frequency * x + phase) + offset
    return y


x_data=np.array([0.        , 0.39130434, 0.7826087 , 1.173913  , 1.5652174 ,
       1.9565217 , 2.347826  , 2.7391305 , 3.1304348 , 3.5217392 ,
       3.9130435 , 4.304348  , 4.695652  , 5.0869565 , 5.478261  ,
       5.869565  , 6.2608695 , 6.652174  , 7.0434785 , 7.4347825 ,
       7.826087  , 8.217391  , 8.608696  , 9.        ], dtype=np.float32)

y_data=np.array([ 5.14749191,  4.73160261,  3.56180872,  2.07035031,  0.08382382,
       -1.81577806, -3.51661368, -4.5393273 , -4.99211742, -4.47738279,
       -3.61221642, -2.08006958, -0.11406613,  1.74609116,  3.46597728,
        4.70999518,  5.09628533,  4.58258861,  3.56415328,  2.0660223 ,
        0.02111145, -1.77618434, -3.27206256, -4.66309001], dtype=np.float64)

initial_parameters=np.array([ 5.06980466,  0.16300641,  1.9141469658, -0.17445164])


lmfit_model = Model(sine)
lmfit_model.set_param_hint('amplitude', min=0)
lmfit_result = lmfit_model.fit(y_data, x=x_data, method='least_squares',
                               **dict(zip(lmfit_model.param_names, initial_parameters)) )

lmfit_result.plot(show_init=True)
plt.title('different dtype: bad fit')
```
For users who the results can be hard to interpret.

###### Type of Changes

- [x] Bug fix
- [ ] New feature

The PR add a check on the type of the independent variable for least squares fitting. The goal is to make the PR backwards compatible, except when the user is supplying data that is invalid and make sure the added overhead is minimal.

###### Tested on

```
Python: 3.8.6 (tags/v3.8.6:db45529, Sep 23 2020, 15:52:53) [MSC v.1927 64 bit (AMD64)]

lmfit: 1.0.2+26.g03daaf8.dirty, scipy: 1.6.1, numpy: 1.20.1, asteval: 0.9.23, uncertainties: 3.1.5
```

###### Verification 
Have you

- [ ] included docstrings that follow PEP 257? No new methods written

- [x] verified that the documentation builds locally? Partly build with `make html`. Fails with error unrelated to this PR

- [ ] added an example? Not required

### DIFF
--- a/doc/whatsnew.rst
+++ b/doc/whatsnew.rst
@@ -27,7 +27,7 @@ Various:
 - remove incorrectly spelled ``DonaichModel`` and ``donaich`` lineshape, deprecated in version 1.0.1 (PR #707)
 - remove occurrences of OrderedDict throughout the code; dict is order-preserving since Python 3.6 (PR #713)
 - update the contributing instructions (PR #718; @martin-majlis)
-
+- add warning if independent data of incorrect type is used in least squares fitting (PR #722)
 
 .. _whatsnew_102_label:
 
--- a/lmfit/minimizer.py
+++ b/lmfit/minimizer.py
@@ -1574,6 +1574,9 @@ class Minimizer:
 
         result.call_kws = kws
         try:
+            if 'x' in self.userkws:
+                if np.asarray(self.userkws['x']).dtype is np.dtype(np.float32):
+                    warnings.warn('input argument x has inccorect type np.float32')
             ret = least_squares(self.__residual, start_vals,
                                 bounds=(lower_bounds, upper_bounds),
                                 kwargs=dict(apply_bounds_transformation=False),


newville:
@peendebak I appreciate the effort and agree that we can try to mitigate the problem.  But I think we can do this better.

Let's avoid the limitation of `scipy.optimize.curve_fit` in requiring independent data to be array data of any type or shape.   We should not care what the user passes into that independent data.  Independent data are, well, independent.   It does not need to be array data at all - it could very well be a dict or other complex structure.   

Anyway, the data type of the independent data is not the problem.  The problem is that the result returned by the model function is not double precision. Let's try to fix that -- coercing the *result* to be double precision.

I think that could just be something like (at the return of `Minimizer.__residual()` in `minimizer.py`):

``` 
#was       return _nan_policy(np.asarray(out).ravel(), nan_policy=self.nan_policy)
      return _nan_policy(np.asarray(out, dtype='float64').ravel(), nan_policy=self.nan_policy)
```

I think that would solve the problem, and more generally and does the coercion to double at the same time as unraveling to 1D.  

Does that make sense to you?

peendebak:
@newville Thanks for the information. I agree that the check on the independent variable is not so good (and will fail if the independent variable is not called `x`).

I did check your suggestion, but it is not working out for me. For the minimal example I provided the output of `__residual` is already `float64`, so the coercing has no effect. The output of `Model._residual` (from `model.py`) is also a `float64`. But inside `Model._residual` the output of `model = self.eval(params, **kwargs)` is a `float32`. I tried casting that to `float64`, but it does not improve the fitting, so the problem is somewhere else. The `self.eval is a direct call to `self.func`, which is my own define `sine` function.

With the above in mind, I see no options for coercing within `lmfit` that will solve the problem. We can add a warning when the result type of `Model.eval` is not `float64`, which will at least alert the user that something might be wrong.

So in `_residual` instead of
```
        model = self.eval(params, **kwargs)
```
we do
```
        model = self.eval(params, **kwargs)
        if model.dtype is np.dtype(np.float32):
            warnings.warn('problem with float32')
```
I was a bit worried that since `_residual` is called multiple times this would add overhead. I did measure the overhead (for my minimal example, using jypyter `%timeit`), but could not find anything significant.

What I don't understand is why for such a simple fitting problem there is a difference between the float32 and float64 results. At first I thought it was because the data is passed to a C implementation that interprets the data incorrectly, but that does not seem to be the case.

One more observation. If I "fix" my datatype by defining
```
def sine(x: Union[float, np.ndarray], amplitude: float, frequency: float,
         phase: float, offset: float) -> Union[float, np.ndarray]:
    y = amplitude * np.sin(2 * np.pi * frequency * x + phase) + offset
    y=np.asarray(y, dtype=np.float64) # cast so the output is of correct type
    return y
```
then the check `if model.dtype is np.dtype(np.float32)` will not generate any warning, but the fitting will still give bad results.

newville:
@peendebak So, in the previous version, you wanted to special case an independent parameter named `x` and its data type.  This version does the same.  This approach is not going to be accepted.

I think you need to start over and demonstrate an issue before trying to fix it.  Like, clearly state what behavior you are seeing and why you think it is a problem. 

For sure, fitting data really requires calculations to be double-precision floating-point.   The most obvious way to make that happen is for the user to ensure that all of the numerical values and input arrays to be double precision.


As it turns out, using `leastsq` and increasing `epsfcn` to `1e-12` will actually fit the problem too:

```python
   lmfit_model = Model(sine)
   lmfit_model.set_param_hint('amplitude', min=0)

   params = lmfit_model.make_params(amplitude=5.0, frequency=0.1, phase=2, offset=0)
   lmfit_result = lmfit_model.fit(y_data, params, x=x_data, fit_kws={'epsfcn':1.e-12})
   print(lmfit_result.fit_report())
   lmfit_result.plot(show_init=True)
   plt.show()
```

newville:
@peendebak just as a follow-up:  We don't actually set `epsfcn` in `Minimizer()`, which leaves it at the rather optimistic value of 2e-16.   Honestly, that is sort of ridiculously low in almost every situation anyway.   

So a reasonable solution might be to set the default value for that (around https://github.com/lmfit/lmfit-py/blob/c448cedbeda1381367141ed4510ea58fef29cef7/lmfit/minimizer.py#L1680)

to have `epsfcn=1.e-7` or so.   For `least_squares`, that would be called `diff_step` (hmm, yeah that deliberate inconsistency in scipy still irritates me, so that's good ;) ).   FWIW, `least_squares` can use fewer function calls, but in my limited experience, it can also go off the rails sometimes.

peendebak:
@newville To address the issue in the minimal example (besides fixing it in the minimal example itself) I see two options. The first option is to generate a warning if single-precision type is used. For that I created #723.
The other option is to change the default behaviour of `lmfit` with respect to the `epsfcn`/`diff_step`. I feel this should be addressed in `scipy` itself. For reference I will include some results below:

For method `least_squares` the value for `diff_step` is determined as `np.finfo(np.float64).eps**0.5=1.4901161193847656e-08`. When setting this manually to a slightly bigger value 1e-7 the minimal example fits correctly. 
For `least_sq` the parameter `epsfcn` is set automatically as
`2.220446049250313e-16`. Setting it manually to 1e-10 (or bigger) fixes the minimal example.
The value of `2.220446049250313e-16` for `least_sq` indeed seems very small (also when comparing to the similar value for `diff_step` in `least_squares).  A related discussion is on https://github.com/scipy/scipy/issues/7117

I will close the PR.

PR #808, opened 2022-10-17
always save function name and importer, load saved function with dill…

newville:
…-encoded data as a last resort, update doc

#### Description

This changes the order of loading of functions in `jsonutils` so that loading with `dill` is the last resort.  That is, loading a function will now always try to use the function name and import module first.  This should do a better job of working across Python versions, and sort of helps to address #792, or at least give a clearer warning message for when loading a dumped Parameters, Model, or ModelResult does not work.

###### Type of Changes

- [x] Bug fix
- [ ] New feature

###### Verification 
Have you

- [ ] added an example?

### DIFF
--- a/doc/model.rst
+++ b/doc/model.rst
@@ -596,11 +596,11 @@ saving a model will always save the *name* of the model function. The
 contain a dictionary of function definitions with the function names as
 keys and function objects as values. If one of the dictionary keys matches
 the saved name, the corresponding function object will be used as the model
-function. If it is not found by ne=ame, and if ``dill`` was used to save
+function. If it is not found by name, and if ``dill`` was used to save
 the model, and if ``dill`` is available at run-time, the ``dill``-encoded
 function will try to be used.  Note that this approach will generally allow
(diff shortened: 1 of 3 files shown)


reneeotten (commented):
thanks @newville - that looks like a good change in de `decode4sj` function (as far as I can tell the `encode4js` doesn't actually change any logic it's just a refactoring - correct?). 

There are probably two typos in the documentation that would benefit from correcting, other than that this looks good!

newville:
@reneeotten Thanks very much - I'll fix the doc typos. 

The logic is changed a little bit.  Currently, if a function has been saved as "dill encoded", it is assumed that it can be restored from that dill encoding.   With this change, restoring a function by name is always tried first, and dill encoding is only used as a backup.   

I guess my view is that dump/load really is not that viable as a robust long-term solution for user-defined functions that are added to `params._asteval` interpreter or even user-defined model functions.  At some point, we really sort of need to be able to get at the Python code for these.

newville:
Ok, merging...

PR #834, opened 2022-11-28
BLD: Remove redundant wheel dep from pyproject.toml

mgorny:
#### Description
Remove the redundant `wheel` dependency, as it is added by the backend automatically.  Listing it explicitly in the documentation was a historical mistake and has been fixed since, see: https://github.com/pypa/setuptools/commit/f7d30a9529378cf69054b5176249e5457aaf640a

###### Type of Changes
- [ ] Bug fix
- [ ] New feature

###### Tested on
n/a

###### Verification
Have you
- [ ] updated the documentation and/or added an entry to the release notes (doc/whatsnew.rst)? [should I add one for such a trivial and not-really-user-facing change?]

--------

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["setuptools>=45", "wheel", "setuptools_scm>=6.2"]
+requires = ["setuptools>=45", "setuptools_scm>=6.2"]
 build-backend = "setuptools.build_meta"
 
 [tool.setuptools_scm]


newville:
@mgorny Huh?  I do not see an actual problem to be solved here. What is wrong with listing redundant dependencies?  

We list both `scipy` and `numpy` as dependencies.  Listing `numpy` is redundant.   But it adds clarity, and we can be specific about versions. 

-1 from me.   I don't see an issue.

mgorny:
The issue is that `wheel` is an implementation detail that may change. If `setuptools` stops using `wheel` (and they were considering that, at least some time ago), then you'd be specifying the wrong dependency. Listing `wheel` here doesn't add any clarity — it only confuses people to think you're actually using it while you're not (`setuptools` is using it, not `lmfit`).

newville:
@mgorny  "If setuptools stops using wheel (and they were considering that, at least some time ago)" seems like a reason to treat it as "not redundant".

newville:
closing.

mgorny:
> @mgorny "If setuptools stops using wheel (and they were considering that, at least some time ago)" seems like a reason to treat it as "not redundant".

If they stop using `wheel`, then `wheel` would not be required to build wheels x_x.

newville:
@mgorny OK, if that happens, then we will have an issue.  If "they" don't, then I don't really see an issue. It's hard to worry about some potential deprecation of features from an upstream package.

PR #841, opened 2023-02-10
Fix #840

### LINKED ISSUE #840: `ConstantModel` returns wrong shape
I am guessing people very rarely use the built-in `ConstantModel` without combining it with another model, which is probably why this hasn't cropped up before (when combined broadcasting rules would sort this out, I suspect), but `ConstantModel` returns only a single element rather than matching the shape of the input data. This causes the following to fail:

```python
import numpy as np
from lmfit.models import ConstantModel

x = np.array([1.0, 2.0, 3.0])
y = np.array([1.0, 1.5, 0.5])
model = ConstantModel()
params = model.guess(y, x=x)
fit = model.fit(y, params, x=x)

plt.plot(x, y, "o")
plt.plot(x, fit.best_fit, "--")
```

This raises: `ValueError: x and y must have same first dimension, but have shapes (3,) and (1,)` since `fit.best_fit` has shape `(1,)`.

JeppeKlitgaard:
#### Description
Fixes: #840

###### Type of Changes

- [X] Bug fix

Skipped most steps out of laziness since this is a tiny fix

### DIFF
--- a/lmfit/models.py
+++ b/lmfit/models.py
@@ -168,7 +168,7 @@ class ConstantModel(Model):
                        'independent_vars': independent_vars})
 
         def constant(x, c=0.0):
-            return c
+            return c * np.ones(np.shape(x))
         super().__init__(constant, **kwargs)
 
     def guess(self, data, x=None, **kwargs):
@@ -198,7 +198,7 @@ class ComplexConstantModel(Model):
                        'independent_vars': independent_vars})
 
         def constant(x, re=0., im=0.):
-            return re + 1j*im
+            return (re + 1j*im) * np.ones(np.shape(x))
         super().__init__(constant, **kwargs)
 
     def guess(self, data, x=None, **kwargs):


newville:
@JeppeKlitgaard What if `x` does not have a shape attribute?

JeppeKlitgaard:
@newville I believe I came around this by using `np.shape` rather than relying on the attribute. This relaxes the requirement to `x` being a Numpy `arraylike`, which I believe it should be anyway for use in `lmfit`

newville:
@JeppeKlitgaard I think it could be ok to expect that `x` is array-like here.  While we actually do not demand that in general or for user-defined model functions, most (perhaps all) of the other built-in models do use numpy/scipy ufuncs that will also expect `x` to be array-like.

I think you are right that using `np.shape(x)` vs `x.shape` is a good idea, permitting more types and defaulting to a length-1 array for many cases.  I might suggest having

```
try:
    c = c * np.ones(np.shape(x))
except TypeError:
    pass
return c
```

newville:
@JeppeKlitgaard It would also be good to apply the same approach to the ComplexConstantModel.

newville:
@JeppeKlitgaard It would also be good to apply the same approach to the ComplexConstantModel.

JeppeKlitgaard:
@newville 

Sorry for not getting back to you!

My personal preference would definitely be to only have a single return type that is always reliably arraylike – I am fairly sure there are bits of `lmfit` that rely on this behaviour as well. For example, the following raises a `TypeError` when the output of a model is not an array of shape matching the independent variable array:

```python
x = [1,2,3]
y = [1,2,3]

model = ConstantModel()
params = model.make_params()
fit = model.fit(y, params, x=x)

fit.plot()
```

This suggests to me that:
```python
return c * np.ones(np.shape(x))
```

is preferable over:
```python
try:
    c = c * np.ones(np.shape(x))
except TypeError:
    pass
return c
```

JeppeKlitgaard:
@newville 
I have updated the PR to also include the same fix for `ComplexConstantModel` as you suggested.

newville:
@JeppeKlitgaard OK, I will merge this and then fix it myself.   
We cannot promise that `plot` will work for all cases, but it would work with your example and the try/except block.   

But in case `np.ones(np.shape(x))` fails with a `TypeError`, it would help to return something sensible instead of failing.
And, yeah, almost everyone will pass in something nd-array-like, and all of the builtin models will expect that numpy/scipy ufuncs will work on `x`, but we actually do not require that independent variables be nd-array-like.

PR #843, opened 2023-02-19
feat: add ODR solver

JeppeKlitgaard:
This adds an ODR solver using `scipy.odr`. This is WIP and because of the very different approach used in `scipy.odr` to the other scipy minimizers it requires access the fitting data in a different way which is currently 'hacked' around.

I probably won't have time to finish this work anytime soon, but wanted to leave it here for myself or others to pick up in the future. Adding ODR support is surprisingly simple.

Nothing below this line has been filled out just yet – this is a draft!
---

###### Type of Changes

- [ ] Bug fix
- [ ] New feature

###### Verification 
Have you

- [ ] added an example?

### DIFF
--- a/lmfit/minimizer.py
+++ b/lmfit/minimizer.py
@@ -20,8 +20,10 @@ import inspect
 import multiprocessing
 import numbers
 import warnings
+from typing import Any, Literal
 
 import numpy as np
+import numpy.typing as npt
 from scipy import __version__ as scipy_version
 from scipy.linalg import LinAlgError, inv
 from scipy.optimize import basinhopping as scipy_basinhopping
@@ -36,6 +38,7 @@ from scipy.sparse import issparse
 from scipy.sparse.linalg import LinearOperator
 from scipy.stats import cauchy as cauchy_dist
 from scipy.stats import norm as norm_dist
+from scipy import odr
 import uncertainties
 
 from ._ampgo import ampgo
@@ -842,8 +845,6 @@ class Minimizer:
         if self.scale_covar:
             self.result.covar *= self.result.redchi
 
-        vbest = np.atleast_1d([self.result.params[name].value for name in
-                               self.result.var_names])
 
         has_expr = False
         for par in self.result.params.values():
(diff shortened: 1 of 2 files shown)


JeppeKlitgaard:
@newville Let me know your thoughts on this. I don't have a lot of time to commit towards this, but I think it could be a valuable addition.

Since this was primarily a dirty hack for my own use-case, it doesn't fit with older Pythons in its current form. This could be amended easily.

newville:
@JeppeKlitgaard  You call this a draft and a dirty hack and have no tests, examples, or docs. Your code fails with import errors in our tests. From a quick look, I would be cautious about the typing here, and the black-like function signature does not match our coding style.

But you also say that you do not have a lot of time to work on it. Are you expecting someone else to add docs, tests, and examples and then support this code?

JeppeKlitgaard:
I was more wondering whether you would be open to adding ODR support or not – it should be noted that it requires a bit of a work-around since the ODR interface requires access to the data rather than just minimizing a function.

I realise this isn't done and lacks tests, docs, etc – hence being marked as a draft.

The type hinting was primarily added for my own benefit during development and can be removed later if need be. I would potentially be open to working on this PR at a later point if there is appetite for it

newville:
> I was more wondering whether you would be open to adding ODR support or not – it should be noted that it requires a bit of a work-around since the ODR interface requires access to the data rather than just minimizing a function.

I don't think anyone would object to adding support for ODR to lmfit.  Doing so seems like a fair amount of work.  I have not looked closely at the changes you have made, and if this is meant to be a "draft", then I am not sure about your commitment to this effort.

> I realise this isn't done and lacks tests, docs, etc – hence being marked as a draft.

Well, you could have asked a question about this on the mailing list, or on the Github discussions page (or commented on the related discussion at https://github.com/lmfit/lmfit-py/discussions/789).  But, you created a PR - a proposed addition to the code and presumably a proposed solution to a problem (often, but not always, explicitly stated in an Issue).   There is sort of a normal workflow to these things.  A random dude on the internet tossing some unfinished, undocumented, untested code, and saying "here you go, I don't have time to work on this" is not our preferred mode of working.

> The type hinting was primarily added for my own benefit during development and can be removed later if need be. I would potentially be open to working on this PR at a later point if there is appetite for it

I don't have a strong opinion on type hinting in general: it adds value but it is more code to maintain and test, and think about. In this specific case, I am unsure whether the type hints are actually correct.   The black-like code formatting will need to change.

newville:
@JeppeKlitgaard @reneeotten We are 4 months from this initial, non-working draft which then got no attention. 

I think we should close PRs that go several months without activity or discussion.

JeppeKlitgaard:
> @JeppeKlitgaard @reneeotten We are 4 months from this initial, non-working draft which then got no attention.
> 
> I think we should close PRs that go several months without activity or discussion.

I agree. I just finished exams and may get around to having another look at this, but I have closed the PR for now.

Issue #911, opened 2023-08-15
`result.params.uvars` not correctly created with Minimize


newville:
#### Description

Using the `minimize` / objective function interface, `result.params.uvars` is not created, as described, even when the covariance matrix is available.  In fact, `uvars` are calculated, but incorrectly placed in `result.uvars`

###### A Minimal, Complete, and Verifiable example

```python
from numpy import exp, linspace, pi, random, sign, sin
from lmfit import create_params, fit_report, minimize
p_true = create_params(amp=14.0, period=5.46, shift=0.123, decay=0.032)
def residual(pars, x, data=None):
    """Model a decaying sine wave and subtract data."""
    vals = pars.valuesdict()
    amp = vals['amp']
    per = vals['period']
    shift = vals['shift']
    decay = vals['decay']

    if abs(shift) > pi/2:
        shift = shift - sign(shift)*pi
    model = amp * sin(shift + x/per) * exp(-x*x*decay*decay)
    if data is None:
        return model
    return model - data

random.seed(0)
x = linspace(0.0, 250., 1001)
noise = random.normal(scale=0.7215, size=x.size)
data = residual(p_true, x) + noise

fit_params = create_params(amp=13, period=2, shift=0, decay=0.02)
out = minimize(residual, fit_params, args=(x,), kws={'data': data})
print(fit_report(out))

# this will fail:
print(out.params.uvars)

```

###### Fit report:

```
[[Fit Statistics]]
    # fitting method   = leastsq
    # function evals   = 83
    # data points      = 1001
    # variables        = 4
    chi-square         = 498.811759
    reduced chi-square = 0.50031270
    Akaike info crit   = -689.222517
    Bayesian info crit = -669.587497
[[Variables]]
    amp:     13.9121959 +/- 0.14120321 (1.01%) (init = 13)
    period:  5.48507038 +/- 0.02666520 (0.49%) (init = 2)
    shift:   0.16203673 +/- 0.01405662 (8.67%) (init = 0)
    decay:   0.03264539 +/- 3.8015e-04 (1.16%) (init = 0.02)
[[Correlations]] (unreported correlations are < 0.100)
    C(period, shift) = +0.7974
    C(amp, decay)    = +0.5816
    C(amp, shift)    = -0.2966
    C(amp, period)   = -0.2432
    C(shift, decay)  = -0.1819
    C(period, decay) = -0.1496
```

###### Error message:

```
Traceback (most recent call last):
[...]

PR #912, opened 2023-08-15
ensure result.params has uvars attribute

newville:
#### Description

As at #911, this makes the `uvars` calculated with `Parameters.create_uvars` be an attribute of the `Parameters` instance after a fit.

###### Type of Changes

- [x] Bug fix
- [ ] New feature

###### Verification 
Have you

- [ ] added an example?

(example updated)


newville:
will replace with PR better testing and documenting the existing behavior (which probably does not need to change).

PR #926, opened 2023-11-18
Fix problems with arguments that are neither fit parameters nor independent variables

### LINKED ISSUE #917: Inconsistencies with non-parameter model function arguments
#### Description
I have a model function that contains a keyword argument that is not part of the fit parameters (but not a compulsory independent variable either). My expectation was that I can just pass it as a keyword argument to the model's `fit()` method when I need to set it. Here's an example with `s` being the argument in question.
```python
import lmfit
import numpy as np
from matplotlib import pyplot as plt

# Create a model.
def foo(x, a=3, b=5, s=1):
    return a + s * b * x**2
mdl = lmfit.Model(foo, param_names=['a', 'b'])
mdl.set_param_hint('a', value=3)
mdl.set_param_hint('b', value=5)

# Synthesize some noisy data.
x = np.linspace(-5, 5, 101)
s = 1 + 0.7 * np.sin(x)
y = foo(x, 2.3, 1.7, s) + np.random.normal(0, 1, x.shape)

# Fit the data and report the result
fitres = mdl.fit(y, x=x, s=s)
print(fitres.fit_report())
fitres.plot()
plt.show()

# Manually check the result.
plt.plot(x, y, '.', label='data')
plt.plot(x, fitres.eval(x=x), label='fit result eval()')
plt.plot(x, foo(x, a=fitres.params['a'], b=fitres.params['b'], s=s), label='manual check')
plt.legend()
plt.show()
```

Passing `s` to `fit()` like this results a warning.
```
C:\Users\wilde\miniconda3\lib\site-packages\lmfit\model.py:1053: UserWarning: The keyword argument s does not match any arguments of the model function. It will be ignored.
```
However, the fit still seems to perform as expected, i.e. the argument is passed on to the fitting function, the resulting parameters are correct. Here's the fit report (`a = 2.3` and `b = 1.7` were used during data synthesis).
```
[[Model]]
    Model(foo)
[[Fit Statistics]]
    # fitting method   = leastsq
    # function evals   = 7
    # data points      = 101
    # variables        = 2
    chi-square         = 109.127855
    reduced chi-square = 1.10230157
    Akaike info crit   = 11.8173660
    Bayesian info crit = 17.0476071
    R-squared          = 0.99636556
[[Variables]]
    a:  2.13258073 +/- 0.13627980 (6.39%) (init = 3)
    b:  1.69611547 +/- 0.01029550 (0.61%) (init = 5)
[[Correlations]] (unreported correlations are < 0.100)
    C(a, b) = -0.6421
```

Adding to the confusion, the fit result's `plot()` function draws the correct residuals and data but an incorrect best fit, i.e. it really does ignore `s` here.
image
The `eval()` method on the other hand seems to respect `s`.
image

###### Version information

```
Python: 3.9.18 | packaged by conda-forge | (main, Aug 30 2023, 03:40:31) [MSC v.1929 64 bit (AMD64)]

lmfit: 1.2.2, scipy: 1.11.3, numpy: 1.26.0,asteval: 0.9.31, uncertainties: 3.1.7
```

schtandard:
#### Description

Model function arguments that are neither fit parameters nor independent variables were handled incorrect in two places.
- They were incorrectly identified to not correspond to a model function argument in `Model.fit()` and thus a misleading spurious warning was issued.
- They were not considered in `ModelResult.plot_fit()`, causing `ModelResult.plot()` to produce contradictory plots.

This PR remedies both issues and fixes #917 (moved to #920).

###### Type of Changes

- [x] Bug fix
- [ ] New feature

###### Tested on

```
Python: 3.11.6 | packaged by conda-forge | (main, Oct  3 2023, 10:29:11) [MSC v.1935 64 bit (AMD64)]

lmfit: 0.0.post2796+g4e2470a.d20231118, scipy: 1.11.3, numpy: 1.26.0, asteval: 0.9.31, uncertainties: 3.1.7
```

###### Verification 
Have you

  (No API changes.)

  (I get several warnings and three skipped tests, though.)

  (Except for issues that also appeared when building on `master`, as described in #924.)

  (No tests were affected by the change.)
  (I did not add anything to `doc/whatsnew.rst` yet. I had a look at the file history and other PRs and it seems to be done in batch, separate from the PRs. I will add an entry if you ask me to.)
- [ ] added an example?
  (Same as for test, I don't think any example was affected and I did not add a new one.)

### DIFF
--- a/lmfit/model.py
+++ b/lmfit/model.py
@@ -1047,9 +1047,9 @@ class Model:
                 params[name].set(value=p)
             del kwargs[name]
 
-        # All remaining kwargs should correspond to independent variables.
+        # Check for spurious kwargs.
         for name in kwargs:
-            if name not in self.independent_vars:
+            if name not in self._func_allargs:
                 warnings.warn(f"The keyword argument {name} does not " +
                               "match any arguments of the model function. " +
                               "It will be ignored.", UserWarning)
--- a/lmfit/model.py
+++ b/lmfit/model.py
@@ -2079,7 +2079,7 @@ class ModelResult(Minimizer):
             ax.plot(x_array, reduce_complex(self.data),
                     datafmt, label='data', **data_kws)
 
-        y_eval = self.model.eval(self.params, **{independent_var: x_array_dense})
+        y_eval = self.eval(self.params, **{independent_var: x_array_dense})
         if isinstance(self.model, (lmfit.models.ConstantModel,
                                    lmfit.models.ComplexConstantModel)):
             y_eval *= np.ones(x_array_dense.size)


newville:
@schtandard  I think that arguments of a Model function are either parameters or independent variables.   

We generally prefer a process of agreeing on what an issue is and maybe even a description of a solution to that issue before a PR.   Some cases are easy to identify as bugs, sometimes experienced developers see a problem and know how to fix it (but sometimes I get this wrong ;)). 

I am having a hard time understanding what the actual problem is that you keep misidentifying as something else.  Basically, we have identified an Issue to be solved.
  
It's possible that "ensure that independent variables can be left unspecified in `Model.fit()` and `Model.eval()` if default values are given in the signature" would be a fix that would satisfy what you claim to need and be agreeable to us.     Maybe that is what you are trying to say?  I am not sure.

schtandard:
> We generally prefer a process of agreeing on what an issue is and maybe even a description of a solution to that issue before a PR.

I prefer that, too. I did not get the feeling that I was getting the issue across, though. I decided to just take a look at the code and when I discovered that the problem seemed to be isolated to two lines of code, I thought that maybe a PR would help communicate the issue. This doesn't seem to have been a total success, but I haven't given up hope just yet.

> I think that arguments of a Model function are either parameters or independent variables.

Ok, then this is surely a big part of the confusion. From my point of view, your documentation and the module behavior disagree. Let's look at the example from [your post](https://github.com/lmfit/lmfit-py/discussions/920#discussioncomment-7601472).
```python
def func(x, a=3, b=5, scale=None):
    output = a +x*b
    if scale is not None:  output *= scale
    return output
mod = Model(func)
```
If we ask the model what its independent variables and parameters are
```python
print("independent variables:", mod.independent_vars)
print("parameters:", mod.param_names)
```
we get
```
independent variables: ['x']
parameters: ['a', 'b']
```
Apparently, the model thinks that `scale` is neither.

`model.rst` contains the helpful definition

> independent variable
>
>> A function argument that is not a parameter or otherwise part of the
>> model, and that will be required to be explicitly provided as a
>> keyword argument for each fit with `Model.fit()` or evaluation
>> with `Model.eval()`.

This is both in direct contradiction to your statement "You have repeatedly said that all independent variables must always be specified but that is simply untrue." and contains a reference to function arguments that are "otherwise part of the model" (i.e. not as a parameter or an independent variable).

And then there is `Model.opts` which is documented as

> Extra keyword arguments to pass to model function. [...]

This is where `scale` goes in the example above; not a parameter, not an independent variable but an "extra" keyword argument. This is what's not handled correctly by `Model.fit()` (causing a spurious warning) and `ModelResult.plot_fit()` (being ignored there).

Does this help clarify things?

newville:
@schtandard  Well, I guess I think we should probably make sure that function arguments can only be Parameters or independent variables.  Admittedly, the concept of "independent variable" has evolved here from "first argument, must be 1D ndarry" (as I think is still true in scipy `curve_fit`) to "all the non-Parameter data needed to calculate the model, without limitation on data type".    

So,  I think we shoud assert that independent variables with a default value in the function signature should not be required in `Model.fit()` or `Model.eval()` (or any others I might be forgetting).  And, when there is a default value, that should be used.  

I think that such a fix would probably address most of the issues you have faced.

schtandard:
That sounds like a good idea, though I cannot tell how many things would need changing for it. You probably have a much better overview than me there. Some comments:

- I guess that would mean dropping `Model.opts` and (similar attributes in other classes) storing "extra" keyword arguments?
- An attribute containing only the independent variables with no default arguments would then be useful, so users can check which arguments _must_ be supplied.
- 27db25b7b8f4c1c63db4093449769bcc71e1fd8f should then be obsolete, though bd9b9f1c55f94fab38fa513c63a1ae62661be607 would remain the correct fix for the `ModelResult.plot_fit()` issue, I think.
- Actually, this would mean that the `param_names` and `independent_vars` arguments to `Model.__init__()` are equivalent (being complementary), no? Keeping both for convenience and (limited) backwards compatibility might be good, but supplying both without naming all model function arguments should then raise an error, right?
- Or maybe it would be favorable to have `add_params` and `add_independents` (or something) for naming only those arguments that should fall in the category they wouldn't got to automatically? Would have the advantage that the user knows they have to provide initial values for exactly the parameters in `add_params` (as the others have automatic initial values) but be somewhat less explicit, so maybe not.
- The rule for automatically determining independent variables would need to change to "anything that isn't a keyword argument with a numerical default value" (i.e. not a parameter), of course. Two possible cases warrant special consideration:
  - No independent variables (all arguments have numerical default values). Seems unlikely to be used, but not inconceivable. One could argue for the first argument to be used unless there is an explicit `independent_vars=[]` but I would probably leave it for consistency.
  - No parameters (no arguments have numerical default values). This means there is nothing to fit, though it may still be useful for composite models? Don't know.
- The essence of #925 should be unaffected by this change, as it does not deal with independent variables at all (neither the old nor the new kind, it deals with parameters). The code surrounding it would need to be slightly adapted, of course (as using `independent_vars` should already disable the automatic detection of parameter names, just like `param_names` does now).

newville:
@schtandard I believe this is now better (if not 100% fixed) with the merging of #941

Issue #931, opened 2023-12-14
ModelResult loads does not work with user defined parameter expressions


cassberk:
#### First Time Issue Code

Yes, I read the instructions and I am sure this is a GitHub Issue.

#### Description

Using the `dumps` and `loads` methods on the `ModelResult` class does not properly take into account loading user added functional parameter expressions. The `funcdefs` keyword in `loads` enables users to pass functions to add to the the asteval interpreter, but these are not subsequently passed into `Parameters()` where they are actually added to the asteval symtable. This caused deserialization to not work with `ModelResults` that have parameters with user defined functions as expressions.

###### A Minimal, Complete, and Verifiable example

```python
import json
import numpy as np
import lmfit as lm

def model_function(x,a):
    
    return x*a

def myfun(x):
    return x**3

model = lm.Model(model_function, name="somefunc")
params = lm.Parameters(usersyms={myfun.__name__: myfun})
params.add("a", value=2)

params.add("b", expr="myfun(a)")


xx = np.linspace(0, 1, 10)
yy = 3 * xx


model_result = model.fit(
    yy,
    params=params,
    x=xx,
)


params = lm.Parameters()
model_result_load = lm.model.ModelResult(lm.Model(model_function), params)

funcdefs = {"model_function": model_function, myfun.__name__: myfun}

model_result_load.loads(model_result.dumps(), funcdefs=funcdefs)
```

###### Fit report:

This is not an issue with a fit routine.

###### Error message:

```
TypeError
   <ast.Module object at 0x13eb78280>
'{'__class__': 'Callable', '__name__': 'myfun', 'pyversion': '3.9', 'value': None, 'importer': '__mp_main__'}' is not callable!!
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
~/repos/treeline/qpu-analysis/.venv/lib/python3.9/site-packages/IPython/core/formatters.py in __call__(self, obj)
    343             method = get_real_method(obj, self.print_method)
    344             if method is not None:
--> 345                 return method()
    346             return None
    347         else:

~/repos/treeline/qpu-analysis/.venv/lib/python3.9/site-packages/lmfit/model.py in _repr_html_(self, show_correl, min_correl)
   1723     def _repr_html_(self, show_correl=True, min_correl=0.1):
   1724         """Return a HTML representation of parameters data."""
-> 1725         report = fitreport_html_table(self, show_correl=show_correl,
   1726                                       min_correl=min_correl)
[...]

PR #932, opened 2023-12-14
propagate funcdefs to Parameters in model.loads

cassberk:
#### Description
From ISSUE # #931

Using the `dumps` and `loads` methods on the `ModelResult` class does not properly take into account loading user added functional parameter expressions. The `funcdefs` keyword in `loads` enables users to pass functions to add to the the asteval interpreter, but these are not subsequently passed into `Parameters()` where they are actually added to the asteval symtable. This caused deserialization to not work with `ModelResults` that have parameters with user defined functions as expressions.

###### Type of Changes

- [ x] Bug fix
- [ ] New feature

###### Tested on

Python: 3.9.7 (default, Oct 15 2021, 14:20:07) 
[Clang 13.0.0 (clang-1300.0.29.3)]

lmfit: 1.2.2, scipy: 1.11.1, numpy: 1.23.5,asteval: 0.9.31, uncertainties: 3.1.7

###### Verification 
Have you

- [x ] verified that the documentation builds locally?

- [ ] added an example? N/A

### DIFF
--- a/lmfit/model.py
+++ b/lmfit/model.py
@@ -1891,7 +1891,7 @@ class ModelResult(Minimizer):
                 _par = Parameter(name='')
                 _par.__setstate__(parstate)
                 state['params'].append(_par)
-            _params = Parameters()
+            _params = Parameters(usersyms=funcdefs)
             _params.__setstate__(state)
             setattr(self, target, _params)
 
--- a/lmfit/model.py
+++ b/lmfit/model.py
@@ -1861,7 +1861,7 @@ class ModelResult(Minimizer):
         funcdefs : dict, optional
             Dictionary of custom function names and definitions.
         **kws : optional
-            Keyword arguments that are passed to `json.dumps`.
+            Keyword arguments that are passed to `json.loads`.
 
         Returns
         -------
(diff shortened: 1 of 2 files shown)


newville:
@cassberk Thanks - I think this might be the right solution, but as said in #931, I think I want to spend a little time investigating the history of this...  I feel like I am forgetting something.

Not ignoring, but sort of want to be a bit cautious...

lmfit/model.py
```diff
@@ -1885,13 +1885,14 @@ def loads(self, s, funcdefs=None, **kws):
         self.model = _buildmodel(decode4js(modres['model']), funcdefs=funcdefs)
 
         # params
+        funcdefs.pop(self.model.func.__name__)  # Remove model function to not pass into _asteval
```
cassberk, line 1888: This may not be the most elegant means to do this, but something I noticed was that if the model function is not removed from the funcdefs argument, it gets added to the `_asteval.symtable`. Although a small change, undesired changes to the objects through de/serialization should be minimized, so I think it should be handled somehow.

newville:
@cassberk thanks -- I'm hoping to get to this before the end of the year/

newville:
@cassberk thanks -- I am going to merge this for the test of the doc fix.  But, I think I have a better fix for the actual code that I will try to add shortly....

Issue #951, opened 2024-04-12
`ModelResult.eval_uncertainty` returns a TypError with scalars


jcjaskula-aws:
#### Description
`ModelResult.eval_uncertainty` returns a TypError when providing a singleton to the independent variables.  

###### A Minimal, Complete, and Verifiable example
```
from lmfit.models import LinearModel
import numpy as np
mr=LinearModel().fit(data=[1,2,4], x=[1,2,3])
mr.eval_uncertainty(x=1.2)  # error
mr.eval_uncertainty(x=np.array(1.2)) # error
```

`mr.eval_uncertainty(x=np.array([1.2]))` or `mr.eval_uncertainty(x=np.array([0, 1.2]))` do not throw any error.

###### Fit report:
N/A

###### Error message:

```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[5], line 1
----> 1 mr.eval_uncertainty(x=1.2)

File ~/Documents/Devel/Quantum/External-repos/Misc/lmfit-py/lmfit/model.py:1689, in ModelResult.eval_uncertainty(self, params, sigma, dscale, **kwargs)
   1687 # ensure fjac and df2 are correct size if independent var updated by kwargs
   1688 feval = self.model.eval(params, **userkws)
-> 1689 ndata = len(feval.view('float64'))        # allows feval to be complex
   1690 covar = self.covar
   1691 if any(p.stderr is None for p in params.values()):

TypeError: len() of unsized object
```

###### Version information
Python: 3.10.14 (main, Mar 19 2024, 21:46:16) [Clang 15.0.0 (clang-1500.3.9.4)]

lmfit: 1.3.0.post2+gcf2626b7, scipy: 1.11.4, numpy: 1.26.3,asteval: 0.9.32, uncertainties: 3.1.6

newville:
@jcjaskula-aws  length-1 arrays or scalars where arrays are expected are a perpetual font of errors.   

Probably replacing that. ```ndata = len(feval.view('float64')) ``` with ```ndata = feval.view('float64').size ``` would do the trick.

PR #952, opened 2024-04-12
transform 0-d arrays into 1-d to check their length

jcjaskula-aws:
#### Description
Closes: #951.
The solution implemented here is to transform 0-d arrays into 1-d using np.at_least_1d as recommended here: https://stackoverflow.com/a/4565910. 

###### Type of Changes

- [x] Bug fix
- [ ] New feature

###### Tested on
Python: 3.10.14 (main, Mar 19 2024, 21:46:16) [Clang 15.0.0 (clang-1500.3.9.4)]

lmfit: 1.3.0.post2+gcf2626b7, scipy: 1.11.4, numpy: 1.26.3,asteval: 0.9.32, uncertainties: 3.1.6

###### Verification 
Have you

- [ ] added an example?

### DIFF
--- a/lmfit/model.py
+++ b/lmfit/model.py
@@ -1686,18 +1686,18 @@ class ModelResult(Minimizer):
         nvarys = self.nvarys
         # ensure fjac and df2 are correct size if independent var updated by kwargs
         feval = self.model.eval(params, **userkws)
-        ndata = len(feval.view('float64'))        # allows feval to be complex
+        ndata = np.atleast_1d(feval).view('float64').shape  # allows feval to be complex
         covar = self.covar
         if any(p.stderr is None for p in params.values()):
             return np.zeros(ndata)
 
         # '0' would be an invalid prefix, here signifying 'Full'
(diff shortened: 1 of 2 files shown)


newville:
@jcjaskula-aws  Why not `feval.view('float64').size`?

newville:
Hmm, If the value returrned by `Model.eval` is not strictly 1D, and if len() and size don't agree, I think the calculation here will fail.  Indeed, in `minimize()` the residual array is forced to be unravelled to strictly 1D.   

Can you provide a test case where `.size` fails and where `at_least_1d` passes?

newville:
@jcjaskula-aws  Well, it seems that you have ignored all of my earlier reviews and suggestions so asking for a review seems sort of odd.  Perhaps you were not able to see the earlier reviews?    Otherwise, I cannot tell what is going on.

Seriously, use `ndata = feval.view('float64').ravel().size`.

newville:
@jcjaskula-aws OK, I'm going to merge this.  Thanks for making those changes!

PR #964, opened 2024-08-08
preserve Parameter subclass  when copying Parameters

imnotgonnasithereandtryallday:
#### Description
If an instance of a type derived from `Parameter` is added to a `Parameters` instance, 
and the `Parameters` instance is then copied, the copy contains a parameter of type `Parameter`.
This PR changes the contained parameter type to the derived one.

I moved the code copying `Parameter`  instances to `Parameter.__deepcopy__` so that the derived class can adds attributes.
This should not break any existing code unless someone uses `__deepcopy__` for something else than copying.

###### Type of Changes
- [x] Bug fix

###### Tested on
Python: 3.9.13
lmfit: 0.0.post2850+gb250ed7.d20240807, 
scipy: 1.13.1, 
numpy: 2.0.0, 
asteval: 1.0.2, 
uncertainties: 3.2.2

###### Verification
    No, I have not opened nor found a relevant Issue.
    Tests do not pass, but the result is the same as on the master branch (the added test passes).
    FAILED tests/test_parameters.py::test_check_ast_errors - Failed: DID NOT RAISE <class 'NameError'>
    1 failed, 650 passed, 12 skipped, 6 warnings in 127.66s (0:02:07)
    Does not build:
    Extension error:
    Could not import extension sphinx.builders.epub3 (exception: cannot import name 'jsonimpl' from partially initialized
    module 'sphinxcontrib.serializinghtml' (most likely due to a circular import) 

### DIFF
--- a/lmfit/parameter.py
+++ b/lmfit/parameter.py
@@ -127,17 +127,7 @@ class Parameters(dict):
         parameter_list = []
         for key, par in self.items():
             if isinstance(par, Parameter):
-                param = Parameter(name=par.name,
-                                  value=par.value,
-                                  min=par.min,
-                                  max=par.max)
-                param.vary = par.vary
-                param.brute_step = par.brute_step
-                param.stderr = par.stderr
-                param.correl = deepcopy(par.correl)
-                param.init_value = par.init_value
-                param.expr = par.expr
-                param.user_data = deepcopy(par.user_data)
+                param = deepcopy(par)
                 parameter_list.append(param)
 
         _pars.add_many(*parameter_list)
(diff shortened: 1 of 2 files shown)


newville:
@imnotgonnasithereandtryallday what problem does this solve?  

Do we say somewhere that we support using objects derived from Parameter?   Parameter has a generic `user_data` attribute for adding data. Is there a compelling use case for supporting more than that?

I have not looked at the test failures.  But version 1.3.2 does pass all tests.

newville:
@imnotgonnasithereandtryallday @reneeotten Having heard no motivation for this PR or response to the test failures here in two months I am closing it.

PR #908, opened 2023-08-05
lmfit.Minimizer.least_squares: Bug fix

PhillNo:
#### Description
lmfit.Minimizer.leastsq successfully passes a user Jacobian function to Scipy, while lmfit.Minimizer.least_squares hits an exception. See stack trace in link below.

The discrepancy is apparently due to leastsq passing lmfit.Minimizer.__jacobian to scipy.optimize.leastsq, while lmfit.Minimizer.least_squares passes the unmodified user Jacobian.

(Note that scipy.optimize.leastsq and scipy.optimize.least_squares respectively refer to user Jacobians as 'Dfun' and 'jac'.)

lmfit.Minimizer.__jacobian was also modified to accept unused args. Because Scipy will pass the same args to the user function and Jacobian, and the apply_bounds_transformation kwarg is unnecessary to __jacobian (and this kwarg is not passed to Scipy by leastsq), 

https://groups.google.com/g/lmfit-py/c/dTgC_hn-lsw/m/RjsshTLgBwAJ?utm_medium=email&utm_source=footer&pli=1

###### Type of Changes

- [x] Bug fix
- [ ] New feature
- [ ] Refactoring / maintenance
- [ ] Documentation / examples

###### Tested on

Python: 3.11.4 | packaged by conda-forge | (main, Jun 10 2023, 18:08:17) [GCC 12.2.0]

lmfit: 0.0.post2787+gfc3f7f8, scipy: 1.11.1, numpy: 1.25.2, asteval: 0.9.31, uncertainties: 3.1.7

###### Verification 
Have you

- [x] referenced existing Issue and/or provided relevant link to mailing list?

- [x] verified that existing tests pass locally?

- [x] squashed/minimized your commits and written descriptive commit messages?

- [x] added or updated existing tests to cover the changes?
- [ ] updated the documentation and/or added an entry to the release notes (doc/whatsnew.rst)?
- [x] added an example?

PR #234, opened 2015-06-29
scale analytic jacobian for bounded variables

newville:
This PR scales the gradients calculated by user-supplied jacobian functions for bounded variables, and fixes #233.

PR #973, opened 2024-11-23
Fix issues related to providing a Jacobian function

vyrjana:
#### Description

I encountered some issues while attempting to provide a function to calculate the Jacobian matrix analytically. The issue has been brought up in the mailing list at least a couple of times ([1](https://groups.google.com/g/lmfit-py/c/xwzop5UV82s/m/B-4CexH6EQAJ), [2](https://groups.google.com/g/lmfit-py/c/dTgC_hn-lsw/m/KD_z-SnmBwAJ)). See [PR 908](https://github.com/lmfit/lmfit-py/pull/908) for an earlier pull request (PR) that was ultimately not merged.

The main changes are:

- The `jac` (or `Dfun`) function provided by the user is now wrapped when calling `Minimizer.least_squares`.
  - The signature of `Minimizer.__jacobian` now matches the signature of `Minimizer.__residual`, which fixes a bug where an exception was raised due to `Minimizer.__jacobian` receiving an unexpected `apply_bounds_transformation` parameter. The `apply_bounds_transformation` parameter is taken into account in the body of the `Minimizer.__jacobian` method similarly to `Minimizer.__residual`. This means that the bounds transformation introduced in [PR 234](https://github.com/lmfit/lmfit-py/pull/234) is not applied when `apply_bounds_transformation=False` (e.g., when `Minimizer.least_squares` is called).
  - As a result, the Jacobian function is now provided a `Parameters` object instead of an array of float values. This means that the Jacobian function will receive the same arguments as when providing a Jacobian function to `Minimizer.leastsq` and `Minimizer.scalar_minimize`. **Note, that this should be regarded as a bug fix, but some users may have code that relies on the previous behavior. Thus, this could also be regarded as a breaking change and incrementing the major version number may be warranted.**
- Renamed `Minimizer.__jacobian` to `Minimizer._jacobian` to avoid a bug where an exception is raised when pickling and unpickling private methods using the `ForkingPickler` class that is used by the `multiprocessing` module.
  - See this [long-standing issue in CPython](https://github.com/python/cpython/issues/77188) for more information. There is a pull request to fix the root cause, but it has not been merged (at least not as of the time of writing this).

Some ancillary changes were required to facilitate the changes listed above:

- The `test_least_squares_jacobian_types` in `tests/test_least_squares.py` has been updated so that the parameter values that are being fitted are accessed using the string keys rather than integer indices in the `jac_array` function.
- The `coerce_float64` function in `lmfit/minimizer.py` now has explicit support for sparse matrices and `LinearOperator` objects (same approach as in [PR 589](https://github.com/lmfit/lmfit-py/pull/589)), which was required due to exceptions that were being raised when running `test_least_squares_jacobian_types` after the changes that were made in order to wrap Jacobian functions.

Additionally, the following changes/additions are included:

- Added `test_least_squares_jacobian` to `tests/test_least_squares.py` to verify the expected behavior.
- Added `tests/test_jacobian_pickling.py` to verify that `Minimizer` objects can be pickled when a Jacobian function is provided.
- The docstring of `Minimizer._jacobian` has also been updated to include information about the parameters and return values (similar to `Minimizer.__residual`).

###### Type of Changes

- [x] Bug fix
- [ ] New feature

###### Tested on

Python: 3.12.6 (main, Sep  8 2024, 13:18:56) [GCC 14.2.1 20240805]

lmfit: 1.3.2.post13+g39354856, scipy: 1.14.1, numpy: 2.1.3, asteval: 1.0.5, uncertainties: 3.2.2

###### Verification 
Have you

  - **Yes, once the tests were updated.**
  - **Not yet since I'm not sure a) what the version number should be (1.3.3 or 2.0.0) and b) what other kinds of major changes one might wish to do at the same time if it is necessary to increment the major version number.**
- [ ] added an example?
  - **Not sure if it is needed.**

###### Minimal reproducible example

Here's a modified version of [an example from the documentation](https://lmfit.github.io/lmfit-py/examples/example_fit_with_derivfunc.html) that I used to troubleshoot and test various issues.

```python
from multiprocessing import get_context
import numpy as np
from time import time

from lmfit import minimize, Minimizer, Parameters


def func(pars, x, data=None):
    a, b, c = pars['a'], pars['b'], pars['c']
    model = a * np.exp(-b*x) + c
    if data is None:
        return model
    return model - data


def dfunc(pars, x, data=None):
    a, b = pars['a'], pars['b']
    v = np.exp(-b*x)
    return np.array([v, -a*x*v, np.ones(len(x))])


def jacfunc(pars, x, data=None):
    a, b = pars['a'], pars['b']
    v = np.exp(-b*x)
    jac = np.ones((len(x), 3), dtype=np.float64)
    jac[:, 0] = v
    jac[:, 1] = -a * x * v
    return jac


def f(var, x):
    return var[0] * np.exp(-var[1]*x) + var[2]


def wrapper(args):
    (
        params,
        method,
        x,
        data,
        use_least_squares,
    ) = args
    return minimize(
        func,
        params,
        method=method,
        args=(x,),
        kws={'data': data},
        **(
            ({'jac': jacfunc} if use_least_squares else {'Dfun': dfunc, 'col_deriv': 1})
        ),
    )


if __name__ == "__main__":
    use_least_squares = bool(1)
    use_minimize_function = bool(0)
    swap_order = bool(0)
    use_pool = bool(1)
    num_repeats = 100

    method = 'least_squares' if use_least_squares else 'leastsq'
    print(f"{method=}")

    params = Parameters()
    params.add('a', value=10)
    params.add('b', value=10)
    params.add('c', value=10)

    a, b, c = 2.5, 1.3, 0.8
    x = np.linspace(0, 4, 50)
    y = f([a, b, c], x)
    np.random.seed(2021)
    data = y + 0.15*np.random.normal(size=x.size)

    # Just to warm things up. If this is removed, then the first sample
    # is quite a bit slower compared to all subsequent samples (see the
    # num_repeats variable) regardless of whether or not the Jacobian is
    # estimated numerically or calculated analytically (see the swap_order
    # variable).
    minimize(
        func,
        params,
        method=method,
        args=(x,),
        kws={'data': data},
    )

    deltas = []
    for _ in range(num_repeats):
        start = time()

        if use_minimize_function:
            out1 = minimize(
                func,
                params,
                method=method,
                args=(x,),
                kws={'data': data},
                **(
                    ({'jac': jacfunc} if use_least_squares else {'Dfun': dfunc, 'col_deriv': 1})
                    if swap_order else
                    {}
                ),
            )
        else:
            min1 = Minimizer(func, params, fcn_args=(x,), fcn_kws={'data': data})
            if use_least_squares:
                if swap_order:
                    out1 = min1.least_squares(jac=jacfunc)
                else:
                    out1 = min1.least_squares()
            else:
                if swap_order:
                    out1 = min1.leastsq(Dfun=dfunc, col_deriv=1)
                else:
                    out1 = min1.leastsq()

        deltas.append(time() - start)

    if swap_order:
        print(f"   With: {np.mean(deltas):.6f} +/- {np.std(deltas, ddof=1):.6f} s (n={len(deltas)})")
    else:
        print(f"Without: {np.mean(deltas):.6f} +/- {np.std(deltas, ddof=1):.6f} s (n={len(deltas)})")
    fit1 = func(out1.params, x)

    deltas = []
    for _ in range(num_repeats):
        start =  time()

        if use_minimize_function:
            out2 = minimize(
                func,
                params,
                method=method,
                args=(x,),
                kws={'data': data},
                **(
                    {}
                    if swap_order else
                    ({'jac': jacfunc} if use_least_squares else {'Dfun': dfunc, 'col_deriv': 1})
                ),
            )
        else:
            min2 = Minimizer(func, params, fcn_args=(x,), fcn_kws={'data': data})
            if use_least_squares:
                if swap_order:
                    out2 = min2.least_squares()
                else:
                    out2 = min2.least_squares(jac=jacfunc)
            else:
                if swap_order:
                    out2 = min2.leastsq()
                else:
                    out2 = min2.leastsq(Dfun=dfunc, col_deriv=1)

        deltas.append(time() - start)

    if swap_order:
        print(f"Without: {np.mean(deltas):.6f} +/- {np.std(deltas, ddof=1):.6f} s (n={len(deltas)})")
    else:
        print(f"   With: {np.mean(deltas):.6f} +/- {np.std(deltas, ddof=1):.6f} s (n={len(deltas)})")
    fit2 = func(out2.params, x)

    print(f'"true" parameters are: a = {a:.3f}, b = {b:.3f}, c = {c:.3f}\n\n'
          '|=========================================|\n'
          + (
              '| Statistic/Parameter | With    | Without |\n' if swap_order else '| Statistic/Parameter | Without | With    |\n'
          )
          + '|-----------------------------------------|\n'
          f'|  N Function Calls   | {out1.nfev:d}      | {out2.nfev:d}      |\n'
          f'|     Chi-square      | {out1.chisqr:.4f}  | {out2.chisqr:.4f}  |\n'
          f"|         a           | {out1.params['a'].value:.4f}  | {out2.params['a'].value:.4f}  |\n"
          f"|         b           | {out1.params['b'].value:.4f}  | {out2.params['b'].value:.4f}  |\n"
          f"|         c           | {out1.params['c'].value:.4f}  | {out2.params['c'].value:.4f}  |\n"
          '|-----------------------------------------|')

    plt.plot(x, data, 'o', label='data')
    plt.plot(x, fit1, label=('with' if swap_order else 'without') + ' analytical derivative')
    plt.plot(x, fit2, '--', label=('without' if swap_order else 'with') + ' analytical derivative')

    if use_pool:
        # To confirm that everything is working as intended when used together
        # with multiprocessing.Pool.
        with get_context(method="spawn").Pool(1) as pool:
            for out3 in pool.map(
                wrapper,
                ((
                    params,
                    method,
                    x,
                    data,
                    use_least_squares,
                ),),
            ):
                print(f"\n{out3.nfev=}, {out3.chisqr=:.4f}, {out3.params=}")
                fit3 = func(out3.params, x)
                plt.plot(x, fit3, ':', label='with analytical derivative (Pool)')

    plt.legend()
    plt.show()
```

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -55,7 +55,7 @@ Roam, Alexander Stark, Alexandre Beelen, Andrey Aristov, Nicholas Zobrist,
 Ethan Welty, Julius Zimmermann, Mark Dean, Arun Persaud, Ray Osborn, @lneuhaus,
 Marcel Stimberg, Yoshiera Huang, Leon Foks, Sebastian Weigand, Florian LB,
 Michael Hudson-Doyle, Ruben Verweij, @jedzill4, @spalato, Jens Hedegaard Nielsen,
-Martin Majli, Kristian Meyer, @azelcer, Ivan Usov, and many others.
+Martin Majli, Kristian Meyer, @azelcer, Ivan Usov, Ville Yrjänä, and many others.
 
 The lmfit code obviously depends on, and owes a very large debt to the code
 in scipy.optimize. Several discussions on the SciPy-user and lmfit mailing
--- a/examples/example_fit_jacobian_benchmark.py
+++ b/examples/example_fit_jacobian_benchmark.py
@@ -1,0 +1,178 @@
+"""
+Benchmarks of methods with and without computing the Jacobian analytically
+==========================================================================
+
+Providing a function that calculates the Jacobian matrix analytically can
+reduce the time spent finding a solution. The results from benchmarks comparing
+two methods (``leastsq`` and ``least_squares``) with and without a function to
+calculate the Jacobian matrix analytically are presented below.
+
+First we define the model function, the residual function, and the appropriate
+Jacobian functions:
+"""
+from timeit import timeit
+from types import SimpleNamespace
+
+import matplotlib.pyplot as plt
+import numpy as np
+
+from lmfit import Parameters, minimize
+
+NUM_JACOBIAN_CALLS = 0
+
+
+def func(var, x):
+    return var[0] * np.exp(-var[1]*x) + var[2]
+
+
+def residual(pars, x, data):
+    a, b, c = pars['a'], pars['b'], pars['c']
+    model = func((a, b, c), x)
+    return model - data
+
+
+def dfunc(pars, x, data):
+    global NUM_JACOBIAN_CALLS
+    NUM_JACOBIAN_CALLS += 1
+
+    a, b = pars['a'], pars['b']
+    v = np.exp(-b*x)
+    return np.array([v, -a*x*v, np.ones(len(x))])
+
+
+def jacfunc(pars, x, data):
+    global NUM_JACOBIAN_CALLS
+    NUM_JACOBIAN_CALLS += 1
+
+    a, b = pars['a'], pars['b']
+    v = np.exp(-b*x)
+    jac = np.ones((len(x), 3), dtype=np.float64)
+    jac[:, 0] = v
+    jac[:, 1] = -a * x * v
+    return jac
+
+
+a, b, c = 2.5, 1.3, 0.8
+
+x = np.linspace(0, 4, 50)
+y = func([a, b, c], x)
+
+data = y + 0.15*np.random.RandomState(seed=2021).normal(size=x.size)
+
+
+###############################################################################
+# Then we define the different cases to benchmark (i.e., different methods with
+# and without a function to calculate the Jacobian analytically) and the number
+# of repetitions per case:
+cases = (
+    dict(
+        method='leastsq',
+    ),
+    dict(
+        method='leastsq',
+        Dfun=dfunc,
+        col_deriv=1,
+    ),
+    dict(
+        method='least_squares',
+    ),
+    dict(
+        method='least_squares',
+        jac=jacfunc,
+    ),
+)
+
+num_repeats = 100
+results = []
+
+for kwargs in cases:
+    params = Parameters()
+    params.add('a', value=10)
+    params.add('b', value=10)
+    params.add('c', value=10)
+
+    wrapper = lambda: minimize(
+        residual,
+        params,
+        args=(x,),
+        kws={'data': data},
+        **kwargs,
+    )
+    time = timeit(wrapper, number=num_repeats) / num_repeats
+
+    NUM_JACOBIAN_CALLS = 0
+    fit = wrapper()
+
+    results.append(SimpleNamespace(
+        time=time,
+        num_jacobian_calls=NUM_JACOBIAN_CALLS,
+        fit=fit,
+        kwargs=kwargs,
+    ))
+
+
+###############################################################################
+# Finally, we present the results:
+labels = []
+
+for result in results:
+    label = result.kwargs['method']
+    if result.num_jacobian_calls > 0:
+        label += ' with Jac.'
+
+    labels.append(label)
+
+label_width = max(map(len, labels))
+lines = [
+    '| '
+    + ' | '.join([
+        'Method'.ljust(label_width),
+        'Avg. time (ms)',
+        '# func. (+ Jac.) calls',
+        'Chi-squared',
+        'a'.ljust(5),
+        'b'.ljust(5),
+        'c'.ljust(6),
+    ])
+    + '|'
+]
+
+print(f'The "true" parameters are: a = {a:.3f}, b = {b:.3f}, c = {c:.3f}\n')
+fig, ax = plt.subplots()
+ax.plot(x, data, marker='.', linestyle='none', label='data')
+
+for (result, label) in zip(results, labels):
+    linestyle = '-'
+    if result.num_jacobian_calls > 0:
+        linestyle = '--'
+
+    a = result.fit.params['a'].value
+    b = result.fit.params['b'].value
+    c = result.fit.params['c'].value
+    y = func([a, b, c], x)
+    ax.plot(x, y, label=label, alpha=0.5, linestyle=linestyle)
+
+    columns = [
+        label.ljust(label_width),
+        f'{result.time * 1000:.2f}'.ljust(14),
+        (
+            f'{result.fit.nfev}'
+            + (
+                f' (+{result.num_jacobian_calls})'
+                if result.num_jacobian_calls > 0 else
+                ''
+            )
+        ).ljust(22),
+        f'{result.fit.chisqr:.3f}'.ljust(11),
+        f'{a:.3f}'.ljust(5, '0'),
+        f'{b:.3f}'.ljust(5, '0'),
+        f'{c:.3f}'.ljust(5, '0'),
+    ]
+    lines.append('| ' + ' | '.join(columns) + ' |')
+
+lines.insert(1, '|-' + '-|-'.join('-' * len(col) for col in columns) + '-|')
+print('\n'.join(lines))
+
+ax.set_xlabel('x')
+ax.set_ylabel('y')
+ax.legend()
--- a/lmfit/minimizer.py
+++ b/lmfit/minimizer.py
@@ -2396,7 +2396,12 @@ def coerce_float64(arr, nan_policy='raise', handle_inf=True,
     lists of numbers, pandas.Series, h5py.Datasets, and many other array-like
     Python objects
     """
-    if np.iscomplexobj(arr):
+    if issparse(arr):
+        arr = arr.toarray().astype(np.float64)
+    elif isinstance(arr, LinearOperator):
+        identity = np.eye(arr.shape[1], dtype=np.float64)
+        arr = (arr * identity).astype(np.float64)
+    elif np.iscomplexobj(arr):
         arr = np.asarray(arr, dtype=np.complex128).view(np.float64)
     else:
         arr = np.asarray(arr, dtype=np.float64)
--- a/lmfit/minimizer.py
+++ b/lmfit/minimizer.py
@@ -565,8 +565,11 @@ class Minimizer:
         grad_scale = np.ones_like(fvars)
         for ivar, name in enumerate(self.result.var_names):
             val = fvars[ivar]
-            pars[name].value = pars[name].from_internal(val)
-            grad_scale[ivar] = pars[name].scale_gradient(val)
+            if apply_bounds_transformation:
+                pars[name].value = pars[name].from_internal(val)
+                grad_scale[ivar] = pars[name].scale_gradient(val)
+            else:
+                pars[name].value = val
 
         pars.update_constraints()
 
--- a/lmfit/minimizer.py
+++ b/lmfit/minimizer.py
@@ -553,7 +553,7 @@ class Minimizer:
         else:
             return coerce_float64(out, nan_policy=self.nan_policy)
 
-    def __jacobian(self, fvars, apply_bounds_transformation=True):
+    def _jacobian(self, fvars, apply_bounds_transformation=True):
         """Return analytical jacobian to be used with Levenberg-Marquardt.
 
         modified 02-01-2012 by Glenn Jones, Aberystwyth University
@@ -934,7 +934,7 @@ class Minimizer:
         # Wrap Jacobian function to deal with bounds
         if 'jac' in fmin_kws:
             self.jacfcn = fmin_kws.pop('jac')
-            fmin_kws['jac'] = self.__jacobian
+            fmin_kws['jac'] = self._jacobian
 
         # Ignore jac argument for methods that do not support it
         if 'jac' in fmin_kws and method not in ('CG', 'BFGS', 'Newton-CG',
@@ -1540,7 +1540,7 @@ class Minimizer:
 
         if callable(least_squares_kws['jac']):
             self.jacfcn = least_squares_kws['jac']
-            least_squares_kws['jac'] = self.__jacobian
+            least_squares_kws['jac'] = self._jacobian
 
         least_squares_kws['kwargs'].update({'apply_bounds_transformation': False})
         result.call_kws = least_squares_kws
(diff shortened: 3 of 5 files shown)


vyrjana:
It looks like some lines of [the calculation of the Hessian matrix](https://app.codecov.io/gh/lmfit/lmfit-py/pull/973/indirect-changes) are no longer being reached due to [the changes to `coerce_float64`](https://github.com/vyrjana/lmfit-py/blob/0a3ca2f6504c42acce57ee5a8190fdfcad09cd75/lmfit/minimizer.py#L2413-L2421). Based on reading the code, it seems to me like the calculated values should still be correct. The Jacobian matrix is [coerced into a 1-D `NDArray[float64]` here](https://github.com/vyrjana/lmfit-py/blob/fix-least-squares-jacobian/lmfit/minimizer.py#L593) before it is passed on to the minimizer, so I would expect that the Jacobian included in [the return value from the minimizer (i.e., `ret.jac`)](https://github.com/vyrjana/lmfit-py/blob/fix-least-squares-jacobian/lmfit/minimizer.py#L1563-L1565) should be fine to use as is regardless of whether the user-provided Jacobian function, which has been wrapped, returns a 1-D `NDArray[float64]`, a sparse matrix, or a `LinearOperator`.

There are [quite a few lines that are currently not covered](https://app.codecov.io/gh/lmfit/lmfit-py/pull/973/blob/lmfit/minimizer.py), but these are probably mostly lines that were not reached previously either. Right?

vyrjana:
Regarding the need to increment the major version number; if one wanted to delay the need to increment, then I suppose one could probe the Jacobian function to see if it fails with a `Parameters` object but succeeds with an array of float values before the fitting actually starts (e.g., [after this line](https://github.com/vyrjana/lmfit-py/blob/fix-least-squares-jacobian/lmfit/minimizer.py#L1556)). One would then set a flag within the `Minimizer` class that determines how the parameters are [passed to the Jacobian function at this line](https://github.com/vyrjana/lmfit-py/blob/fix-least-squares-jacobian/lmfit/minimizer.py#L592). This would be a temporary solution that would maintain backwards compatibility until there is a more pressing need to jump to version 2.0.0.

newville:
@vyrjana Thanks - this has indeed been a long-standing (if perhaps not very common) issue.  I think this will need some review, which might take a little bit of time.  

For now, I believe this does not require a change to the major version number - it is not a change in the main API of lmfit, it would be more like cleanup and better uniformity of the intended usage.  I would suggest adding your testing example as an actual example.

vyrjana:
@newville Alright. Do you mean that my example should be included in its entirety? Or should I just amend `lmfit-py/examples/example_fit_with_derivfunc.py` to also include how to provide a Jacobian function when using `Minimizer.least_squares`?

newville:
@vyrjana (very sorry for the delay in responding).  I think it would be OK to include your full example, but probably better to make a separate example like (perhaps copy-and-modify)  "example_fit_with_derivfunc.py" especially when using `least_squares`.    

I think benchmarks are okay.  ***But***: as I have ranted about in the past, any benchmark using `least_squares` should use our measurement of how many times the objective function is called (and now also how many times the Jacobian function is called), and absolutely not the values reported by `least_squares`.  It lies.  It lies deliberately.  The authors of that function are deliberate liars.

I have no doubt that providing a Jacobian will significantly improve performance over not including a Jacobian.

vyrjana:
No problem, @newville.

Below is a cleaned up version (e.g., I removed the multiprocessing part and the various flags such as `swap_order`). I ended up switching over to using the [`timeit`](https://docs.python.org/3/library/timeit.html) module due to the fairly regular spikes I was seeing in the elapsed times. Those spikes were most likely due to the garbage collector (GC), but that should be sorted out now since the GC is disabled by default when using `timeit`. If this looks alright, then I'll add it to the documentation as a separate example.

```python
"""
Benchmarks of methods with and without computing the Jacobian analytically
==========================================================================

Providing a function that calculates the Jacobian matrix analytically can
reduce the time spent finding a solution. The results from benchmarks comparing
two methods (``leastsq`` and ``least_squares``) with and without a function to
calculate the Jacobian matrix analytically are presented below.

First we define the model function, the residual function, and the appropriate
Jacobian functions:
"""
from timeit import timeit
from types import SimpleNamespace
import numpy as np
import scipy as sp

from lmfit import minimize, Parameters
from lmfit.minimizer import MinimizerResult


NUM_JACOBIAN_CALLS = 0


def func(var, x):
    return var[0] * np.exp(-var[1]*x) + var[2]


def residual(pars, x, data):
    a, b, c = pars['a'], pars['b'], pars['c']
    model = func((a, b, c), x)
    return model - data


def dfunc(pars, x, data):
    global NUM_JACOBIAN_CALLS
    NUM_JACOBIAN_CALLS += 1
    
    a, b = pars['a'], pars['b']
    v = np.exp(-b*x)
    return np.array([v, -a*x*v, np.ones(len(x))])


def jacfunc(pars, x, data):
    global NUM_JACOBIAN_CALLS
    NUM_JACOBIAN_CALLS += 1

    a, b = pars['a'], pars['b']
    v = np.exp(-b*x)
    jac = np.ones((len(x), 3), dtype=np.float64)
    jac[:, 0] = v
    jac[:, 1] = -a * x * v
    return jac


a, b, c = 2.5, 1.3, 0.8

x = np.linspace(0, 4, 50)
y = func([a, b, c], x)

np.random.seed(2021)
data = y + 0.15*np.random.normal(size=x.size)


###############################################################################
# Then we define the different cases to benchmark (i.e., different methods with
# and without a function to calculate the Jacobian analytically) and the number
# of repetitions per case:
cases = (
    dict(
        method='leastsq',
    ),
    dict(
        method='leastsq',
        Dfun=dfunc,
        col_deriv=1,
    ),
    dict(
        method='least_squares',
    ),
    dict(
        method='least_squares',
        jac=jacfunc,
    ),
)

num_repeats = 100
results = []

for kwargs in cases:
    params = Parameters()
    params.add('a', value=10)
    params.add('b', value=10)
    params.add('c', value=10)

    wrapper = lambda: minimize(
        residual,
        params,
        args=(x,),
        kws={'data': data},
        **kwargs,
    )
    time = timeit(wrapper, number=num_repeats) / num_repeats

    NUM_JACOBIAN_CALLS = 0
    fit = wrapper()

    results.append(SimpleNamespace(
        time=time,
        num_jacobian_calls=NUM_JACOBIAN_CALLS,
        fit=fit,
        kwargs=kwargs,
    ))


###############################################################################
# Finally, we present the results:
labels = []

for result in results:
    label = result.kwargs['method']
    if result.num_jacobian_calls > 0:
        label += ' with Jac.'

    labels.append(label)

label_width = max(map(len, labels))
lines = [
    '| '
    + ' | '.join([
        'Method'.ljust(label_width),
        '# func. (+ Jac.) calls',
        'Avg. time (ms)',
        'Chi-squared',
        'a'.ljust(5),
        'b'.ljust(5),
        'c'.ljust(6),
    ])
    + '|'
]

print(f'The "true" parameters are: a = {a:.3f}, b = {b:.3f}, c = {c:.3f}\n')
fig, ax = plt.subplots()
ax.plot(x, data, marker='.', linestyle='none', label='data')

for (result, label) in zip(results, labels):
    linestyle = '-'
    if result.num_jacobian_calls > 0:
        linestyle = '--'

    a = result.fit.params['a'].value
    b = result.fit.params['b'].value
    c = result.fit.params['c'].value
    y = func([a, b, c], x)
    ax.plot(x, y, label=label, alpha=0.5, linestyle=linestyle)

    columns = [
        label.ljust(label_width),
        (
            f'{result.fit.nfev}'
            + (
                f' (+{result.num_jacobian_calls})'
                if result.num_jacobian_calls > 0 else
                ''
            )
        ).ljust(22),
        f'{result.time * 1000:.2f}'.ljust(14),
        f'{result.fit.chisqr:.3f}'.ljust(11),
        f'{a:.3f}'.ljust(5, '0'),
        f'{b:.3f}'.ljust(5, '0'),
        f'{c:.3f}'.ljust(5, '0'),
    ]
    lines.append('| ' + ' | '.join(columns) + ' |')
    if len(lines) < 3:
        lines.insert(1, 
            '|-'
            + '-|-'.join(('-' * len(col) for col in columns))
            + '-|'
        )

print('\n'.join(lines))

ax.set_xlabel('x')
ax.set_ylabel('y')
ax.legend()
```

Here is an example of the results from my machine:

| Method                  | # func. (+ Jac.) calls | Avg. time (ms) | Chi-squared | a     | b     | c     |
|-------------------------|------------------------|----------------|-------------|-------|-------|-------|
| leastsq                 | 39                     | 1.28           | 1.092       | 2.564 | 1.359 | 0.824 |
| leastsq with Jac.       | 12 (+10)               | 1.01           | 1.092       | 2.564 | 1.359 | 0.824 |
| least_squares           | 38                     | 2.88           | 1.092       | 2.564 | 1.359 | 0.824 |
| least_squares with Jac. | 11 (+9)                | 1.78           | 1.092       | 2.564 | 1.359 | 0.824 |

So, in this case the relative changes in elapsed times were about -21 % for `leastsq` and -38 % for `least_squares` when providing a Jacobian function.

newville:
@vyrjana Thanks.  I'm not too surprised that the number of function calls with `leastsq` and `least_squares` is "about the same".   Providing the Jacobian function reduced the total number of function (residual+Jacobian) calls by almost 2x. I would call that "pretty good!".  Curiously, the run time did not go down by that much.  Maybe this 3-parameter problem is "too easy" - my experience is that exponential decay is challenging for `leastsq`, so I would not have guessed that.

The observation that `least_squares` is ~2x slower than `leastsq` when not using a Jacobian is similar to my experiences, though I have not explored that systematically.   It seems likely that "pure Python least_squares" has some performance hit over "wrapper around Fortran/C", though I cannot guess how easy it would be to fix that.

In general, I continue to have a poor opinion of `least_squares`.  I sort of understand the desire to update `leastsq`, except that  a) MINPACK is simply the benchmark for speed and accuracy (and gives the matrices for pretty accurate 1-sigma error bars too), and b) the actual implementation and API of `least_squares` is just a mess.   
It is a shame that `scipy` lists `leastsq` as "Legacy" and "not recommended".

Anyway, thanks, and adding support for a Jacobian with `least_squares` is certainly a good idea!

Is this ready to merge?  I think so...

vyrjana:
The PR should now be ready to merge. Here is a suggestion for the release notes:

```
.. _whatsnew_133_label:

Version 1.3.3 Release Notes (December 4, 2024)
====================================================

Bug fixes/enhancements:

- ``Minimizer.least_squares`` now properly wraps the optional Jacobian function so that the first argument of the Jacobian function is a ``Parameters`` object (PR #973)
- renamed the ``Minimizer.__jacobian`` method to ``Minimizer._jacobian`` to avoid raising an exception when (un)pickling the method while making use of the ``multiprocessing`` module (PR #973)
- updated the ``coerce_float64`` function to explicitly support sparce matrices and ``scipy.sparse.linalg.LinearOperator`` objects (PR #973)
- added/updated tests to ensure that the Jacobian function is provided a ``Parameters`` object (PR #973)
- added tests to ensure successful (un)pickling of a ``Minimizer`` object that has been provided a Jacobian function (PR #973)


Documentation:

- added an example to demonstrate the performance benefits of providing a function that calculates the Jacobian matrix analytically (PR #973)
- updated the docstring of ``Minimizer._jacobian`` (PR #973)
```

vyrjana:
@newville You're welcome and thank you to you as well. The lmfit package has been very useful in my own projects, so I'm glad I could contribute.

PR #997, opened 2025-03-25
Optimize _func_allargs by using set instead of list

allrob23:
#### Description

Description:
The variable `_func_allargs` is used in 7 cases within an if `key in _func_allargs:` check. When using a `list`, the membership test (`in`) has a time complexity of O(N). By switching to a `set`, this lookup becomes O(1), significantly improving performance, especially when `_func_allargs` grows in size.

##### Changes
- Replaced list initialization with a set.
- Updated .append(key) calls to .add(key).
- Converted list concatenation to set union to maintain efficiency.

This should improve performance while keeping functionality intact. Let me know if any adjustments are needed!

###### Type of Changes

- [ ] Bug fix
- [ ] New feature

###### Verification 
Have you

- [ ] added an example?

### DIFF
--- a/lmfit/model.py
+++ b/lmfit/model.py
@@ -293,7 +293,7 @@ class Model:
 
         self._param_root_names = param_names  # will not include prefixes
         self.independent_vars = independent_vars
-        self._func_allargs = []
+        self._func_allargs = set()
         self._func_haskeywords = False
         self.nan_policy = nan_policy
 
@@ -548,10 +548,10 @@ class Model:
         # inspection done
 
         self._func_haskeywords = keywords_ is not None
-        self._func_allargs = list(default_vals.keys())
+        self._func_allargs = set(default_vals.keys())
         for key in kw_args:
             if key not in self._func_allargs:
-                self._func_allargs.append(key)
+                self._func_allargs.add(key)
 
         if len(self._func_allargs) == 0 and keywords_ is not None:
             return
@@ -1275,7 +1275,7 @@ class CompositeModel(Model):
     def _parse_params(self):
         self._func_haskeywords = (self.left._func_haskeywords or
                                   self.right._func_haskeywords)
-        self._func_allargs = (self.left._func_allargs +
+        self._func_allargs = set(self.left._func_allargs +
                               self.right._func_allargs)
         self.def_vals = deepcopy(self.right.def_vals)
         self.def_vals.update(self.left.def_vals)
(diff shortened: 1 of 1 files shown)


newville:
@allrob23 Please run the tests locally -- it looks like they are failing because you have `set1 + set2`, which is not allowed.  

But also: what problem does this solve?  `Model.func_allargs` is made at model creation time, when the function signature of the model function is parsed.  It is then used (only) at model evaluation time.   I would guess (but I would be happy to be shown actual evidence either way) that it does not dominate model creation time.  

If model creation time is dominating or even a significant fraction of model use time, then `lmfit.Model` may not be the right approach to whatever problem you are trying to solve ;)

allrob23:
Sorry for the mistake, I updated the commit now.

I don’t have much knowledge of the codebase, but I created a tool that looks for situations where a list is used for membership tests when a set could be more efficient. This tool flagged this location, and while running tests for another project that uses this library (github.com/GeoscienceAustralia/sira), I noticed that this part of the code was executed many times. Given that, I thought this update could make sense and wanted to contribute by opening the PR.

lmfit/model.py
```diff
@@ -1275,7 +1275,7 @@ def _tmp(self, *args, **kws):
     def _parse_params(self):
         self._func_haskeywords = (self.left._func_haskeywords or
                                   self.right._func_haskeywords)
-        self._func_allargs = (self.left._func_allargs +
+        self._func_allargs = set(self.left._func_allargs +
```
VascoSch92, line 1278: this should be

```python
set(self.left._func_allargs).union(self.right._func_allargs)
```

allrob23: done

newville:
@allrob23 If I understand correctly:

a) you stated that you ran the tests locally when you, in fact, did not.
b) when told to fix a syntax error and to run the tests locally, you changed the code again, and again did not run tests locally or bother looking into how sets work in Python. 
c) when asked to explain what problem this solves, you said that you developed a tool for analyzing Python code.  Given the demonstration in a) and b), you have not tested your code well.
d) You have still not run the tests locally.  They still show a clear error with the code you changed.
e) It appears you have not ever used this library.   
f) You claim your fix is a performance improvement.  You have offered no evidence of that, and add no tests.

Wishing you well in your future endeavors.

This is a mature, stable project, and we're going to have to pass on your contributions.

Issue #999, opened 2025-04-21
Getting UFloat test failure with uncertainties 3.2.3


doronbehar:
** Description **

When attempting to build `lmfit` with `uncertainties` 3.2.3, I recieved the following test failure:

```
=================================== FAILURES ===================================
___________________________ test_confidence_warnings ___________________________

data = (array([ 0.3       ,  0.3979798 ,  0.4959596 ,  0.59393939,  0.69191919,
        0.78989899,  0.88787879,  0.98585859,...29492,  3.09446376,  3.16012366,  3.08715018,
        3.1114479 ,  3.03133411,  3.19857469,  3.02258614,  3.04019894]))
pars = Parameters([('a', <Parameter 'a', value=0.1, bounds=[-inf:inf]>), ('b', <Parameter 'b', value=1, bounds=[-inf:inf]>)])

    def test_confidence_warnings(data, pars):
        """Make sure the proper warnings are emitted when needed."""
        minimizer = lmfit.Minimizer(residual, pars, fcn_args=data)
        out = minimizer.minimize(method='leastsq')
    
        with pytest.warns(UserWarning) as record:
            lmfit.conf_interval(minimizer, out, maxiter=1)
>           assert 'maxiter=1 reached and prob' in str(record[0].message)
E           AssertionError: assert 'maxiter=1 reached and prob' in 'Using UFloat objects with std_dev==0 may give unexpected results.'
E            +  where 'Using UFloat objects with std_dev==0 may give unexpected results.' = str(UserWarning('Using UFloat objects with std_dev==0 may give unexpected results.'))
E            +    where UserWarning('Using UFloat objects with std_dev==0 may give unexpected results.') = <warnings.WarningMessage object at 0x7fff693090a0>.message

tests/test_confidence.py:201: AssertionError
=============================== warnings summary ===============================
```
 
** A Minimal, Complete, and Verifiable example **

Just run the tests with `uncertainties` 3.2.3?

** Fit report: **

Irrelevant

** Error message: **

Pasted above.

** Version information: **

Manually taken from distribution:

- `sys.version`: `'3.12.9 (main, Feb  4 2025, 14:38:38) [GCC 14.2.1 20241116]'`
- `lmfit`: 1.3.3
- `numpy`: 2.2.4
- `scipy`: 1.15.2
- `asteval`: 1.0.6
- `uncertainties`: 3.2.3, as stated in title. Issue does not occur with 3.2.2.

** Link(s): **

[...]

PR #1000, opened 2025-04-22
Fix changed stddev0 warning

newville:
#### Description

This addresses #999 for the recent changes in how `uncertainties` warns about `ufloat()` values with `std_dev=0`. 

The changes here are:
1.  do not test for the content of warning messages from upstream libraries.  They can change.
2. capture and ignore warnings about `stderr=0` when creating `uncertainties.ufloat()` in `conf_interval()`.

###### Type of Changes

- [x] Bug fix
- [ ] New feature

###### Tested on

Python: 3.12.8 | packaged by conda-forge | (main, Dec  5 2024, 14:25:12) [Clang 18.1.8 ]

lmfit: 1.3.3.post3+g132fef4e, scipy: 1.15.1, numpy: 1.26.4, asteval: 1.0.6, uncertainties: 3.2.3

###### Verification 
Have you

- [ ] added an example?

### DIFF
--- a/lmfit/parameter.py
+++ b/lmfit/parameter.py
@@ -574,7 +574,7 @@ class Parameters(dict):
             stderr = getattr(par, 'stderr', 0.0)
             if stderr is None:
                 stderr = 0.00
-            if stderr < 1.e-13*abs(par.value):
+            if stderr < tiny*max(tiny, abs(par.value)):
                 with warnings.catch_warnings():
                     warnings.simplefilter("ignore")
                     uvars[par.name] = ufloat(par.value, stderr)
@@ -596,7 +596,6 @@ class Parameters(dict):
             # parameters and reset the Parameters to best-fit value
             wrap_ueval = uwrap(asteval_with_uncertainties)
             for par in self.values():
-                print("xPAR ", par)
                 if getattr(par, '_expr_ast', None) is not None:
                     try:
                         uval = wrap_ueval(*corr_uvars, obj=par,
--- a/lmfit/parameter.py
+++ b/lmfit/parameter.py
@@ -2,6 +2,7 @@
 
 from copy import deepcopy
 import json
+import warnings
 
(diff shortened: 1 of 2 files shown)


lmfit/parameter.py
```diff
@@ -572,8 +573,13 @@ def create_uvars(self, covar=None):
                     par.stderr = sqrt(covar[vindex, vindex])
             stderr = getattr(par, 'stderr', 0.0)
             if stderr is None:
-                stderr = 0.0
-            uvars[par.name] = ufloat(par.value, stderr)
+                stderr = 0.00
+            if stderr < 1.e-13*abs(par.value):
```
wshanks, line 577: `par.value == 0` might be an edge case to handle? I wonder if you could just always filter the warnings here without testing stderr or only test for `stderr==0` instead of with the inequality?

wshanks (commented):
I was checking in here since this new warning started causing some tests I work with to fail because they use `vary=False` and are set to fail if any warnings are generated, so I thought I would leave a review.

lmfit/parameter.py
```diff
@@ -590,6 +596,7 @@ def create_uvars(self, covar=None):
             # parameters and reset the Parameters to best-fit value
             wrap_ueval = uwrap(asteval_with_uncertainties)
             for par in self.values():
+                print("xPAR ", par)
```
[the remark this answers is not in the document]
newville: @wshanks Thanks!

lmfit/parameter.py
newville: Thanks, that's a good suggestion.  I also see that `tiny` is defined here (as 1.e-15).  
I will change to `if stderr < tiny*max(tiny, abs(par.value)):`, just to be super-paranoid about floating point.

wshanks (approved):

newville:
Merging...

PR #1011, opened 2025-07-20
fix import in test_composite_plotting

mgorny:
#### Description
Fix `matplotlib` import to be `matplotlib.pyplot`, as otherwise running the test fails with:

```
$ pytest tests/test_model.py::TestUserDefiniedModel::test_composite_plotting
========================================================= test session starts =========================================================
platform linux -- Python 3.13.5, pytest-8.4.1, pluggy-1.6.0
rootdir: /tmp/lmfit-py
configfile: pyproject.toml
plugins: forked-1.6.0, cov-6.2.1, flaky-3.8.1
collected 1 item

tests/test_model.py F                                                                                                           [100%]

============================================================== FAILURES ===============================================================
____________________________________________ TestUserDefiniedModel.test_composite_plotting ____________________________________________

self = <tests.test_model.TestUserDefiniedModel testMethod=test_composite_plotting>

    @pytest.mark.skipif(not lmfit.model._HAS_MATPLOTLIB,
                        reason="requires matplotlib.pyplot")
    def test_composite_plotting(self):
        # test that a composite model has non-empty best_values
        import matplotlib
        try:
>           matplotlib.pyplot.close('all')
            ^^^^^^^^^^^^^^^^^

tests/test_model.py:1132:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

name = 'pyplot'

    @functools.cache
    def __getattr__(name):
        if name in props:
            return props[name].__get__(instance)
>       raise AttributeError(
            f"module {cls.__module__!r} has no attribute {name!r}")
E       AttributeError: module 'matplotlib' has no attribute 'pyplot'

.venv/lib/python3.13/site-packages/matplotlib/_api/__init__.py:218: AttributeError
=========================================================== tests coverage ============================================================
___________________________________________ coverage: platform linux, python 3.13.5-final-0 ___________________________________________

Coverage HTML written to dir htmlcov
======================================================= short test summary info =======================================================
FAILED tests/test_model.py::TestUserDefiniedModel::test_composite_plotting - AttributeError: module 'matplotlib' has no attribute 'pyplot'
========================================================== 1 failed in 1.10s ==========================================================
```

This normally does not happen because `matplotlib.pyplot` gets imported elsewhere before, but it makes parallel testing with pytest-xdist flaky.

###### Type of Changes

- [x] Bug fix
- [ ] New feature

###### Tested on
```
Python: 3.13.5 (main, Jun 12 2025, 03:41:01) [GCC 14.3.0]

lmfit: 1.3.4, scipy: 1.16.0, numpy: 2.3.1, asteval: 1.0.6, uncertainties: 3.2.3
````

###### Verification 
Have you

### DIFF
--- a/tests/test_model.py
+++ b/tests/test_model.py
@@ -1127,7 +1127,7 @@ class TestUserDefiniedModel(CommonTests, unittest.TestCase):
                         reason="requires matplotlib.pyplot")
     def test_composite_plotting(self):
         # test that a composite model has non-empty best_values
-        import matplotlib
+        import matplotlib.pyplot
         try:
             matplotlib.pyplot.close('all')
         except ValueError:


newville:
@mgorny. All of our tests pass.   I see no such failures.

That isn't to say that we should not make some change here, but can you help explain why you are seeing this and none of our tests show the problem?   

Dunno anything about `pytest-xdist`.   Is the point (perhaps slightly overstated, but, well, maybe only slightly) of `pytest-xdist` to deliberately break assumptions built in to existing tests?   

If a test failure is happening only with a different set of testing tools (or platforms) than the ones we use, you first have to convince us that we *should* use and support those.

newville:
@mgorny I hear what you are saying, and I repeat that our tests do, in fact, pass.   We simply cannot guarantee that our tests will pass for others in other environments, platforms, situations.   We, literally, do not test those.

The purpose of our tests is for us to know that our code works as we expect.  When our tests fail for others, it is nice to know, but not really something we can be obligated to support.  

All that said, this seems like a simple fix for a non-standard usage of our tests that we can "just merge".

PR #1039, opened 2026-07-15
allow loading modelresult from text

### LINKED ISSUE #1013: load_modelresult() should accept a string or file-like object
as suggested at https://groups.google.com/g/lmfit-py/c/xu4WsUWufjw/m/y0-xLnpGDQAJ

newville:
will (finally) close #1013

#### Description

allow `load_model` and `load_modelresult` to accept a string of text as well as a filenname

###### Type of Changes

- [x] Bug fix
- [ ] New feature

###### Verification 
Have you

- [x] added an example?

### DIFF
--- a/examples/doc_model_loadmodelresult3.py
+++ b/examples/doc_model_loadmodelresult3.py
@@ -1,4 +1,4 @@
-# <examples/doc_model_loadmodelresult2.py>
+# <examples/doc_model_loadmodelresult3.py>
 import os
 import sys
 
--- a/examples/doc_model_loadmodelresult3.py
+++ b/examples/doc_model_loadmodelresult3.py
@@ -1,0 +1,23 @@
+# <examples/doc_model_loadmodelresult2.py>
+import os
+import sys
+
+import matplotlib.pyplot as plt
+import numpy as np
+
+from lmfit.model import load_modelresult
+
+if not os.path.exists('nistgauss_modelresult.sav'):
+    os.system(f"{sys.executable} doc_model_savemodelresult2.py")
+
+dat = np.loadtxt('NIST_Gauss2.dat')
+x = dat[:, 1]
+y = dat[:, 0]
+
+result = load_modelresult(open('nistgauss_modelresult.sav').read())
+print(result.fit_report())
+
+plt.plot(x, y, 'o')
+plt.plot(x, result.best_fit, '-')
+plt.show()
+# <end examples/doc_model_loadmodelresult2.py>
--- a/lmfit/model.py
+++ b/lmfit/model.py
@@ -5,6 +5,7 @@ from functools import wraps
 import inspect
 import json
 import operator
+from pathlib import Path
 import warnings
 
 from asteval import valid_symbol_name
@@ -21,6 +22,7 @@ from .minimizer import MinimizerResult
 from .printfuncs import ci_report, fit_report, fitreport_html_table
 
 tiny = 1.e-15
+MAXFILENAME_LENGTH = 2048
 
 # Use pandas.isnull for aligning missing data if pandas is available.
 # otherwise use numpy.isnan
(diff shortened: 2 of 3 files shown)


lmfit/model.py
```diff
@@ -1363,10 +1365,12 @@ def load_model(fname, funcdefs=None):
         Model object loaded from file.
 
     """
-    m = Model(lambda x: x)
-    with open(fname) as fh:
-        model = m.load(fh, funcdefs=funcdefs)
-    return model
+    mod = Model(lambda x: x)
+    mtext = fname
+    if len(fname) < MAXFILENAME_LENGTH and Path(fname).exists():
```
reneeotten, line 1372: isn't checking whether it's a file and if not assuming it's the text one wants to load an easier solution?

examples/doc_model_loadmodelresult3.py
```diff
@@ -0,0 +1,23 @@
+# <examples/doc_model_loadmodelresult2.py>
```
reneeotten:
```suggestion
# <examples/doc_model_loadmodelresult3.py>
```

examples/doc_model_loadmodelresult3.py
```diff
@@ -0,0 +1,23 @@
...
+x = dat[:, 1]
+y = dat[:, 0]
+
+result = load_modelresult(open('nistgauss_modelresult.sav').read())
+print(result.fit_report())
+
+plt.plot(x, y, 'o')
+plt.plot(x, result.best_fit, '-')
+plt.show()
+# <end examples/doc_model_loadmodelresult2.py>
```
reneeotten, line 23:
```suggestion
# <end examples/doc_model_loadmodelresult3.py>
```

reneeotten (commented):
looks good to me; a typo and one suggestion

examples/doc_model_loadmodelresult3.py
newville: thanks.. applied.

lmfit/model.py
newville: that's sort of how I read line 1379:
```if the text is short enough to plausibly be a filename and that file exists: ```

Presetting `mtext.= fname` just avoids an extra "else".  Is that pre-mature optimization?

Comment (reneeotten) [reply lmfit/model.py:1372]: well, to me the logic seems backwards. The "default" assumption should be that we read from a file, if that's not the case the provided text is likely the ModelResult that the user wants to read. 

I would prefer the code to follow that default logic, but your suggestion is fine with me as well and is a bit shorted indeed. In that case I still think that checking the length of "fname" is irrelevant; if the file exists (with whatever name) we can be pretty sure that is what we should use. So I'd prefer to remove the added variable and check for length.

Comment (reneeotten) [reply lmfit/model.py:1372]: or even shorter, but less clear/specific: (1) assume that it is a file, (2) try to open it, (3) if an OSError is thrown assume it is indented as the text of the ModelResult.

Comment (newville) [reply lmfit/model.py:1372]: Yeah, one thing I noticed while writing this is that `Path(fname)` will raise an `OSError` if `fname` is longer than 255 (at least on some OSes -- and so MAXFILENAME_LENGTH should be 256).   So if we are expecting to take a filename or a stored ModelResult, we have to sort of expect long strings. So it could be

```Python
if len(fname) < 256 and Path(fname).exists():
    with open(fname) as fh:
        mtext = fh.read()
else:
    mtext = fname
```

Would that be preferable?  `Path(fname)` or even `len(fname)` could still throw some error that we are not checking for, but those would be "invalid input" to report to the user.

But using try/except:

```Python
try:
     path_exists = Path(fname).exists()
except OSError as exc:
     # really ought to check that the error message was "File Name too long" and reraise if not
     path_exists = False
if path_exists:
    with open(fname) as fh:
        mtext = fh.read()
else:
    mtext = fname
```

That seems a little worse to me

newville:
@reneeotten OK to (rebase and) merge?

PR #1041, opened 2026-07-26
Fix TypeError when independent variable is a list of custom objects (Issue #1040)

### LINKED ISSUE #1040: TypeError when fitting and using a collection of user-defined objects as an independent variable

** Description **
If a collection of user-defined objects which cannot be converted to floats is used as an independent variable in a fit, lmfit hits a `TypeError` when it passes the collection into `coerce_arraylike`. I tried fixing this by adding exception handling to `coerce_arraylike` to just return the input if `coerce_arraylike` hits a `TypeError`, which seems to work. I can submit a pull request if the maintainers believe this is a sensible way to fix the issue.

** A Minimal, Complete, and Verifiable example **
```
# import modules
import lmfit
from dataclasses import dataclass
from collections.abc import Callable
import numpy as np

# define a class and create a collection of custom class instances
@dataclass
class ExampleClass:
    method: Callable[[float], float]
    data: float

def method_1(x):
    return x**2
def method_2(x):
    return 1/(x+3)

custom_object_list = [
    ExampleClass(method_1, 1),
    ExampleClass(method_1, 1),
    ExampleClass(method_2, 5),
]

# define a model function which accepts a custom object list
# in this case, the custom object list doesn't do anything for simplicity's sake
def gaussian(x, amp, cen, wid, custom_object_list):
    return amp * np.exp(-(x-cen)**2 / wid)

gmodel = lmfit.Model(gaussian, independent_vars=['x', 'custom_object_list'])
params = gmodel.make_params(cen=0.3, amp=3, wid=1.25)

# generate some noisy example data
x_eval = np.arange(-10, 10, 0.1)
example_data = (gaussian(x_eval, 5, 2, 3, custom_object_list)
                + np.random.normal(0, 0.3, len(x_eval)))

# evaluate the initial guess of the parameters (works fine)
initial_guess_eval = gmodel.eval(
    x = x_eval,
    params = params,
    custom_object_list = custom_object_list)

plt.plot(x_eval, example_data, ls = 'none', marker = 'o', label = 'data')
plt.plot(x_eval, initial_guess_eval, label = 'initial eval')
plt.legend()

# attempt to evaluate the fit (this yields the error)
fit_result = gmodel.fit(
    example_data,
    params,
    x = x_eval,
    custom_object_list = custom_object_list)
```

** Fit report: **
N/A due to error before during fitting procedure

** Error message: **
```
Cell In[6], line 1
----> 1 fit_result = gmodel.fit(
      2     example_data,
      3     params,
      4     x = x_eval,
      5     custom_object_list = custom_object_list)

File [~/Desktop/main_venv/jupyter_venv/lib/python3.13/site-packages/lmfit/model.py:1172](http://localhost:8889/lab/tree/main_venv/jupyter_venv/lib/python3.13/site-packages/lmfit/model.py#line=1171), in Model.fit(self, data, params, weights, method, iter_cb, scale_covar, verbose, fit_kws, nan_policy, calc_covar, max_nfev, coerce_farray, **kwargs)
   1170     for var in self.independent_vars:
   1171         if var not in params and var in kwargs:
-> 1172             kwargs[var] = coerce_arraylike(kwargs[var])
   1174 if fit_kws is None:
   1175     fit_kws = {}

File [~/Desktop/main_venv/jupyter_venv/lib/python3.13/site-packages/lmfit/model.py:201](http://localhost:8889/lab/tree/main_venv/jupyter_venv/lib/python3.13/site-packages/lmfit/model.py#line=200), in coerce_arraylike(x)
    199 if isinstance(x, (list, tuple, Series)) or hasattr(x, '__array__'):
    200     if np.isrealobj(x):
--> 201         return np.asarray(x, dtype=np.float64)
    202     if np.iscomplexobj(x):
    203         return np.asarray(x, dtype=np.complex128)

TypeError: float() argument must be a string or a real number, not 'ExampleClass'
```

** Version information: **
Python: 3.13.5 (main, May  5 2026, 21:05:52) [GCC 14.2.0]

lmfit: 1.3.4.post55+gcb5c2d8a4, scipy: 1.16.3, numpy: 2.3.5,asteval: 1.0.9, uncertainties: 3.2.3

** Link(s): **

I couldn't find a relevant discussion on the GitHub page. The docstring for `coerce_arraylike` states the intent is to leave other data structures and objects unchanged, and allowing user-defined classes as independent variables seems to be intended, so I feel reasonably confident this `TypeError` I'm hitting is a bug and warrants a GitHub issue. I apologize if I'm mistaken.

[lmfit bug scratch paper.ipynb](https://github.com/user-attachments/files/30363305/lmfit.bug.scratch.paper.ipynb)

lntutor:
#### Description

Closes #1040.

When a list/tuple of custom (non-float) objects is used as an independent
variable in a `Model` fit, `Model.fit` passes it through `coerce_arraylike`,
which raised:

```
TypeError: float() argument must be a string or a real number, not 'ExampleClass'
```

**Root cause.** `coerce_arraylike` used `numpy.isrealobj(x)` to decide whether
to cast the input to `float64`. For a list of custom objects, `np.asarray`
produces an **object-dtype** array, and `numpy.isrealobj` returns `True` for
object dtype (it only checks that the dtype is *not* complex). The code then
reached `np.asarray(x, dtype=np.float64)`, which fails because the objects
have no `float()` conversion — contradicting the function's documented intent
to "leave other data structures and objects unchanged". This matches
@newville's diagnosis in the issue that `numpy.isrealobj()` returns `True` for
a list of dataclasses.

**Fix.** Coerce only data that can actually be represented as a numeric
ndarray. Keep the existing `complex128` path, then *attempt* the `float64`
cast and fall back to returning the input unchanged when the cast raises
`TypeError`/`ValueError`:

```python
if isinstance(x, (list, tuple, Series)) or hasattr(x, '__array__'):
    if np.iscomplexobj(x):
        return np.asarray(x, dtype=np.complex128)
    try:
        return np.asarray(x, dtype=np.float64)
    except (TypeError, ValueError):
        pass
return x
```

This preserves current behavior for real numeric arrays/lists and scalars —
including numeric lists that contain `None`, which NumPy still casts to `NaN`
so that `nan_policy` continues to apply (covered by
`test_model.py::TestUserDefiniedModel::test_lists_become_arrays`) — while no
longer raising for a list of non-float objects. This is the try/except
approach @newville suggested in the issue thread. Thanks to @ChristianGeci for
the clear report and minimal reproducer.

**Tests.** `tests/test_custom_independentvar.py` is extended with:
- `test_list_of_custom_objects_independentvar`: a `Model` fit whose
  independent variable is a list of custom objects now completes without
  raising and the list is passed through unchanged (fails before the fix);
- `test_coerce_arraylike_dtypes`: a direct unit test of `coerce_arraylike`
  covering numeric → `float64`, complex → `complex128`, and object/scalar
  pass-through.

###### Type of Changes
- [x] Bug fix
- [ ] New feature

###### Tested on
```
Python: 3.14.4
lmfit: 1.4.0.dev (branch), scipy: 1.18.0, numpy: 2.5.1, asteval: 1.0.9, uncertainties: 3.2.3
```

###### Verification
- [x] included docstrings that follow PEP 257 (updated the `coerce_arraylike` docstring)
- [x] referenced existing Issue and/or provided relevant link to mailing list (Closes #1040)
- [x] verified that existing tests pass locally (`pytest tests/` → 672 passed, 30 skipped, 1 xfailed)
- [ ] verified that the documentation builds locally
- [x] squashed/minimized your commits and written descriptive commit messages
- [x] added or updated existing tests to cover the changes
- [x] updated the documentation and/or added an entry to the release notes (`doc/whatsnew.rst`)
- [ ] added an example (not applicable for this bug fix)

### DIFF
--- a/doc/whatsnew.rst
+++ b/doc/whatsnew.rst
@@ -28,6 +28,9 @@ Fixes:
 
 - fix cross-Python-version serialization of ``ConstantModel`` and
   ``ComplexConstantModel`` by moving their functions to ``lineshapes`` (Issue #1033)
+- fix ``TypeError`` when an independent variable is a list/tuple of custom,
+  non-float objects by only coercing array-like inputs that can actually be
+  cast to a numeric ndarray (Issue #1040)
 
 .. _whatsnew_134_label:
 
--- a/lmfit/model.py
+++ b/lmfit/model.py
@@ -197,12 +197,22 @@ def coerce_arraylike(x):
     coerce lists, tuples, and pandas Series, hdf5 Groups, etc to an
     ndarray float64 or complex128, but leave other data structures
     and objects unchanged
+
+    Only data that can actually be represented as float64/complex128 is
+    coerced. ``numpy.isrealobj`` returns ``True`` for an object-dtype
+    array (for example a list of custom, non-float objects), so relying on
+    it would attempt -- and fail with a ``TypeError`` -- to cast such data
+    to float64 (see GitHub issue #1040). Instead the conversion is tried
+    and any array-like that cannot be cast to a numeric ndarray is returned
+    unchanged.
     """
     if isinstance(x, (list, tuple, Series)) or hasattr(x, '__array__'):
-        if np.isrealobj(x):
-            return np.asarray(x, dtype=np.float64)
         if np.iscomplexobj(x):
             return np.asarray(x, dtype=np.complex128)
+        try:
+            return np.asarray(x, dtype=np.float64)
+        except (TypeError, ValueError):
+            pass
     return x
 
 
--- a/tests/test_custom_independentvar.py
+++ b/tests/test_custom_independentvar.py
@@ -1,6 +1,7 @@
 import numpy as np
 
 from lmfit.lineshapes import gaussian
+from lmfit.model import coerce_arraylike
 from lmfit.models import Model
 
 
@@ -14,8 +15,21 @@ class Stepper:
         return np.linspace(self.start, self.stop, self.npts)
 
 
+class CustomObject:
+    """A minimal user-defined object that cannot be coerced to a float."""
+
+    def __init__(self, value):
+        self.value = value
+
+
 def gaussian_mod(obj, amplitude, center, sigma):
     return gaussian(obj.get_x(), amplitude, center, sigma)
+
+
+def gaussian_with_objlist(x, amplitude, center, sigma, custom_object_list=None):
+    """Gaussian model that also accepts a list of custom objects as an
+    independent variable; the list is not used in the calculation."""
+    return gaussian(x, amplitude, center, sigma)
 
 
 def test_custom_independentvar():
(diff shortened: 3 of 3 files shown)


newville:
@lntutor  @ChristianGeci thanks.  That looks good.  

At #1040 we also discussed changing the default value of `Model.fit(..., coerce_farray=True)` from `True` to `False`.     I think that is probably enough like "API-change" that we should reserve that for a minor version number change.

My inclination is to merge this as-is in a few days, unless there are concerns or suggestions.

newville:
@lntutor Thanks!

PR #1042, opened 2026-07-30
Fix corrupted brute() candidates for a single varying parameter

steps-re:
#### Description
When `Minimizer.brute()` runs with a **single** varying parameter, `scipy.optimize.brute` returns the grid (`ret[2]`) as a bare 1-D array, dropping the leading "nvars" axis it includes for 2+ parameters. The existing single-parameter branch didn't account for that, so every entry in `result.candidates` stored the parameter value as a 1-element ndarray (`array([1.9])`) instead of a scalar (`1.9`). That leaks into anything that reads or compares `candidate.params[name].value`.

The fix normalizes the single-parameter grid to match the multi-parameter shape (`grid_points = [np.asarray(ret[2]).ravel()]`) and unifies the `grid_result` construction so candidates always carry scalar values.

###### Type of Changes
- [x] Bug fix

###### Tested on
Python: 3.13.9, lmfit: 1.3.4.post63, scipy: 1.18.0, numpy: 2.5.1

###### Verification
- [x] existing tests pass locally (full suite: 673 passed, 30 skipped, 1 xfailed)
- [x] added a regression test `test_brute_one_parameter_candidates` that fails on current master (`array([1.9])`) and passes with the fix

—
Disclosure: I used AI assistance in finding and preparing this fix. I verified the bug and fix by hand against current master (the added test reproduces it) and ran the full suite locally.

### DIFF
--- a/lmfit/minimizer.py
+++ b/lmfit/minimizer.py
@@ -1943,14 +1943,14 @@ class Minimizer:
 
             # sort the results of brute and populate .candidates attribute
             grid_score = ret[3].ravel()  # chisqr
-            grid_points = [par.ravel() for par in ret[2]]
 
             if len(result.var_names) == 1:
-                grid_result = np.array([res for res in zip(zip(grid_points), grid_score)],
-                                       dtype=[('par', 'O'), ('score', 'float')])
+                grid_points = [np.asarray(ret[2]).ravel()]
             else:
-                grid_result = np.array([res for res in zip(zip(*grid_points), grid_score)],
-                                       dtype=[('par', 'O'), ('score', 'float')])
+                grid_points = [par.ravel() for par in ret[2]]
+
+            grid_result = np.array([res for res in zip(zip(*grid_points), grid_score)],
+                                   dtype=[('par', 'O'), ('score', 'float')])
             grid_result_sorted = grid_result[grid_result.argsort(order='score')]
 
             result.candidates = []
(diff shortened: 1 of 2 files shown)


newville:
@steps-re Thanks - I can certainly believe that using `brute` with 1 variable is not well tested.  Or perhaps a change in NumPy 2 altered an assumption made in the original code?

I am in favor of merging this, but will wait a few days to allow others to give any comments or suggestions.

lmfit/minimizer.py
```diff
@@ -1943,14 +1943,21 @@ def brute(self, params=None, Ns=20, keep=50, workers=1, max_nfev=None):
...
             grid_score = ret[3].ravel()  # chisqr
-            grid_points = [par.ravel() for par in ret[2]]
 
+            # NOTE: for a single varying parameter, scipy.optimize.brute
+            # returns the grid (ret[2]) as a bare 1-D array of grid values
+            # (it strips the leading "nvars" axis that is present for 2+
+            # varying parameters). Iterating over that bare array yields
+            # individual scalars rather than one array per parameter, so
+            # it must be wrapped in a list to match the convention used
+            # for the multi-parameter case below.
```
reneeotten:
```suggestion
```

reneeotten (changes requested):
please remove the 7 lines of comments in the code; there is no need for long explanations. The issue is fixed that's enough. 

Additionally, please rephrase the commit summary as:
```FIX: return correct candidates in 'brute' for single varying parameter``` and remove the text in the commit body.

steps-re:
Done. Removed the comment block, renamed the commit summary to the wording you gave, and dropped the commit body.

reneeotten:
thanks @steps-re

